PR #65, opened 2024-01-26
2 double kubectl instances

abhitoshj:
### Motivation
Currently, the bin folder in k2s installation directory comprises of 2 kubectl binaries present in different locations. The PR is fix for issue #2 

### Modifications
The following changes are implemented:
- Removed the redundant kubectl binary from k2s installation directory.
- Ensured that the k2s implementation always uses the kubectl binary present in the k2s installation directory. This also ensures that the user does not experience any glitch due to some other kubectl binary (probably an old version) present in %PATH%. The adaptation is extended to the libs.

### Verification
- Executed all Unit tests.
- Executed all e2e acceptance tests for installation and addons.
- Manual test on host variant with refactored modules


smallsetup/UninstallK8s.ps1
```diff
@@ -108,11 +108,8 @@ if ($global:PurgeOnUninstall) {
     Remove-Item -Path "$global:NssmInstallDirectory\nssm.exe" -Force -ErrorAction SilentlyContinue
     Remove-Item -Path "$global:SetupJsonFile" -Force -ErrorAction SilentlyContinue
     Remove-Item -Path "$global:KubernetesImagesJson" -Force -ErrorAction SilentlyContinue
-    Remove-Item -Path "$global:BinPath\kube*.exe" -Force -ErrorAction SilentlyContinue
     Remove-Item -Path "$global:BinPath\cri*.exe" -Force -ErrorAction SilentlyContinue
-    Remove-Item -Path "$global:BinPath\vc_redist*.exe" -Force -ErrorAction SilentlyContinue
```
mahenarayan, line 113: @abhitoshj  these removal implies kube tools are cleaned as part of ExecutableFolderPath correct?

mahenarayan (approved):

smallsetup/UninstallK8s.ps1
mahenarayan: Also, for upgrade scenario, we should cleanup the old instance during the upgrade.

mahenarayan (changes requested):
Upgrade usecase, we should cleanup the old instance during the upgrade.

smallsetup/UninstallK8s.ps1
abhitoshj: kubetools are deleted as part of ExecutableFolderPath

For upgrade scenario, I can bring back the removal from bin folder for kubectl. vc_redist and helm are not being downloaded so probably their removal can remain unchanged.

Some points to note here:
- k2s does not have yet an old version. Should the upgrade scenario be handled now ?
- When the old version of k2s is uninstalled, the bin folder will be removed from PATH. At the same time, the new version always uses kubectl from its own bin directory, so even if kubectl is not deleted from old version's bin, it does not cause hinderance. 
- After the upgrade is finished, the user will be asked to reload shell so the old version's kubectl can never be resolved from PATH
- If we bring back this code, then at some point of time, we should also plan to remove this because it will not be applicable.

mahenarayan (approved):

Issue #88, opened 2024-02-01
k2s addons enable exthttpaccess outputs an error


jgriegershs:
When running `k2s addons enable exthttpaccess`, an error occurs even though the exit code is 0:
```
⏳ [13:20:07] Obtaining IPs of active physical net adapters
⏳ [13:20:09] Creating nginx configuration file
⏳ [13:20:09] Downloading nginx executable
⏳ [13:20:09] Downloading 'https://nginx.org/download/nginx-1.23.2.zip' to 'C:\ws\k2sv2\bin\nginx\nginx.zip' (no proxy)
⏳ x nginx.exe
⏳ [13:20:10] Registering nginx service
⏳ [13:20:10] Service "ExtHttpAccess-nginx" installed successfully!
⏳ ExtHttpAccess-nginx: Unexpected status SERVICE_PAUSED in response to START control.
⏳
E0201 13:20:10.880625   12612 psexecutor.go:197] ExtHttpAccess-nginx: Unexpected status SERVICE_PAUSED in response to START control.
```

- [ ] fix the error
- [x] createn an e2e test for *exthttpaccess* addon

gabriel-rosso:
The problem arises because there is already a process listening on the port 80. Just opening the browser an entering [http://localhost:80](http://localhost/) you'll see that the IIS server is accessed. 
A hard check for the ports 80 and 443 has to be introduced. If it is detected that they are already used then the alternative ports 8080 and 8443 must be used after user confirmation.

PR #121, opened 2024-02-07
Added parameter to force the use of alternative ports (#88)

gabriel-rosso:
#88 
Added a parameter to the Enable.ps1 of the addon exthttpaccess to force the use of alternative ports (8080/8443) instead of 80/443 without user confirmation.
This replaces the environment variable that was used before.


addons/exthttpaccess/addon.manifest.yaml
```diff
@@ -26,6 +26,10 @@ spec:
             shorthand: p
             default: ""
             description: HTTP Proxy
+          - name: force-use-alternative-ports-if-needed
```
mahenarayan, line 29: It would be nice to have a simple flag name e.g. **--force-use-ports** and explain further in the usage notes/description. Too long flag is error prone for user. What do you think @gabriel-rosso ?

mahenarayan (commented):
Please do not merge k2s.exe, this is done by the workflow.

addons/exthttpaccess/addon.manifest.yaml
gabriel-rosso: That flag name conveys the real intention of what the user does/gets when using it.
Using a shorter name would expect that the user remembers what he read in the help description (when using it the first time, and when he comes back to the code some time later).

The drawbacks of shorter names:
'--force-use-ports': the addon is not deliberately changing the ports.
'--force': the addon is not forcing the enabling on the usual HTTP(S) ports.

For a shorter name there is the shortcut '-f'

Examples of long flag names are:
--allow-missing-template-keys (kubectl create)
--discovery-token-unsafe-skip-ca-verification (kubeadm join)

mahenarayan: @gabriel-rosso Yes, kubectl is notorious for long flag names which we should avoid. 
Can this help? In this case, it clearly indicates addon will switch to alternative ports when 80/443 are busy.
`-a --auto-alt-ports   Automatically switch to alternative ports 8080/8443 if 80/443 are busy without user confirmation`

gabriel-rosso: @mahenarayan What about `--autoconfirm-alt-ports` ?

PR #142, opened 2024-02-12
67 k2s uninstall k2s cli logs get deleted during uninstall

achimkist:
k2s uninstall k2s cli logs get deleted during uninstall


mahenarayan (commented):
Looks good.

PR #220, opened 2024-02-27
Cache k2s vswitches across restarts

abhitoshj:
Fixes #TODO

### Motivation

Sometimes, when the cluster is started and stopped quite a number of times on a host, it is not possible to set the vEthernet (Loopbackk2s) or vEthernet(KubeSwitch) to private. The system reaches a state that no communication works. The host has to be restarted in order to able to start the cluster successfully. This could be due to some corruption in HNS or network card driver.

### Modifications

Provided an option in stop called ***--cache-vswitch*** with which it shall be possible to cache the cbr0 and kubeswitch for later use. 
Provided an option in start called ***--autouse-cached-vswitch*** with which it shall be possible to use the cached cbr0 and kubeswitch.
Since the cbr0 and kubeswitch are created only once and re-used, the corruption in HNS and/or network driver is prevented.


mahenarayan:
I have two flag names which might be suitable:

```
k2s start --autouse-cached-vswitch             Automatically utilizes the cached vSwitch 'cbr0' for node-to-node communication on the host machine.

k2s start --autouse-cached-bridge             Automatically utilizes the cached vSwitch 'cbr0' for node-to-node communication on the host machine.


```

abhitoshj:
> I have two flag names which might be suitable:
> 
> ```
> k2s start --autouse-cached-vswitch             Automatically utilizes the cached vSwitch 'cbr0' for node-to-node communication on the host machine.
> 
> k2s start --autouse-cached-bridge             Automatically utilizes the cached vSwitch 'cbr0' for node-to-node communication on the host machine.
> ```

I think we can use --autouse-cached-vswitch because we cache cbr0 and kubeswitch on stop to prevent the corruption.

Issue #257, opened 2024-03-04
Automatically takeover system proxy of Windows


krotz-dieter:
In case that the host system has an system proxy set (Settings-> Network & Internet -> Manual proxy setup)
the install command shall automatically pick that proxy up if no explicit proxy was set.
The Install command shall behave in this case in the same way as if the same proxy would have been provided
by the proxy parameter.

PR #309, opened 2024-04-01
#257 automatically takeover system proxy of windows

abhitoshj:
If the proxy is configured in Proxy Settings of windows host, then the K2s installation automatically considers this proxy for installation process.

In such a scenario, the user does not need specify the proxy using --proxy parameter during k2s install.


mahenarayan (changes requested):
Would it be possible to move this duplicate code to a common module, eg. proxy module under windows? At least for addons, duplicates in install is okay for now.

```
if ($Proxy -eq "") {
  Write-Log "Determining if proxy is configured by the user in Windows Proxy settings." -Console
  $proxyEnabledStatus = Get-ProxyEnabledStatusFromWindowsSettings
  if ($proxyEnabledStatus) {
      $Proxy = Get-ProxyServerFromWindowsSettings
      Write-Log "Configured proxy server in Windows Proxy settings: $Proxy" -Console
  } else {
      Write-Log "No proxy configured in Windows Proxy Settings." -Console
  }
}
```

mahenarayan:
I would also include HTTPS_PROXY env for automatic proxy detection. This is a common convention in most of the k8s distributions during install. What do you think @abhitoshj ?

mahenarayan (approved):
Looks good, thanks for making changes.

PR #313, opened 2024-04-02
Fix handling for enabling missing windows features during install

achimkist:
304 enabling missing windows features throws error for server oss feature name is unknown


achimkist:
@krotz-dieter with those changes user gets informed about required reboot after enabling missing windows features and installation is stopped. Furthermore windows features list is now seperated for Windows 10/11 and Windows OSs.

Issue #258, opened 2024-03-04
[addons] New addon security for the K2s cluster 


krotz-dieter:
The new security addon shall provide two basic modes of security:
1. basic security
- provide to the outside of cluster secure communication using TLS termination & OAuth2 in the ingress controller
- inside the cluster still the communication is unsecure
to checked all 2 addons: gateway-nginx, ingress-nginx, traefik
Variants which can be used: default variant (host & virtual machine)

2. enhanced security
- provide an overall secure communication through the usage of sidecar containers which shall hijack the traffic
and add the security aspects: TLS and authentication/authorization

Used open source components:
Certificate management (with own certificate option): cert manager, ...
OAuth2 Identity Provider (with own identity provider): key cloak, ...
Service mesh: linkerd, istio, ...

The overall target shall be the idea that security comes in K2s through the infrastructure.
By just enabling the security addon the cluster shall provide security through HTTPS and usage of an identity provider using oAuth2:
`k2s addons enable security (parameters)`
-> afterwards login needs to be done and communication encrypted

`k2s addons disable security`
-> afterwards no login and pure HTTP

krotz-dieter:
1. For the basic security all capabilities of the existing ingress/gatewayAPI addons in K2s shall be checked.
2. For the extended security:
- usage of service meshes like istio, linkerd or other alternatives (sidecar handling)
- certificate handling: cert manager, ..
- identity server: keycloak, ...

krotz-dieter:
Completed now with addons also included in service mesh

PR #465, opened 2024-05-08
central ingress tls

artur-pusztai:
### Motivation
This is part of the feature **#258: New addon security for the K2s cluster** and is addressing the TLS termination at Ingress (both: NGINX and Traefik)

This PR makes it possible for more web applications hosted on K2s to shall share one common ingress host, and expose their functionality under relative paths: `k2s.cluster.net`

Both `ingress-nginx` and `traefik` addons come now with a central TLS Ingress with no paths, but with a tls entry matching the mentioned hostname.

All Web UIs of our AddOns contain now a second ingress route, which is merged with the central one, so that each UI is now available under 3 URLs:
- Dashboard:  
  http://k2s-dashboard.local
  https://k2s-dashboard.local
  **new:** https://k2s.cluster.net/dashboard/
- Logging:
  http://k2s-logging.local
  https://k2s-logging.local
  **new:** https://k2s.cluster.net/logging/
- Monitoring:
  http://k2s-monitoring.local
  https://k2s-monitoring.local
  **new:** https://k2s.cluster.net/monitoring/

All TLS endpoints are secured with a certificate generated by **cert-manager**, which is installed with the **security** addon. The configuration files and scripts of all other addons are desined in such a way that they work regardless of the installation status of the **security** addon.

Security can be enabled at the end, at beginning or in between. As soon as it is enabled, secure Certificates will be used for all TLS-terminated ingresses.

### Modifications
These addons and their e2e tests where updated:
- dashboard
- logging
- monitoring
- ingress-nginx
- traefik
- security

### Verification
`k2s/test/e2e` tests are adapted / extended for addons mentioned above

### DIFF
--- a/addons/monitoring/manifests/plutono/configmap.yaml
+++ b/addons/monitoring/manifests/plutono/configmap.yaml
@@ -30,7 +30,7 @@ data:
     provisioning = /etc/plutono/provisioning
     [server]
     # these settings needed to make it work with Ingress
-    root_url = https://k2s-monitoring.local/monitoring
+    root_url = https://k2s.cluster.net/monitoring
     domain = k2s.cluster.net
     protocol = https
     cert_key = /cert/tls.key
--- a/addons/monitoring/manifests/plutono/configmap.yaml
+++ b/addons/monitoring/manifests/plutono/configmap.yaml
@@ -30,8 +30,8 @@ data:
     provisioning = /etc/plutono/provisioning
     [server]
     # these settings needed to make it work with Ingress
-    root_url = https://k2s-monitoring.local
-    domain = k2s-monitoring.local
+    root_url = https://k2s-monitoring.local/monitoring
+    domain = k2s.cluster.net
     protocol = https
     cert_key = /cert/tls.key
     cert_file = /cert/tls.crt
--- a/addons/security/Enable.ps1
+++ b/addons/security/Enable.ps1
@@ -37,6 +37,7 @@ $addonsModule = "$PSScriptRoot\..\addons.module.psm1"
 $securityModule = "$PSScriptRoot\security.module.psm1"
 
 Import-Module $infraModule, $clusterModule, $nodeModule, $addonsModule, $securityModule
+Import-Module PKI;
 
 Initialize-Logging -ShowLogs:$ShowLogs
 
--- a/addons/security/Enable.ps1
+++ b/addons/security/Enable.ps1
@@ -128,7 +128,11 @@ Write-Log 'Importing CA root certificate to trusted authorities of your computer
 $b64secret = (Invoke-Kubectl -Params '-n', 'cert-manager', 'get', 'secrets', 'ca-issuer-root-secret', '-o', 'jsonpath', '--template', '{.data.ca\.crt}').Output
 $tempFile = New-TemporaryFile
 [Text.Encoding]::Utf8.GetString([Convert]::FromBase64String($b64secret)) | Out-File -Encoding utf8 -FilePath $tempFile.FullName -Force
-certutil -addstore -f -enterprise -user root $tempFile.FullName
+$params = @{
+    FilePath          = $tempFile.FullName
+    CertStoreLocation = 'Cert:\LocalMachine\Root'
+}
+(Import-Certificate @params).Output | Write-Log
 Remove-Item -Path $tempFile.FullName -Force
 
 Add-AddonToSetupJson -Addon ([pscustomobject] @{Name = 'security' })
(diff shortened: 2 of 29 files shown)


k2s/test/e2e/addons/security/security_test.go
```diff
@@ -110,7 +112,29 @@ var _ = Describe("'security' addon", Ordered, func() {
...
+	})
+
 	It("disables the addon", func(ctx context.Context) {
 		suite.K2sCli().Run(ctx, "addons", "disable", addonName, "-o")
 	})
+
+	It("uninstalls cmctl.exe, the cert-manager CLI", func(ctx context.Context) {
+		cmCtlPath := path.Join(suite.RootDir(), "bin", "exe", "cmctl.exe")
+		_, err := os.Stat(cmCtlPath)
+		Expect(err).NotTo(BeNil())
```
jgriegershs, line 133: To be more precise, you could check the `err` for `os.IsNotExist(err)`

jgriegershs (changes requested):

jgriegershs:
Added @achimkist as reviewer since he has more insights into the manifests :-).

krotz-dieter (approved):
k2s.cluster.net is great to have, thanks for the changes

addons/monitoring/manifests/plutono/configmap.yaml
achimkist, line 1: I really like the new approach with `https://k2s.cluster.net/monitoring` but this will not work anymore now with the old one `k2s-monitoring.local`.

addons/monitoring/manifests/plutono/configmap.yaml
```diff
@@ -30,8 +30,8 @@ data:
     provisioning = /etc/plutono/provisioning
     [server]
     # these settings needed to make it work with Ingress
-    root_url = https://k2s-monitoring.local
-    domain = k2s-monitoring.local
+    root_url = https://k2s-monitoring.local/monitoring
+    domain = k2s.cluster.net
```
achimkist, line 34: I really like the new approach with `https://k2s.cluster.net/monitoring` but this will not work anymore now with the old one `k2s-monitoring.local`.

achimkist (changes requested):
I would go with Artur's new approach `https://k2s.cluster.net/<addon>` and remove the old one completely e.g. "https://k2s-<addon>.local" (both will not work e.g. for monitoring addon see my comment).

addons/monitoring/manifests/plutono/configmap.yaml
artur-pusztai: I also wondered why, but it DOES work - I manually tested all combinations, and I also e2e test them.
But I will double-check once more!

Yes, we could deprecate the old approach, of using the hostnames for the ingress path matches.
On the other hand, these addons serve as "how to" for product developers, and I believe it is also a cool use case - sometimes you want this and sometimes that.

I want us to think more on when and what to remove - one solution would be to keep e.g. the k2s-dashboard.local, and move all others to the central cluster ingress...

krotz-dieter: We should keep the possibility to have specific host names for each app. The default shall be k2s.cluster.net. I wonder why it does not work in your case Joachim ?

achimkist: I haven't tested it (will test it today) but now there is a mix of both - `https://k2s-monitoring.local/monitoring` should be `https://k2s.cluster.net/monitoring` right?

artur-pusztai: absolutely, it is inconsistent and I'll correct it.

addons/security/Enable.ps1
```diff
@@ -128,7 +128,11 @@ Write-Log 'Importing CA root certificate to trusted authorities of your computer
 $b64secret = (Invoke-Kubectl -Params '-n', 'cert-manager', 'get', 'secrets', 'ca-issuer-root-secret', '-o', 'jsonpath', '--template', '{.data.ca\.crt}').Output
 $tempFile = New-TemporaryFile
 [Text.Encoding]::Utf8.GetString([Convert]::FromBase64String($b64secret)) | Out-File -Encoding utf8 -FilePath $tempFile.FullName -Force
-certutil -addstore -f -enterprise -user root $tempFile.FullName
+$params = @{
+    FilePath          = $tempFile.FullName
+    CertStoreLocation = 'Cert:\LocalMachine\Root'
+}
+(Import-Certificate @params).Output | Write-Log
```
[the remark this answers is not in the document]
artur-pusztai: Added this line in both scripts - is it enough?
Import-Module PKI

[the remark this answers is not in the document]
artur-pusztai: This will be no-op when imported, will silently import it if available and fail with an error if the module is not available.

PR #510, opened 2024-05-29
Implemented NlaSvc restart

abhitoshj:
Implemented a restart of Network Location Awareness service to eliminate issues with cbr0 switch creation.

### DIFF
--- a/lib/scripts/k2s/start/Start.ps1
+++ b/lib/scripts/k2s/start/Start.ps1
@@ -192,6 +192,7 @@ $switchname = Get-ControlPlaneNodeDefaultSwitchName
 Write-Log 'Configuring network for windows node' -Console
 Restart-WinService 'vmcompute'
 Restart-WinService 'hns'
+Restart-NlaSvc
 
 $adapterName = Get-L2BridgeName
 Write-Log "Using network adapter '$adapterName'"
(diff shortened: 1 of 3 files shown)


lib/scripts/k2s/start/Start.ps1
```diff
@@ -328,6 +328,8 @@ route delete $clusterCIDRServicesWindows >$null 2>&1
 Write-Log "Add route to $clusterCIDRServicesWindows"
 route -p add $clusterCIDRServicesWindows $ipControlPlane METRIC 7 | Out-Null
 
+Restart-NlaSvc
```
mahenarayan, line 331: Can we do this when HNS, vmcompute services are restarted? Is there a particular significance doing it here? Same goes to Start script as well.
https://github.com/Siemens-Healthineers/K2s/blob/488681bb7a79f3a808be0d9f91208bbc67a96a6f/lib/scripts/k2s/start/Start.ps1#L193

smallsetup/StartK8s.ps1
```diff
@@ -423,6 +423,8 @@ route delete $clusterCIDRServicesWindows >$null 2>&1
 Write-Log "Add route to $clusterCIDRServicesWindows"
 route -p add $clusterCIDRServicesWindows $ipControlPlane METRIC 7 | Out-Null
 
+Restart-NlaSvc
```
mahenarayan, line 426: Can we do this when HNS, vmcompute services are restarted? Is there a particular significance doing it here? Same goes to Start script as well.

mahenarayan (changes requested):

lib/scripts/k2s/start/Start.ps1
abhitoshj: The idea was to do it anytime before BeforeStartK8sNetwork hook. It can be moved also to be restarted with HNS and vmcompute

mahenarayan (approved):
Looks good to me.

PR #546, opened 2024-06-05
Only restart NlaSvc when inconsistency is suspected

abhitoshj:
Rather than restarting the NlaSvc everytime, we restart the NlaSvc after some iterations are exhausted and the interface is not set to private.


mahenarayan (approved):
So now we restart nlasvc when setting interface setting to private fails. Fair enough. Looks good.

Issue #511, opened 2024-05-29
cni0 in kubemaster is sprodically not assigned with pod network cidr on K2s start


abhitoshj:
Sometimes during k2s start, it is observed that the cni0 interface on kubemaster is not assigned with ip address 172.20.0.1 but a random IP address. 
When the user wants to start a linux pod in this state, the pod remains stuck in ContainerCreating state. 

On describing the pod the following is seen
```
[12-05-2024 21:39:39.577 INF]   Warning  FailedCreatePodSandBox  5s (x2 over 16s)  kubelet            (combined from similar events): Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create pod network sandbox k8s_traefik-6479f6fff4-qhctl_traefik_522bf2e3-f7d2-40a7-892e-2131c5016f36_0(c0935cf26e4fb6f99157a6e5ae7570937d098307e1e54a86c1aa7a9b258a45a4): error adding pod traefik_traefik-6479f6fff4-qhctl to CNI network "cbr0": plugin type="flannel" failed (add): failed to delegate add: failed to set bridge addr: "cni0" already has an IP address different from 172.20.0.1/2
```

At this point, if we check the network configurations in kubemaster, the following is seen:
```
kubemaster:~$ ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1400 qdisc mq state UP group default qlen 1000
    link/ether 00:15:5d:2a:be:6f brd ff:ff:ff:ff:ff:ff
    inet 172.19.1.100/24 brd 172.19.1.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::215:5dff:fe2a:be6f/64 scope link
       valid_lft forever preferred_lft forever
3: cni0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether c2:8b:f9:da:55:eb brd ff:ff:ff:ff:ff:ff
[...]

PR #569, opened 2024-06-13
#511 Restart kubemaster when cni0 not initialized correctly

abhitoshj:
Implemented a restart mechanism for Kubemaster when cni0 is not initialized with the correct IP address. Kubemaster shall be restarted atmost 3 times when this issue happens.

### DIFF
--- a/smallsetup/StartK8s.ps1
+++ b/smallsetup/StartK8s.ps1
@@ -135,6 +135,77 @@ function CheckKubeSwitchInExpectedState() {
     return $true
 }
 
+function Restart-ControlPlane {
+    param (
+        [string] $ControlPlaneVMName,
+        [string] $ControlPlaneIpAddr,
+        [bool] $WSL,
+        [string] $ReusingExistingLinuxMachine
+    )
+    Write-Log "Restarting $ControlPlaneVMName..."
+    #check VM status
+    $Iteration = 0
+    if (!$WSL -and !$ReusingExistingLinuxComputer) {
+        while ($true) {
+            $Iteration++
+            Write-Log "VM Handling loop (iteration #$Iteration):"
+            Start-Sleep -s 4
+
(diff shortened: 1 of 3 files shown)


smallsetup/StartK8s.ps1
```diff
@@ -312,67 +383,30 @@ if (!$WSL -and !$isReusingExistingLinuxComputer) {
...
-            $startCycle = 0
-            $startState = (Get-VM -Name $controlPlaneVMHostName).State -eq [Microsoft.HyperV.PowerShell.VMState]::Running
-            while (!$startState -And $startCycle -lt 3) {
-                $startCycle++
-                Write-Log "           still waiting for start, current VM state: $(Get-VM -Name $controlPlaneVMHostName | Select-Object -expand 'State')"
-                Start-VM -Name $controlPlaneVMHostName
-                Start-Sleep -s 4
-                $startState = (Get-VM -Name $controlPlaneVMHostName).State -eq [Microsoft.HyperV.PowerShell.VMState]::Running
-            }
+            throw "cni0 interface in $controlPlaneVMHostName is not correctly initialized after $i retries."
```
mahenarayan, line 403: Before throwing exception, is there a workaround which we can inform user to try like start and stop? Not mandatory, just lay out a set of instructions to user if there a way to overcome this problem.

mahenarayan (approved):

smallsetup/StartK8s.ps1
abhitoshj: Ideally, the restart of the kubemaster should solve the incorrectly initialized cni0. But if it is not the case, then we have to look as to why that is the case.

Currently, when the exception is thrown at this stage, all K2s services are not deployed. This means that we cannot connect to kubemaster via system ssh commands because the command would first check if k2s is running.

PR #622, opened 2024-06-27
feat(upgrade): specify backup directory

achimkist:
feat(upgrade): specify backup directory

### DIFF
--- a/lib/modules/k2s/k2s.cluster.module/upgrade/upgrade.module.psm1
+++ b/lib/modules/k2s/k2s.cluster.module/upgrade/upgrade.module.psm1
@@ -150,7 +150,6 @@ function Export-NotNamespacedResources {
                 .items[].metadata.uid,
                 .items[].metadata.selfLink,
                 .items[].metadata.creationTimestamp,
-                .items[].metadata.annotations,
                 .items[].metadata.generation,
                 .items[].metadata.ownerReferences,
                 .items[].spec.finalizers,
@@ -160,7 +159,6 @@ function Export-NotNamespacedResources {
                 .metadata.uid,
                 .metadata.selfLink,
                 .metadata.creationTimestamp,
-                .metadata.annotations,
                 .metadata.generation,
                 .metadata.ownerReferences)'
             $filter = $filter -replace '\r*\n', ''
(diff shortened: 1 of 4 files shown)


lib/modules/k2s/k2s.cluster.module/upgrade/upgrade.module.psm1
```diff
@@ -150,7 +150,6 @@ function Export-NotNamespacedResources {
                 .items[].metadata.uid,
                 .items[].metadata.selfLink,
                 .items[].metadata.creationTimestamp,
```
mahenarayan, line 152: What was the reason to remove annotations here?

lib/scripts/k2s/system/upgrade/Start-ClusterUpgrade.ps1
```diff
@@ -81,17 +79,21 @@ function Start-ClusterUpgrade {
...
     )
     $errUpgrade = $null
     $addonsBackupPath = $null
     $logFilePathBeforeUninstall = $null
-    $tpath = $null
     $coresVM = $null
     $memoryVM = $null
     $storageVM = $null
 
+    if ($BackupDir -eq '') {
```
mahenarayan, line 93: Does it makes sense to check if BackupDir really exists when passed by user?. If not throw before upgrade procedure starts.

lib/modules/k2s/k2s.cluster.module/upgrade/upgrade.module.psm1
achimkist: This topic came up during evaluation of Openebs provisioner e.g. the storage class uses annotations to specify quota enforcement:

```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    cas.openebs.io/config: |
      - name: StorageType
        value: "hostpath"
      - name: BasePath
        value: "/var/openebs/local"
      - name: EXT4Quota
        enabled: "true"
        data:
          softLimitGrace: "0%"
          hardLimitGrace: "0%"
```

If we exclude annotations during upgrade quota enforcement is not active anymore. This could also happen in other use cases.

lib/scripts/k2s/system/upgrade/Start-ClusterUpgrade.ps1
achimkist: Default value of BackupDir is empty string even if the user does not specify a path. Then the temp dir should be used

mahenarayan: I mean, if the user passes invalid path, do we check that?

achimkist: yes you're right, this is missing

PR #625, opened 2024-06-28
feat(security): ExternalDNS

artur-pusztai:
### Motivation

This is part of #258   
Hosts of endpoints exposed by the Kubernetes cluster (ingresses) shall be looked up by the DNS proxy.  
Today, these hosts need to be added to the hosts file and also to the coredns configuration, which are additional steps
and can lead to inconsistencies. 

### Modifications

See the [ExternalDNS Operator Manual](https://github.com/Siemens-Healthineers/K2s/blob/external-dns/docs/op-manual/external-dns.md)

1. ExternalDNS is deployed with both `ingress-nginx` and  `traefik`. It observes ingresses and adds DNS entries in SKYDNS format to the /skydns prefix in the configured `etcd`
2. CoreDNS is extended with an `etcd` plugin, and answers DNS queries from these entries made by ExternalDNS in etcd.
3. For dashboard, logging and monitoring, the second host names where renamed e.g. from `k2s-dashboard.local` to `k2s-dashboard.cluster.local`. **Not yet** for `registry.local` and `k2s-gateway.local`
4. None of the hostnames on the domain 'cluster.local' is added anymore to hosts file. 
5. During `k2s start`, the regular DNS servers of the physical interfaces are configured as upstream of `dnsproxy`. After that, the DNS of the physical interfaces are then pointing to `dnsproxy`
6. During `k2s stop`, the DNS servers of these physical interfaces is reset to automatic.
7. CoreDNS and ExternalDNS need access to etcd. We use the main cluster etcd. The CA and client certificates created during installation are imported also as K8s secrets, and used from there - opening the way to manage them with Cert Manager in a subsequent step.

overview
### Verification

No functionality is added, the verification is that the whole functionality still works.  
Verification Build:
http://10.81.79.18/argo/workflows/kondor-ci/test-k2s-addons-online-windows10-khtkb?tab=workflow&uid=ee1c68cf-ff29-47da-8f22-9508bc9b1584


addons/traefik/manifests/deployment.yaml
```diff
@@ -110,6 +110,8 @@ spec:
           - "--providers.kubernetescrd.ingressclass=traefik"
           - "--providers.kubernetesingress"
           - "--providers.kubernetesingress.ingressclass=traefik"
+          - "--providers.kubernetesIngress.ingressEndpoint"
+          - "--providers.kubernetesIngress.ingressEndpoint.ip=172.19.1.100"
```
mahenarayan, line 114: @artur-pusztai Fixed masterip of 172.19.1.100 is not always true, in k2s user can change the masterip through config.json. Get the configured masterip through `Get-ConfiguredIPControlPlane`.

mahenarayan (commented):
Looks good

addons/traefik/manifests/deployment.yaml
artur-pusztai: oh, I was too fast - or you too late :-)
We need to fix this later, I agree. 
We can add this two line in Enable.ps1, so we can inject the IP address dynamically.

mahenarayan: @artur-pusztai No issues, you can fix it in next PR.

PR #634, opened 2024-07-02
fix(traefik): ingressEndpoint set dynamically

artur-pusztai:
Fixes the hard-coded ingressEndpoint in traefik deyploment file

### Motivation
traefik manifests was containing hardcoded IP address of the control plane, which is not the same in all deployments.

### Modifications

The IP Address is set now during enabling of the addon

### Verification

n.a.

### DIFF
--- a/addons/traefik/Enable.ps1
+++ b/addons/traefik/Enable.ps1
@@ -98,7 +98,7 @@ Write-Log 'Installing external-dns' -Console
 $externalDnsConfig = Get-ExternalDnsConfigDir
 (Invoke-Kubectl -Params 'apply' , '-k', $externalDnsConfig).Output | Write-Log
 
-Write-Log "Preparing kusomization with $controlPlaneIp as an external IP for traefik service" -Console
+Write-Log "Preparing kustomization with $controlPlaneIp as an external IP for traefik service" -Console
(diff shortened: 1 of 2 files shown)


addons/traefik/Enable.ps1
```diff
@@ -98,25 +98,55 @@ Write-Log 'Installing external-dns' -Console
 $externalDnsConfig = Get-ExternalDnsConfigDir
 (Invoke-Kubectl -Params 'apply' , '-k', $externalDnsConfig).Output | Write-Log
 
-Write-Log 'Installing Traefik Ingress controller' -Console
-$traefikYamlDir = Get-TraefikYamlDir
+Write-Log "Preparing kusomization with $controlPlaneIp as an external IP for traefik service" -Console
```
mahenarayan, line 101: Typo kusomization -> kustomization.
Rest looks good, I like the approach of patch using kustomize. Should be used in all addons as well.

mahenarayan (approved):

addons/traefik/Enable.ps1
artur-pusztai: Thanks for your feedback, Mahesh

PR #654, opened 2024-07-24
refactor(addons): renaming of addons in order to indicate more clear the purpose of the addon

achimkist:
refactor(addons): renaming of addons in order to indicate more clear the purpose of the addon


mahenarayan:
@achimkist Few observations:

- Similar to `ingress`, `gateway-api` can have implementation nginx, because that's what we support at the moment. This is not clear from current list of addons.
- `storage` also has `smb` as the implementation which is not clear from the current list of addons.

PR #657, opened 2024-07-25
fix: Re-implemented Vswitch caching feature after refactoring changes

abhitoshj:


lib/modules/k2s/k2s.node.module/linuxnode/setup/control-plane-node.module.psm1
```diff
@@ -255,9 +255,14 @@ function Start-ControlPlaneNodeOnNewVM {
 
             # connect VM to switch
             Connect-KubeSwitch
-        } 
+        }
+        # route for VM
+        Write-Log "Remove obsolete route to $ipControlPlaneCIDR"
```
mahenarayan, line 260: @abhitoshj Any reason to re-apply routes from the previous place?

mahenarayan (approved):
Looks good to me.

PR #708, opened 2024-09-02
feat(security): dashboard enable updates its configuration

artur-pusztai:
### Motivation

Some addons configuration (example: dashboard) depend on whether other addons are enables or not (example: which ingress if any, security)

We need a way to keep the configurations of all addons consistent with each other.

We try this out for `dashboard` addon, so that we can enable and disable the addons liste below in any order we want:

- `dashboard`
- `ingress nginx`  or `ingress traefik`
- `security`

And after each change, the dashboard configuration is always consistent with the current state of the other addons.

### Modifications

- removed the ingress endpoint `k2s-dashboard.cluster.local`, keeping only `k2s.cluster.local/dashboard/`.
- `dashboard` can be anabled several times now. When it is enabled, it will evaluate its dependencies (ingress, security) and choose the proper kustomization to be applied.
- At the end of the Enabling script, the `ingress` addons and the `security` addon triggers the `Update-IngressForAddons`
- `security` can be enabled now with traefik as well, there was a bug causing the oauth-proxy to crash in this case, now it is working.

### Verification

Tests adapted to reflect the new behavior of being able to re-enable `dashboard`.


achimkist (approved):
looks good 👍 FYI: I'm taking care of the other "dashboard" addons (logging, monitoring and new rollout addon)

PR #723, opened 2024-09-10
feat(vfp): 677 pipeline to build dll

krotz-dieter:
feat(vfp): 677 pipeline to build dll


krotz-dieter:
feat(vfp): 677 pipeline to build dll (#723)

Issue #782, opened 2024-10-04
Allow usage of  image-id to perform tag and push operations


abhitoshj:
**Is your feature request related to a problem? Please describe.**
Currently, some image commands expect the user to only provide the name of the image to perform the following operations:
- tag
- push

**Describe the solution you'd like**
It should be possible to perform these operations using image id.
E.g. k2s image tag --id <some-tag> --name <image-name> -o

jakobkol:
Since `k2s image tag` creates a copy of the image with a new name, the imageId stays the same. 
If we then use `k2s image push --id` and reference the image we tagged, it will create an error since there are "two" images with the same Id.

![image](https://github.com/user-attachments/assets/607a024e-9509-4656-a3ce-8f1ad4d10402)
![image](https://github.com/user-attachments/assets/729a97e6-804c-46d3-8e9a-139daca4c30b)

How should this be handled? @abhitoshj

abhitoshj:
@jakobkol Since usage of imageId results in ambiguity in the case that you mention, we can either 
- take the latest image and attempt to push it. This could be determined with the age of the image.
- Notify the user regarding ambiguity and that the user must explicitly use the image name for push or delete one of the images.

According to me, the first option may not always result in a desired behaviour for the user. Image, that the user did not intend to push, may get pushed.

Hence, I would prefer the second option.

However, I just realised this. Will it be possible to check the behaviour for when the *id* is used for **k2s image rm --id** ? Maybe this also needs similar handling.

jakobkol:
@abhitoshj I've just checked for `k2s image rm --id` and it will delete all images that have the specified id. I guess that shouldn't happen? 

As you already mentioned, notifying the user to explicitly use the image name seems like a good option. Should I implement it this way for **push** and while I am at it also for **rm** if there is more than one image?

abhitoshj:
@jakobkol Thanks for checking. I think we can implement for now only for **push**. I think the behaviour is fine for **rm**.

achimkist:
tests failing

PR #838, opened 2024-11-04
782 allow usage of image id to perform tag and push operations

jakobkol:
related to #782 

fixes CLI build


achimkist (changes requested):
change the PR name to something like "fix(tests): fix setup related image tests"

achimkist (approved):

PR #1388, opened 2025-11-09
Common proxy handling

abhitoshj:
- Defined a new parameter for providing hosts that need to be excluded from proxy.

- Restructed communication in such a way that all k2s windows services and kubemaster send their requests to httpproxy, which decides if the request has to be forwarded to the forward-proxy or not. 

- Defined new commands with which it is possible to manage proxy settings.


abhitoshj:
incorrect pr

PR #1434, opened 2025-11-19
1271 windows exporter moved from core part to addons who need it  (Monitoring/metrics) 

### LINKED ISSUE #1271: Move windows exporter to addons who need it
Currently windows exporter is used as an windows service as part of the K2s core parts.
Since it is used from metrics and monitoring addons windows exporter should convert in an host process container which is coming from both addons, if already there then it shall not be instantiated.
This can be handled within the Update parts of these two addons.
Maybe also reuse the common part of the addons.

Shashank-Wagle23:
Fixes #1271 

### Motivation

Move the windows-exporter installation from the core part i.e. during k2s installation to available on demand during monitoring/metrics addons are enabled.

### Modifications

Windows-exporter is now enabled on demand only during monitoring or metrics addons are enabled. Default enabling during k2s installation is now removed.

### Verification

Checked basic use cases. If windows-exporter is enabled on demand while monitoring or metrics addon are enabled and removed while both addons are disabled but available if one of them is enabled. Also checked if the windows metrics are being scraped by Prometheus

### DIFF
--- a/build/bom/merge/k2s-static.json
+++ b/build/bom/merge/k2s-static.json
@@ -258,9 +258,9 @@
       "type": "application"
     },
     {
-      "group": "https://download.docker.com/win/static/stable/x86_64/docker-28.5.2.zip",
+      "group": "https://download.docker.com/win/static/stable/x86_64/docker-29.0.4.zip",
       "name": "dockerd",
-      "version": "v28.5.2",
+      "version": "v29.0.4",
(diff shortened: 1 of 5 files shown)


Shashank-Wagle23:
Also checked the exporting addons monitoring and metrics for offline use case. Windows-exporter image is exported as a win_tar file

Issue #2560, opened 2026-06-08
Move `clusterip-webhook` certificate generation into the Deployment


krotz-dieter:
# Move `clusterip-webhook` certificate generation into the Deployment and use `emptyDir` for runtime TLS files

## Problem

The `clusterip-webhook` currently creates its TLS certificate through separate Kubernetes Jobs:

- `clusterip-webhook-certgen-create`
- `clusterip-webhook-certgen-patch`

This works for initial deployment, but certificate lifecycle handling is separated from the webhook workload itself. If the certificate expires or becomes invalid, the recovery path is not obvious and is not integrated with the existing `k2s system certificate` command set.

The webhook Deployment currently consumes the generated certificate from the `clusterip-webhook-tls` Secret and mounts it as:

- `/certs/tls.crt`
- `/certs/tls.key`

A better lifecycle model would be to generate the certificate as part of the webhook Pod startup. This way, recreating the Pod can also recreate a fresh certificate with a known validity period.

## Proposal

Replace the separate certgen create/patch Jobs for `clusterip-webhook` with certificate generation handled by an init container in the `clusterip-webhook` Deployment.

The Deployment should use an `emptyDir` volume for runtime certificate files.

The init container should:

1. Generate a self-signed TLS certificate valid for one year.
2. Write `tls.crt` and `tls.key` directly into the shared `emptyDir` volume.
3. Patch the webhook configuration `caBundle` with the generated certificate.
4. Exit successfully only after certificate generation and webhook patching completed.

The main webhook container should continue to use:

- `--tls-cert=/certs/tls.crt`
- `--tls-key=/certs/tls.key`

but `/certs` should be backed by the shared `emptyDir` volume populated by the init container, not by a required Secret volume.

## Why `emptyDir`

Using `emptyDir` avoids a startup dependency loop.

If the Pod mounts `clusterip-webhook-tls` as a required Secret and that Secret is missing, Kubernetes will not start the Pod. That also means the init container cannot run to create or repair the Secret.

With `emptyDir`:

- Kubernetes can always start the Pod.
- The init container can generate fresh certificate files before the webhook container starts.
- The webhook container receives the generated files from the shared volume.
- Recreating the Pod recreates the runtime certificate files.
- The certificate lifecycle becomes tied to the webhook Pod lifecycle.

A Secret can still be created or updated by the init container as an optional compatibility/debugging artifact, but the running webhook should not depend on that Secret being present before Pod startup.

[...]

PR #2571, opened 2026-06-11
#2560 implement init container for TLS cert generation

GermanGLLopez:
- Added an init container to the clusterip-webhook deployment that generates a self-signed TLS certificate on pod startup.
- Updated deployment configuration to remove the previous certgen jobs and utilize an emptyDir volume for certificate storage.
- Modified webhook configuration to allow the init container to patch the caBundle directly.
- Removed legacy certgen job and patch job logic from deployment scripts.
- Updated related RBAC permissions to accommodate changes in certificate management.
- Enhanced the renewal process for the webhook certificate to trigger a rollout restart, ensuring a fresh certificate is generated.
- Introduced a new Go module for certificate generation and patching of the MutatingWebhookConfiguration.

Uses #2560


krotz-dieter (commented):
Here is my review analysis:

---

## PR Review: #2571 — Init container for TLS cert generation

The overall approach is sound (replacing certgen Jobs with an init container), but there are a few real issues:

---

### 🐛 Bug: Race condition / missing conflict retry in `patchWebhookCABundle`

**File:** `k2s/cmd/clusterip-webhook/certgen.go`, lines 172–192

The function does a `Get` followed by an `Update` with no conflict-detection retry:

```go
webhookCfg, err := clientset.AdmissionregistrationV1().MutatingWebhookConfigurations().Get(...)
// ...
_, err = clientset.AdmissionregistrationV1().MutatingWebhookConfigurations().Update(...)
```

During a rolling update, two pods are briefly alive simultaneously. Both init containers will run concurrently, read the same `resourceVersion`, and race to `Update`. The second write will fail with a `409 Conflict` error — causing that init container to exit non-zero, triggering a pod restart loop (CrashLoopBackOff).

**Fix:** Retry on `k8s.io/apimachinery/pkg/api/errors.IsConflict`. This is the standard Kubernetes optimistic-concurrency pattern:

```go
for {
    // Get, mutate, Update ...
    _, err = clientset.AdmissionregistrationV1().MutatingWebhookConfigurations().Update(...)
    if errors.IsConflict(err) {
        continue // re-read and retry
    }
    return err
}
```

---

### 🔒 Security regression: `readOnlyRootFilesystem: false` in init container

**File:** `lib/manifests/clusterip-webhook/deployment.yaml`

```yaml
securityContext:
  readOnlyRootFilesystem: false   # <-- regression
```

The old certgen Jobs used `readOnlyRootFilesystem: true`. Writes to volume mounts (like the `/certs` emptyDir) are permitted even when the container's root filesystem is read-only. The init container only writes to `/certs/tls.crt` and `/certs/tls.key`, both on the emptyDir mount, so `readOnlyRootFilesystem: true` is fully compatible here. This should be reverted to `true` to match the hardening posture of the removed Jobs.

---

### 🐛 Latent bug: `writeCertFiles` only creates the directory for `certPath`, not `keyPath`

**File:** `k2s/cmd/clusterip-webhook/certgen.go`, lines 148–163

```go
if dir := filepath.Dir(certPath); dir != "" {
    if err := os.MkdirAll(dir, 0755); err != nil { ... }
}
// keyPath directory is never created
if err := os.WriteFile(keyPath, keyPEM, 0600); err != nil { ... }
```

If `certPath` and `keyPath` are ever in different directories (e.g., via misconfiguration or future refactoring), the key write will fail with `no such file or directory`. The fix is to also call `MkdirAll(filepath.Dir(keyPath), ...)` before writing the key file.

---

### ℹ️ Minor: `Restore-ClusterIPWebhook` now unconditionally re-applies `webhook-config.yaml` (which resets `caBundle: ""`)

The previous code deliberately skipped re-applying the MWC if it already existed (to avoid clearing the caBundle). This guard is now removed because the init container re-patches the caBundle on rollout restart. This is architecturally correct, but there is a brief window between `kubectl apply -f webhook-config.yaml` (which resets `caBundle: ""`) and the init container completing where the webhook will reject admission requests. This is acceptable during a restore but worth documenting in the function's comment so future maintainers don't re-add the guard unnecessarily.

krotz-dieter (commented):
Here is my review analysis:

---

## PR #2571 Review: Init Container for TLS Cert Generation

The PR replaces two ephemeral certgen Jobs + a Kubernetes Secret with a self-contained init container using the webhook binary itself. The approach is architecturally sound, but there are three issues worth addressing:

---

### 🔴 Security: Permanent privilege escalation via merged RBAC

**File:** `lib/manifests/clusterip-webhook/rbac.yaml`

The old design used a separate `clusterip-webhook-certgen` ServiceAccount with its own ClusterRole granting `get/update` on `mutatingwebhookconfigurations`. This permission was scoped to the short-lived certgen Jobs.

In the new design, those permissions are merged into the `k2s:clusterip-webhook` ClusterRole used by the main webhook runtime:
```yaml
- apiGroups: ["admissionregistration.k8s.io"]
  resources: ["mutatingwebhookconfigurations"]
  verbs: ["get", "update"]
```

Because init containers share a ServiceAccount with the main container, this was required — but it means the long-running webhook pod **permanently** holds cluster-wide `update` on `mutatingwebhookconfigurations`. A compromised webhook process could silently modify or disable any `MutatingWebhookConfiguration` in the cluster. This is a meaningful ambient privilege the original design deliberately avoided. Consider whether an additional admission control or audit policy should flag this.

---

### 🟡 No backoff between conflict retries

**File:** `k2s/cmd/clusterip-webhook/certgen.go`, lines ~200–215 (`patchWebhookCABundle`)

```go
for attempt := 0; attempt < maxConflictRetries; attempt++ {
    // Get, modify, Update...
    if k8serrors.IsConflict(err) {
        slog.Info("Conflict updating webhook configuration, retrying", "attempt", attempt+1)
        continue   // ← no sleep
    }
}
```

There is no `time.Sleep` (or jitter) between retries. If `replicas > 1` and multiple pods start concurrently during a rolling update, all pods may hit a conflict simultaneously, re-fetch, and conflict again — exhausting all 5 retries within milliseconds. A short backoff with jitter (e.g., `time.Sleep(time.Duration(attempt+1)*100*time.Millisecond + jitter)`) would prevent thundering-herd exhaustion here.

---

### 🟡 Missing error handling in `Invoke-WebhookCertificateRenewal`

**File:** `lib/scripts/k2s/system/certificate/renew.ps1`, lines ~253 and ~259

```powershell
$restartOutput = (Invoke-CmdOnControlPlaneViaSSHKey $restartCmd).Output      # no -IgnoreErrors
$statusOutput  = (Invoke-CmdOnControlPlaneViaSSHKey $statusCmd).Output       # no -IgnoreErrors
```

Both calls omit `-IgnoreErrors`, unlike most other `Invoke-CmdOnControlPlaneViaSSHKey` calls in the codebase and unlike the analogous code in `system_linux.go`. If `kubectl rollout restart` or `kubectl rollout status` fails (transient API error, kubeconfig issue, etc.), an uncaught exception will propagate out of `Invoke-WebhookCertificateRenewal`, bypassing the caBundle validation entirely. The `.Success` field should be checked, or `-IgnoreErrors:$true` should be added with explicit failure handling.

krotz-dieter (commented):
Here is my review analysis:

---

## Review: PR #2571 — Init container for TLS cert generation

Overall this is a well-structured refactor. I found several real issues worth addressing:

---

### 🐛 Bug: `checkCmd.Run()` silently swallows non-"not-found" errors
**File:** `k2s/internal/provider/system_linux.go`, lines ~83–88

```go
checkCmd := exec.Command("kubectl", "get", "deployment", "clusterip-webhook",
    "-n", "k2s-webhook", "--no-headers")
if err := checkCmd.Run(); err != nil {
    slog.Info("[System] clusterip-webhook deployment not found - skipping webhook cert renewal")
    return nil
}
```

Any `kubectl` failure — network error, RBAC issue, API server unreachable — is silently treated as "deployment not found" and returns `nil`, masking real errors. Certificate renewal is skipped without any indication of the real problem. At minimum, log at `Warn` with the actual error; ideally distinguish exit code 1 ("not found") from other errors.

---

### 🐛 Bug: Missing `-IgnoreErrors` and `-Timeout` on caBundle validation call
**File:** `lib/scripts/k2s/system/certificate/renew.ps1`, function `Invoke-WebhookCertificateRenewal` (~line 268)

```powershell
$caBundle = (Invoke-CmdOnControlPlaneViaSSHKey $caBundleCmd).Output
```

Unlike every other call in this function, this one is missing `-IgnoreErrors:$true` and `-Timeout`. If the SSH command fails or times out, this can throw an unhandled exception that propagates out of the `try` block and disrupts the entire certificate renewal flow. Should be:

```powershell
$caBundleResult = Invoke-CmdOnControlPlaneViaSSHKey -CmdToExecute $caBundleCmd -Timeout 30 -IgnoreErrors:$true
$caBundle = ($caBundleResult.Output | Out-String).Trim()
```

---

### ⚠️ Security regression: Permanent cluster-scoped `update` on `mutatingwebhookconfigurations` granted to the long-running webhook SA
**File:** `lib/manifests/clusterip-webhook/rbac.yaml`

The old design used a dedicated, short-lived ServiceAccount for caBundle patching. This PR merges that permission into the main `clusterip-webhook` ServiceAccount, which runs indefinitely. A compromised webhook process can now permanently modify any `MutatingWebhookConfiguration` in the cluster — a significant privilege escalation path. The in-code comment acknowledges this but frames it as "architecturally required." It is not: an alternative is to use a dedicated init container image with its own ServiceAccount (as before), or use `kubectl patch` with a scoped projected token mounted only to the init container. This trade-off deserves an explicit risk acceptance from the security team, not just a code comment.

---

### ⚠️ Admission webhook outage window during restore
**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`, Step 2

The new code unconditionally re-applies `webhook-config.yaml`, which resets `caBundle: ""`. The init container only repopulates it after the pod rollout completes (~30–120s later). If the webhook's `failurePolicy: Fail`, **all admission requests cluster-wide will be rejected** during this window (blocking pod creation, configmap updates, etc.). The old code intentionally avoided this by skipping the `webhook-config.yaml` apply when the MWC already existed. The new comment says this is "acceptable during restore/update" — but that's only true if `failurePolicy: Ignore`, which should be verified and documented explicitly.

---

### ⚠️ No replica count enforcement for concurrent init-container issue
**File:** `lib/manifests/clusterip-webhook/deployment.yaml`

The RBAC comment states "The webhook deployment runs with a single replica." However, `replicas: 1` is not enforced in the deployment spec (not visible as added in the diff). During a rolling update, two pods can overlap: old pod serving with its cert while new pod's init container patches `caBundle` to its own new CA — causing TLS validation failures on in-flight requests against the old pod. Either add `replicas: 1` explicitly or use `RollingUpdate` with `maxSurge: 0, maxUnavailable: 1`.

PR #2573, opened 2026-06-12
#2478 feat(storage/smb): add opt-in SMB 3.1.1 POSIX extensions support

### LINKED ISSUE #2478: Support SMB 3.1.1 POSIX extensions in storage smb addon
### Description

The current `storage smb` addon provides SMB-backed shared storage and dynamically generated SMB CSI `StorageClass` resources, but it does not support SMB 3.1.1 POSIX extensions today.

Current behavior:

- Linux client mounts are hardcoded to SMB 3.0/3.x style options:
  - Windows-hosted SMB path uses `vers=3.0` in `addons/storage/smb/module/Smb-share.module.psm1`
  - Linux/Samba-hosted path uses `vers=3` in `addons/storage/smb/module/Smb-share.module.psm1`
- The SMB CSI `StorageClass` template uses permission-emulation-oriented mount options such as `dir_mode`, `file_mode`, `uid`, `gid`, `noperm`, `mfsymlinks`, `cache=strict`, and `noserverino` in `addons/storage/smb/manifests/base/storage-classes/template_StorageClass.yaml`
- `addons/storage/smb/config/SmbStorage.json` does not expose SMB dialect, POSIX extension, or inode behavior settings.

This means workloads that rely on stronger POSIX-like filesystem semantics, such as correct mode handling, ownership behavior, symlink behavior, hard links, or stable inode semantics, cannot explicitly request SMB 3.1.1 POSIX behavior through the addon.

### Goal

Add opt-in support for SMB 3.1.1 POSIX extensions in the `storage smb` addon while preserving the existing default behavior.

The feature should be configurable per storage entry, so existing users continue to get the current SMB behavior unless they explicitly request POSIX mode.

### Proposed Scope

Add optional fields to the SMB storage config, for example:

- `smbDialect`: requested SMB dialect, such as `3`, `3.0`, `3.1.1`, or `auto`
- `enablePosixExtensions`: enables POSIX-extension-oriented mount/server configuration
- `useServerInode`: optional switch if POSIX mode requires server-side inode tracking instead of the current `noserverino` behavior

Update the addon implementation to:

- Normalize and validate these fields in `Get-StorageConfigFromRaw`
- Generate CIFS mount options dynamically instead of hardcoding `vers=3.0` and `vers=3`
- Generate SMB CSI `StorageClass` `mountOptions` dynamically
- Configure Samba-specific POSIX settings when `SmbHostType=linux` and POSIX mode is enabled
- Preserve current default behavior when POSIX mode is not enabled
- Warn clearly when POSIX extensions are requested with `SmbHostType=windows`, because Windows SMB servers can support SMB 3.1.1 dialect negotiation but are not expected to provide SMB POSIX extension semantics

### Important Caveats

Windows-hosted SMB shares should not be advertised as POSIX-compliant unless runtime validation proves the requested semantics are actually negotiated and working.

The current `noserverino` option was added to avoid data corruption risks. True POSIX behavior may require server-side inode behavior for hard links and stable inode reporting, so this needs careful validation before changing any defaults.

SMB POSIX extension support depends on the Linux kernel CIFS client, `cifs-utils`, Samba/server support, and filesystem capabilities on the Samba host.

### Acceptance Criteria

- [ ] Existing `storage smb` behavior remains unchanged by default.
- [ ] Users can opt into SMB 3.1.1/POSIX mode per `SmbStorage.json` entry.
- [ ] Generated host fstab entries use the requested SMB dialect and POSIX options only when configured.
- [ ] Generated SMB CSI `StorageClass` resources include the requested mount options only when configured.
- [ ] Linux/Samba-hosted shares configure and validate required Samba settings for POSIX mode.
- [ ] Windows-hosted shares emit a clear warning if POSIX extensions are requested.
- [ ] Unit tests cover config parsing, defaults, validation, and generated mount options.
- [ ] E2E or acceptance coverage verifies at least one POSIX-enabled Linux/Samba-hosted storage class.
- [ ] Documentation explains the feature, limitations, and rollback path.

### Out Of Scope

- Changing the default SMB mount behavior.
- Upgrading the SMB CSI driver unless validation proves it is required.
- Claiming POSIX compliance for Windows workloads or Windows-hosted SMB shares without runtime proof.
- Migrating existing SMB share data or changing existing storage class names automatically.

Saurabh-Kumar24:
Fixes #2478

## Motivation

The `storage smb` addon did not support SMB 3.1.1 POSIX extensions. Workloads requiring POSIX semantics (correct ownership, permissions, symbolic links) on a Linux-hosted (Samba) SMB share could not be served correctly. This PR adds opt-in POSIX extension support, configurable per storage entry, without changing the default behavior of existing shares.

## Modifications

- **Per-entry POSIX config** in `SmbStorage.json`: `smbDialect` (default `auto` → `3.0`), `enablePosixExtensions` (default `false`), `useServerInode` (default `false`).
- **Mount options** (`Smb-posix.module.psm1`): when POSIX is enabled, drops `noperm`, applies `dir_mode=0777,file_mode=0777,uid=1001,gid=1001,mfsymlinks,cache=strict`, omits `noserverino` when `useServerInode` is set, and pins `vers=` only for non-`auto` dialects.
- **Samba VFS**: configures `vfs objects = streams_xattr` + `store dos attributes = no` for POSIX shares, and installs the `samba-vfs-modules` Debian package (which ships `streams_xattr.so`) on the Linux control-plane host via the offline-aware installer — only downloaded when not already cached.
- **POSIX negotiation** hardened against the smbd-restart race with a bounded `testparm` retry (warn, not fail).
- **Docs**: addon `README.md` documents the POSIX fields, the `samba-vfs-modules`/offline-caching behavior, limitations, rollback, and now includes concrete StorageClass/PVC deployment examples plus a POSIX-specific example (`-t linux`, `SmbStorage.json` entry, disable-before-edit config workflow). The enable banner now points users to the addon README instead of the internal e2e workloads folder.

## Verification

- Unit tests: `Smb-share.module.unit.tests.ps1` (123/123) and `Smb-posix.module.unit.tests.ps1` (POSIX mount options, dialect defaults, `streams_xattr` config) — all green.
- E2E: `storage-smb_test.go` adds an "SMB 3.1.1 POSIX Extensions" spec that enables a Linux-hosted POSIX share, asserts `vers=3.1.1` and absence of `handletimeout`/`noperm`, and verifies symbolic-link traversal over the share.
- Live cluster validation: POSIX Linux-host share mounts successfully, both StorageClasses created, `dpkg -l samba-vfs-modules` = `ii`, `streams_xattr.so` present.
- e2e tests passed for the `addon-storage-smb` usecase.

## Acceptance Criteria

- [x] AC1: `smbDialect` `auto`/empty resolves to `vers=3.0`.
- [x] AC2: POSIX-enabled mounts drop `noperm`/`handletimeout` and configure `streams_xattr`.
- [x] AC3: `useServerInode` omits `noserverino` from mount options.
- [x] AC4: e2e tests passed for the storage smb usecase.


krotz-dieter (commented):
Now I have enough context to write the review. Let me compile the findings:

Here is the code review:

---

## Review of PR #2573 — SMB 3.1.1 POSIX Extensions

### Bug: `samba-vfs-modules` may not be installed when needed

**`addons/storage/smb/module/Smb-share.module.psm1`, `New-SmbHostOnLinuxIfNotExisting`**

The `samba-vfs-modules` package is only installed if `$Config.EnablePosixExtensions` is true **on the specific config entry** that first triggers host creation. Since the function name implies it guards against re-running ("IfNotExisting"), later share entries with POSIX enabled won't cause it to re-run and install the package. If a user's first share has `enablePosixExtensions: false` and a subsequent share has `enablePosixExtensions: true`, the VFS modules will be absent and `vfs objects = streams_xattr` will silently fail to load. The fix is to check if *any* entry across the full config set has POSIX enabled when deciding what packages to install.

---

### Behavioral regression: `vers=3` → `vers=3.0` in Linux host fstab

**`addons/storage/smb/module/Smb-share.module.psm1`, `New-SharedFolderMountOnLinuxHost`**

The original fstab entry used `vers=3` (instructs the kernel CIFS driver to negotiate *any* SMB 3.x dialect). With `smbDialect: auto`, `Get-FstabVersionOption` returns `vers=3.0`, pinning to exactly SMB 3.0. These are not equivalent: `vers=3` permits the kernel to negotiate up to 3.1.1, while `vers=3.0` forbids it. Existing Linux host mounts will be downgraded from flexible dialect negotiation to pinned 3.0.

---

### Shell injection via `$ShareName` in `testparm` command

**`addons/storage/smb/module/Smb-posix.module.psm1`, `Test-SambaPosixNegotiation` (~line 57)**

```powershell
$cmd = "sudo testparm -s --section-name '$ShareName' 2>/dev/null"
```

`$ShareName` comes from `$Config.LinuxShareName`, which is read from `SmbStorage.json`. A share name containing a single quote (e.g., `k2s'share`) breaks out of the shell single-quote context. While this is a local config file, not a network-facing input, any config-tampering scenario (supply chain, misconfig) leads to arbitrary command execution on the control plane via SSH. The value should be validated or shell-escaped before interpolation.

---

### `Get-SmbHostType` called during config deserialization

**`addons/storage/smb/module/Smb-share.module.psm1`, `Get-StorageConfigFromRaw` (~line 1367)**

```powershell
if ($enablePosix -and (Get-SmbHostType) -eq 'windows') {
    Write-Log "[SMB] Warning: ..."
}
```

`Get-StorageConfigFromRaw` is a config-parsing function that may be called early (e.g., during `Get-Status`). Calling `Get-SmbHostType` here ties a potentially expensive or failable operation (inferring host type) to pure config reading. If the host is not yet reachable, this silently swallows an error instead of warning at setup time. The warning belongs in the `Enable-SmbShare` flow after the host type is known.

---

### Regex replacement with unsanitized substitution string

**`addons/storage/smb/module/Smb-share.module.psm1`, `New-StorageClassManifest` (~line 675)**

```powershell
$manifestContent = $templateContent -replace '(?m)^MOUNT_OPTIONS\s*$', $mountOptsYaml
```

PowerShell's `-replace` operator interprets `$1`, `$2`, etc. in the *replacement* string as back-references. `$mountOptsYaml` is constructed from `$mountOpts` entries. Although current values are safe strings, this is a latent bug: any mount option containing a `$` followed by a digit (e.g., a path like `/mnt/$1-share`) will silently corrupt the generated YAML. Use `-replace` with a literal substitution or sanitize `$mountOptsYaml` before use (e.g., replace `$` with `$$` in the replacement string).

---

### Minor: Dead `else` branch in fstab suffix calculation

**`addons/storage/smb/module/Smb-share.module.psm1`, lines ~288 and ~502**

```powershell
$fstabVersSuffix = if ($fstabVersOption) { ",$fstabVersOption" } else { '' }
```

`Get-FstabVersionOption` always returns a non-empty string (returns `"vers=3.0"` for `auto`). The `else ''` branch is unreachable. This is harmless now but misleads readers into thinking a no-`vers` fstab entry is possible.

Saurabh-Kumar24:
All critical findings addressed with commit 280da24 above. Workflow passed.

PR #2575, opened 2026-06-12
feat(webhook): enhance IP address tracking and update RBAC permissions

GermanGLLopez:
Uses #2560

### DIFF
--- a/k2s/cmd/clusterip-webhook/main.go
+++ b/k2s/cmd/clusterip-webhook/main.go
@@ -347,6 +347,17 @@ func (h *WebhookHandler) getUsedClusterIPs() (map[string]bool, error) {
 		}
 	}
 
+	// Also include IPs tracked by the API server's internal allocator.
+	// This catches stale bitmap entries and IPs allocated during webhook downtime.
+	ipAddresses, err := h.clientset.NetworkingV1().IPAddresses().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		slog.Warn("Failed to list IPAddress objects, relying on service list only", "error", err)
+	} else {
+		for _, ipa := range ipAddresses.Items {
+			used[ipa.Name] = true
+		}
+	}
+
 	return used, nil
 }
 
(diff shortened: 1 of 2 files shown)


krotz-dieter (commented):
Working from the provided diff directly.

---

## Code Review: PR #2575

### `k2s/cmd/clusterip-webhook/main.go`

**Kubernetes version compatibility gap**

The new code uses `NetworkingV1().IPAddresses()`, which refers to the `networking.k8s.io/v1` resource promoted to GA in Kubernetes **1.31**. On clusters running **1.27–1.30**, `IPAddress` exists only as `networking.k8s.io/v1alpha1`. The list call will return a "not found" error on those clusters, the warning fires, and the supplemental tracking silently does nothing — defeating the stated purpose ("catches stale bitmap entries and IPs allocated during webhook downtime").

If K2s supports clusters older than 1.31, this should either:
- Try `NetworkingV1alpha1().IPAddresses()` as a fallback when the `v1` call fails (distinguishing "resource not found" from other errors), or
- Document the minimum supported Kubernetes version as 1.31+ and enforce that elsewhere.

As written, the degraded path is safe (no crash, service-list fallback still works), but the feature is silently ineffective on pre-1.31 clusters.

**Error handling philosophy inconsistency**

All other `List` calls in `getUsedClusterIPs` (nodes, pods, services, etc.) return the error to the caller, which surfaces it as an admission webhook failure. The new `IPAddresses` call swallows the error as a warning. This is intentional per the comment, but it means a misconfigured RBAC or a transient API server error here will silently pass instead of alerting. At minimum the warning message should make it clear this could indicate a permission problem (e.g., missing RBAC).

### `lib/manifests/clusterip-webhook/rbac.yaml`

The new rule is correctly scoped (`list` only). However, for consistency with the version gap noted above: if fallback to `v1alpha1` is added in the Go code, a corresponding RBAC rule for `networking.k8s.io/v1alpha1` `ipaddresses` would also be needed. The current RBAC rule does **not** cover `v1alpha1`.

---

## Files Reviewed
- ✅ k2s/cmd/clusterip-webhook/main.go
- ✅ lib/manifests/clusterip-webhook/rbac.yaml

PR #2583, opened 2026-06-14
Headlamp plugin integration changes

### LINKED ISSUE #1805: Replace kubernetes dashboard with headlamp
Official kubernetes dashboard has reached it's end: https://github.com/kubernetes-retired/dashboard,
please use https://github.com/kubernetes-sigs/headlamp as an alternative for the k2s addon dashboard.

vvkrishnakanth9:
Fixes #1805 

Motivation
Migrated Headlamp plugins to upstream archives and removed unnecessary plugin build infrastructure.

Modifications
Updated the flow for sync between headlamp and plugins.

Verification
Validated unit tests, plugin synchronization, addon lifecycle flow


krotz-dieter (commented):
## PR Review: Headlamp plugin tgz model (#2583)

This PR introduces a capability-based Headlamp plugin injection framework using Kubernetes init-containers. The architecture is solid (offline-first, capability-detected, strategic-merge patched), but there are a few real bugs worth addressing.

---

### 🐛 Bug: `Sync-HeadlampPlugins` unconditionally waits 300s even when no patch is applied

**`addons/dashboard/dashboard.module.psm1`** — `Sync-HeadlampPlugins` (near end of file):

```powershell
Apply-HeadlampPluginPatch -InitContainers $initContainers
# Observability only: wait for the rollout...
Wait-ForHeadlampRollout | Out-Null   # ← always called
```

`Apply-HeadlampPluginPatch` returns early with no kubectl call when state is already up to date (`"Plugin init-containers already up to date; no patch required"`). But `Sync-HeadlampPlugins` still calls `Wait-ForHeadlampRollout` afterwards, blocking the caller for up to **300 seconds** via `kubectl rollout status --timeout=300s` even when no rollout was triggered.

Since `Sync-HeadlampPlugins` is called from **every capability-related addon's** Enable/Disable/Update scripts (ingress nginx, traefik, nginx-gw, monitoring, fluxcd, security), this means those addon operations will silently hang for up to 5 minutes whenever the plugin state was already converged.

**Fix:** Have `Apply-HeadlampPluginPatch` return `$true` when a patch was actually issued, and condition the `Wait-ForHeadlampRollout` call on that return value:

```powershell
$patched = Apply-HeadlampPluginPatch -InitContainers $initContainers
if ($patched) {
    Wait-ForHeadlampRollout | Out-Null
}
```

---

### 🐛 Bug: `Get-RegisteredHeadlampPlugins` is missing from `Export-ModuleMember`

**`addons/dashboard/dashboard.module.psm1`** — final `Export-ModuleMember` line:

`Get-RegisteredHeadlampPlugins` is the canonical registry (documented as authoritative, validated by the parity test), but it is **not exported**. The parity test in `headlamp-plugin-images.unit.tests.ps1` works around this using:

```powershell
$registered = & (Get-Module $script:moduleName) { Get-RegisteredHeadlampPlugins }
```

This is fragile. Any future script that calls `Get-RegisteredHeadlampPlugins` directly (without the `& (Get-Module ...)` trick) will fail silently. Add it to `Export-ModuleMember`.

---

### 🐛 Bug: `Build-PluginFromSource` permanently mutates global npm proxy config

**`addons/dashboard/build/Build-HeadlampPluginImages.Methods.ps1`** — `Build-PluginFromSource`:

```powershell
if ($Proxy) { & npm config set proxy $Proxy; & npm config set https-proxy $Proxy }
& npm ci
```

`npm config set` writes to the **user's global `~/.npmrc`**. The proxy settings are never unset after the build. On developer machines this permanently changes their global npm proxy config. Use `npm_config_proxy`/`npm_config_https_proxy` environment variables instead (scoped to the process), or restore previous values in a `finally` block.

---

### ⚠️ Issue: `Expand-PluginTarball` payload detection is fragile for mixed-content archives

**`addons/dashboard/build/Build-HeadlampPluginImages.Methods.ps1`** — `Expand-PluginTarball`:

```powershell
$payloadDir = if ($roots.Count -eq 1 -and -not (Get-ChildItem -Path $tmp -File)) { $roots[0].FullName } else { $tmp }
```

If an archive has exactly one directory **and also** some loose files at the top level (non-standard but not impossible for upstream plugin releases), this falls through to `$tmp` and then `Copy-Item -Path (Join-Path $tmp '*')` copies both the directory and loose files. The downstream `Test-PluginBundleLayout` will catch missing `main.js`/`package.json` in the wrong place, but the error message won't be actionable. Adding a check for this case with a clear error message would save debugging time.

---

### ⚠️ Incomplete diff — 15 addon files cannot be fully reviewed

The diff is truncated: `addons/ingress/nginx-gw/Disable.ps1` ends at:
```
+if (Test-Path $dashboardModule
```
The diffs for `Enable.ps1`/`Disable.ps1`/`Update.ps1` of `ingress/nginx`, `ingress/traefik`, `ingress/nginx-gw`, `monitoring`, `rollout/fluxcd`, and `security` are not shown. These files presumably add `Sync-HeadlampPlugins` calls, which (given Bug #1) means all of these addon operations are affected by the 300s unnecessary wait.

---

### Minor: `headlamp-plugin-images.unit.tests.ps1` manifest parser matches comments

```powershell
$script:manifestImages = Get-Content -Path $script:manifestPath |
    Where-Object { $_ -match 'shsk2s\.azurecr\.io/headlamp-plugin-' }
```

This regex matches any line — including YAML comments — containing the pattern. A comment like `# was: shsk2s.azurecr.io/headlamp-plugin-flux:0.5.0` would cause a false parity failure. A tighter parse (e.g., requiring the line starts with optional whitespace then `-`) would be more robust.

---

## Files Reviewed

- ✅ addons/dashboard/Disable.ps1
- ✅ addons/dashboard/Enable.ps1
- ✅ addons/dashboard/README.md
- ✅ addons/dashboard/Update.ps1
- ✅ addons/dashboard/addon.manifest.yaml
- ✅ addons/dashboard/build/Build-HeadlampPluginImages.Methods.ps1
- ✅ addons/dashboard/build/Build-HeadlampPluginImages.ps1
- ✅ addons/dashboard/build/README.md
- ✅ addons/dashboard/build/headlamp-plugin-images.unit.tests.ps1
- ✅ addons/dashboard/build/headlamp-plugins.lock.json
- ✅ addons/dashboard/build/headlamp-plugins.lock.json.license
- ✅ addons/dashboard/build/plugins/cert-manager-0.1.0.tar.gz (binary)
- ✅ addons/dashboard/build/plugins/cert-manager-0.1.0.tar.gz.license
- ✅ addons/dashboard/build/plugins/flux-0.6.0.tar.gz (binary)
- ✅ addons/dashboard/build/plugins/flux-0.6.0.tar.gz.license
- ✅ addons/dashboard/build/plugins/prometheus-0.8.2.tar.gz (binary)
- ✅ addons/dashboard/build/plugins/prometheus-0.8.2.tar.gz.license
- ✅ addons/dashboard/dashboard.module.psm1
- ✅ addons/dashboard/dashboard.module.unit.tests.ps1
- ⚠️ addons/ingress/nginx-gw/Disable.ps1 (diff truncated — partially reviewed)
- ⚠️ addons/ingress/nginx-gw/Enable.ps1 (diff not shown)
- ⚠️ addons/ingress/nginx/Disable.ps1 (diff not shown)
- ⚠️ addons/ingress/nginx/Enable.ps1 (diff not shown)
- ⚠️ addons/ingress/traefik/Disable.ps1 (diff not shown)
- ⚠️ addons/ingress/traefik/Enable.ps1 (diff not shown)
- ⚠️ addons/monitoring/Disable.ps1 (diff not shown)
- ⚠️ addons/monitoring/Enable.ps1 (diff not shown)
- ⚠️ addons/monitoring/Update.ps1 (diff not shown)
- ⚠️ addons/rollout/fluxcd/Disable.ps1 (diff not shown)
- ⚠️ addons/rollout/fluxcd/Enable.ps1 (diff not shown)
- ⚠️ addons/rollout/fluxcd/Update.ps1 (diff not shown)
- ⚠️ addons/security/Disable.ps1 (diff not shown)
- ⚠️ addons/security/Enable.ps1 (diff not shown)
- ⚠️ addons/security/Update.ps1 (diff not shown)

krotz-dieter (commented):
Here is my review based on the full diff provided:

---

## Code Review: PR #2583 — Headlamp plugin tgz model

### 🐛 Bug: Staging directory collision on concurrent CI builds

**`addons/dashboard/build/Build-HeadlampPluginImages.ps1` (~line 81)**

```powershell
$staging = Join-Path $tempRoot ("hlplugin-" + (Get-Date -Format 'ddMMyyyy-HHmmss'))
```

Using a timestamp with second-level precision means two builds starting in the same second share a staging directory, causing them to corrupt each other's plugin bundles. Use `[Guid]::NewGuid().ToString('N')` for guaranteed uniqueness — the same pattern already used elsewhere in the codebase (e.g. `Expand-PluginTarball`, `Test-PluginImageLayout`).

---

### 🐛 Bug: `[Parameter(Mandatory = $true)]` on an explicitly unused parameter

**`addons/dashboard/build/Build-HeadlampPluginImages.Methods.ps1`, `Test-PluginImageLayout` (~line 261)**

```powershell
param(
    [Parameter(Mandatory = $true)] [string] $CraneExe,
    ...
)
$null = $CraneExe  # retained for signature compatibility; layout is validated via tar
```

Marking an unused parameter `Mandatory` is contradictory: every caller is forced to pass a dummy value that's silently discarded. The param should be `[string] $CraneExe = ''` (optional) or removed entirely — the call sites in `Build-HeadlampPluginImages.ps1` could be updated trivially.

---

### ⚠️ Potential: Manual JSON construction with no input sanitisation

**`addons/dashboard/dashboard.module.psm1`, `Build-PluginPatchJson` (~line 303)**

```powershell
$icParts += '{\"name\":\"' + $n + '\",\"image\":\"' + $img + '\",...}'
```

Plugin `Name` and `Image` values are interpolated directly into a hand-rolled escaped JSON string. A double-quote or backslash in either value would produce malformed JSON that `kubectl patch --patch-file` would reject with an opaque error. The current registry is safe because all values are hardcoded string literals, but there is no guard that enforces this invariant for future plugin registrations. Adding a `ValidatePattern` or a JSON-serialised intermediate (e.g. via `ConvertTo-Json`) on `$n`/`$img` would close the gap.

---

### ⚠️ Unnecessary rollout on disable

**`addons/dashboard/Disable.ps1`**

```powershell
Remove-HeadlampPluginPatch          # patches the live Deployment (triggers a rollout)
Uninstall-HeadlampViaHelm           # deletes the Deployment entirely
```

`Remove-HeadlampPluginPatch` patches and initiates a pod rollout that is immediately cancelled by the subsequent Helm uninstall. The patch adds no value here. If the dashboard Deployment happens to be in a degraded state (e.g. an init-container image pull is already failing), `Apply-HeadlampPluginPatch` will call `kubectl patch`, receive an error, and **throw**, blocking the rest of `Disable.ps1` including `Uninstall-HeadlampViaHelm`. Calling `Remove-HeadlampPluginPatch` after the Helm uninstall (or simply removing it from `Disable.ps1` since the Deployment is about to be deleted) avoids both the unnecessary rollout and the failure-blocking risk.

---

### ℹ️ Diff is incomplete — 16 files cannot be reviewed

The following files appear in the PR's changed-files list but **no diff is shown**:

- `addons/ingress/nginx-gw/Disable.ps1`, `Enable.ps1`
- `addons/ingress/nginx/Disable.ps1`, `Enable.ps1`
- `addons/ingress/traefik/Disable.ps1`, `Enable.ps1`
- `addons/monitoring/Disable.ps1`, `Enable.ps1`, `Update.ps1`
- `addons/rollout/fluxcd/Disable.ps1`, `Enable.ps1`, `Update.ps1`
- `addons/security/Disable.ps1`, `Enable.ps1`, `Update.ps1`

These are the addon scripts that are presumably being updated to call `Sync-HeadlampPlugins`. The README and `dashboard.module.psm1` both state that `Sync-HeadlampPlugins` **must** be called from every addon Enable/Disable that can affect a registered capability. If any of those files are missing the call, the corresponding plugin will not activate/deactivate correctly. They could not be verified from the provided diff.

---

## Files Reviewed

- ✅ addons/dashboard/Disable.ps1
- ✅ addons/dashboard/Enable.ps1
- ✅ addons/dashboard/README.md
- ✅ addons/dashboard/Update.ps1
- ✅ addons/dashboard/addon.manifest.yaml
- ✅ addons/dashboard/build/Build-HeadlampPluginImages.Methods.ps1
- ✅ addons/dashboard/build/Build-HeadlampPluginImages.ps1
- ✅ addons/dashboard/build/README.md
- ✅ addons/dashboard/build/headlamp-plugin-images.unit.tests.ps1
- ✅ addons/dashboard/build/headlamp-plugins.lock.json
- ✅ addons/dashboard/build/headlamp-plugins.lock.json.license
- ✅ addons/dashboard/build/plugins/cert-manager-0.1.0.tar.gz *(binary — content not reviewable)*
- ✅ addons/dashboard/build/plugins/cert-manager-0.1.0.tar.gz.license
- ✅ addons/dashboard/build/plugins/flux-0.6.0.tar.gz *(binary — content not reviewable)*
- ✅ addons/dashboard/build/plugins/flux-0.6.0.tar.gz.license
- ✅ addons/dashboard/build/plugins/prometheus-0.8.2.tar.gz *(binary — content not reviewable)*
- ✅ addons/dashboard/build/plugins/prometheus-0.8.2.tar.gz.license
- ✅ addons/dashboard/dashboard.module.psm1
- ✅ addons/dashboard/dashboard.module.unit.tests.ps1
- ⚠️ addons/ingress/nginx-gw/Disable.ps1 *(diff not provided)*
- ⚠️ addons/ingress/nginx-gw/Enable.ps1 *(diff not provided)*
- ⚠️ addons/ingress/nginx/Disable.ps1 *(diff not provided)*
- ⚠️ addons/ingress/nginx/Enable.ps1 *(diff not provided)*
- ⚠️ addons/ingress/traefik/Disable.ps1 *(diff not provided)*
- ⚠️ addons/ingress/traefik/Enable.ps1 *(diff not provided)*
- ⚠️ addons/monitoring/Disable.ps1 *(diff not provided)*
- ⚠️ addons/monitoring/Enable.ps1 *(diff not provided)*
- ⚠️ addons/monitoring/Update.ps1 *(diff not provided)*
- ⚠️ addons/rollout/fluxcd/Disable.ps1 *(diff not provided)*
- ⚠️ addons/rollout/fluxcd/Enable.ps1 *(diff not provided)*
- ⚠️ addons/rollout/fluxcd/Update.ps1 *(diff not provided)*
- ⚠️ addons/security/Disable.ps1 *(diff not provided)*
- ⚠️ addons/security/Enable.ps1 *(diff not provided)*
- ⚠️ addons/security/Update.ps1 *(diff not provided)*

krotz-dieter (commented):

---

## Code Review: PR #2583 — Headlamp plugin tgz model

### Critical: Incomplete diff — 15 files unreviewed

The PR header lists 30 changed files, but the diff only contains ~15 of them. Missing entirely:
- `addons/ingress/nginx-gw/{Enable,Disable}.ps1`
- `addons/ingress/nginx/{Enable,Disable}.ps1`
- `addons/ingress/traefik/{Enable,Disable}.ps1`
- `addons/monitoring/{Enable,Disable,Update}.ps1`
- `addons/rollout/fluxcd/{Enable,Disable,Update}.ps1`
- `addons/security/{Enable,Disable,Update}.ps1`

These are the files that call `Sync-HeadlampPlugins` from other addons — the most critical integration point of the entire feature. Without seeing them it's impossible to verify:
1. That each file imports `dashboard.module.psm1` before calling `Sync-HeadlampPlugins`
2. That there are no circular import issues
3. That errors from `Sync-HeadlampPlugins` are handled consistently

---

### Bug: JSON command string not escaped in `Build-PluginPatchJson`

**`addons/dashboard/dashboard.module.psm1`**, function `Build-PluginPatchJson`:

```powershell
$cp = 'mkdir -p ' + $pluginsDir + '/' + $n + ' && cp -r /plugins/' + $n + '/. ' + $pluginsDir + '/' + $n + '/'
$icParts += '{\"name\":\"' + $n + '\",\"image\":\"' + $img + '\",' +
            '\"command\":[\"sh\",\"-c\",\"' + $cp + '\"],' + ...
```

The `$cp` string is embedded directly into JSON without being JSON-escaped. `$n` (e.g., `cert-manager-plugin`) and `$pluginsDir` (`/tmp/headlamp/plugins`) are safe today, but any future plugin name or path containing `"`, `\`, or newlines would silently produce invalid JSON passed to kubectl. This is especially fragile because `Apply-HeadlampPluginPatch` then does a broad `-replace '\\"', '"'` pass that could further corrupt it.

The fix is to use `[System.Text.Json.JsonSerializer]::Serialize($cp)` or `ConvertTo-Json $cp` for the command string, rather than raw string interpolation.

---

### Bug: `--patch-file` path on Windows with spaces

**`addons/dashboard/dashboard.module.psm1`**, function `Apply-HeadlampPluginPatch`:

```powershell
$patchFile = Join-Path ([System.IO.Path]::GetTempPath()) ("headlamp-plugin-patch-{0}.json" -f ([System.Guid]::NewGuid().ToString('N')))
...
$patchResult = Invoke-Kubectl -Params 'patch', 'deployment', 'headlamp', '-n', 'dashboard', '--type=strategic', '--patch-file', $patchFile
```

`GetTempPath()` on Windows can return a path with spaces (e.g., `C:\Users\Some User\AppData\Local\Temp`). When this is passed as a separate element in the `-Params` array to `Invoke-Kubectl`, the behavior depends entirely on how `Invoke-Kubectl` handles the array-to-process argument mapping. If it uses `Start-Process` or `&` with splatting, this is safe. If it joins to a string and passes to a shell, the path with spaces would be mis-tokenized. This should be verified against the `Invoke-Kubectl` implementation.

---

### Issue: `Wait-ForHeadlampRollout` blocks the enable/update flow for up to 300 seconds

**`addons/dashboard/dashboard.module.psm1`**, `Sync-HeadlampPlugins`:

```powershell
if ($patched) {
    Wait-ForHeadlampRollout | Out-Null
}
```

`Wait-ForHeadlampRollout` has a default timeout of 300s and is called synchronously in both `Enable.ps1` and `Update.ps1`. In an environment where an init-container image is unavailable (e.g., ACR unreachable, image not yet pushed), the user's `k2s addons enable dashboard` will silently block for up to 5 minutes before any diagnostic output appears. The function itself is non-fatal (logs then returns), but the wall-clock delay is significant and undiscoverable from the user perspective. Consider reducing the default or adding a progress indicator.

---

### Issue: `headlamp-plugin-images.unit.tests.ps1` — sentinel count will block future plugin additions

```powershell
It 'declares exactly three plugins' {
    $script:lock.plugins.Count | Should -Be 3
}
```

This hard-coded `3` will fail CI whenever a fourth plugin is added. The contributor adding the plugin would need to know to find and update this test. Consider using a comment explaining this is intentional, or replacing with `Should -BeGreaterOrEqual 1` and relying on the parity tests to catch mismatches.

---

### Minor: `dashboard.module.psm1` exports internal capability detectors

```powershell
Export-ModuleMember -Function ..., `
    Test-FluxCapabilityAvailable, Test-CertManagerCapabilityAvailable, Test-PrometheusCapabilityAvailable, ...
```

These detectors have no callers outside the module other than the unit tests (which use `InModuleScope`). Exporting them expands the public API surface unnecessarily and makes it harder to refactor detection logic without a breaking change. The unit tests in `dashboard.module.unit.tests.ps1` assert these are exported — that test itself would need changing if they're made internal. This should be intentional.

---

### Minor: Lock file parity test uses fragile YAML line-parse

**`addons/dashboard/build/headlamp-plugin-images.unit.tests.ps1`**:

```powershell
$script:manifestImages = Get-Content -Path $script:manifestPath |
    Where-Object { $_ -match 'shsk2s\.azurecr\.io/headlamp-plugin-' } |
    ForEach-Object { ($_ -replace '^\s*-\s*', '').Trim() }
```

This works for the current YAML layout, but would silently miss entries if the YAML is reformatted (e.g., flow sequence `[img1, img2]`). A YAML parser would be more robust. Low risk since the format is machine-managed.

---

## Files Reviewed

- ✅ addons/dashboard/Disable.ps1
- ✅ addons/dashboard/Enable.ps1
- ✅ addons/dashboard/README.md
- ✅ addons/dashboard/Update.ps1
- ✅ addons/dashboard/addon.manifest.yaml
- ✅ addons/dashboard/build/Build-HeadlampPluginImages.Methods.ps1
- ✅ addons/dashboard/build/Build-HeadlampPluginImages.ps1
- ✅ addons/dashboard/build/README.md
- ✅ addons/dashboard/build/headlamp-plugin-images.unit.tests.ps1
- ✅ addons/dashboard/build/headlamp-plugins.lock.json
- ✅ addons/dashboard/build/headlamp-plugins.lock.json.license
- ✅ addons/dashboard/build/plugins/cert-manager-0.1.0.tar.gz *(binary — not reviewable)*
- ✅ addons/dashboard/build/plugins/cert-manager-0.1.0.tar.gz.license
- ✅ addons/dashboard/build/plugins/flux-0.6.0.tar.gz *(binary — not reviewable)*
- ✅ addons/dashboard/build/plugins/flux-0.6.0.tar.gz.license
- ✅ addons/dashboard/build/plugins/prometheus-0.8.2.tar.gz *(binary — not reviewable)*
- ✅ addons/dashboard/build/plugins/prometheus-0.8.2.tar.gz.license
- ✅ addons/dashboard/dashboard.module.psm1
- ✅ addons/dashboard/dashboard.module.unit.tests.ps1
- ❌ addons/ingress/nginx-gw/Disable.ps1 *(not in diff)*
- ❌ addons/ingress/nginx-gw/Enable.ps1 *(not in diff)*
- ❌ addons/ingress/nginx/Disable.ps1 *(not in diff)*
- ❌ addons/ingress/nginx/Enable.ps1 *(not in diff)*
- ❌ addons/ingress/traefik/Disable.ps1 *(not in diff)*
- ❌ addons/ingress/traefik/Enable.ps1 *(not in diff)*
- ❌ addons/monitoring/Disable.ps1 *(not in diff)*
- ❌ addons/monitoring/Enable.ps1 *(not in diff)*
- ❌ addons/monitoring/Update.ps1 *(not in diff)*
- ❌ addons/rollout/fluxcd/Disable.ps1 *(not in diff)*
- ❌ addons/rollout/fluxcd/Enable.ps1 *(not in diff)*
- ❌ addons/rollout/fluxcd/Update.ps1 *(not in diff)*
- ❌ addons/security/Disable.ps1 *(not in diff)*
- ❌ addons/security/Enable.ps1 *(not in diff)*
- ❌ addons/security/Update.ps1 *(not in diff)*

vvkrishnakanth9:
All major issues from PR review have been fixed.

krotz-dieter (commented):
Here is my review of PR #2583:

---

## Review: Headlamp plugin integration changes

The overall design is sound — capability-based plugin activation via init-containers is idempotent, offline-friendly, and well-structured. The lock file + parity tests are a strong safety net. A few issues warrant attention:

---

### Bug: Counter reset inside `InModuleScope` is ineffective in unit tests

**Files:** `addons/dashboard/dashboard.module.unit.tests.ps1` (lines 652, 708, 781)

In all three "fallback to CRD" test contexts, the call counter is reset inside `InModuleScope`:
```powershell
It 'returns true when the Flux CRD is found' {
    InModuleScope $moduleName {
        $script:fluxKubectlCall = 0   # sets the MODULE's $script: scope
        $result = Test-FluxCapabilityAvailable
    }
}
```

The `Mock` closure at `BeforeAll` captures `$script:fluxKubectlCall` from the **test file's** script scope. `InModuleScope` runs code in the **dashboard module's** script scope. These are different `$script:` scopes, so the reset is a no-op on the variable the mock increments.

Currently harmless (each `Context` has exactly one `It`, so `BeforeAll`'s initialization suffices), but this will silently break if a second `It` is added to these contexts. The counter resets inside `InModuleScope` should be moved outside `InModuleScope`, or the test should use a `BeforeEach` to reset the counter.

---

### Bug: `Sync-HeadlampPlugins` can leave the dashboard in a broken state after `Add-AddonToSetupJson`

**File:** `addons/dashboard/Enable.ps1` (lines 122–125)

```powershell
Add-AddonToSetupJson -Addon ([pscustomobject] @{Name = 'dashboard' })  # addon is now "enabled"

Write-Log '[Dashboard] Syncing Headlamp plugins' -Console
Sync-HeadlampPlugins   # no try/catch — throws if kubectl patch fails
```

`Apply-HeadlampPluginPatch` throws on `kubectl patch` failure. If that happens after `Add-AddonToSetupJson`, the addon is registered as enabled but plugins are not activated. A subsequent `k2s addons enable dashboard` will fail with "addon already enabled", leaving the user with no clean recovery path short of manually editing `setup.json` or calling `k2s addons disable dashboard`.

The fix is either to wrap the call in `try/catch` (log and continue, since plugins can be re-synced via `k2s addons update dashboard`) or to move `Add-AddonToSetupJson` after the sync succeeds.

---

### Minor: `Test-PluginImageLayout` retains an intentionally unused parameter

**File:** `addons/dashboard/build/Build-HeadlampPluginImages.Methods.ps1` (line ~285)

```powershell
param(
    [string] $CraneExe = '',   # retained for call-site compatibility; unused
    ...
)
$null = $CraneExe
```

The comment explains the rationale (backward compatibility). The `$null =` suppressor is sufficient, but callers (in `Build-HeadlampPluginImages.ps1`) always pass `-CraneExe $crane`, which is misleading since the value is ignored. Consider adding a note in the `.DESCRIPTION` that the parameter is a no-op, or just removing it since it's not a public API — the only caller is the orchestrator in the same file.

---

### Observation: `Update.ps1` called from `Enable.ps1` also runs `Sync-HeadlampPlugins` (no-op in that context)

`Enable.ps1` line 120 calls `Update.ps1`, which ends with `Sync-HeadlampPlugins`. At that point the dashboard hasn't been added to `setup.json` yet, so the guard in `Sync-HeadlampPlugins` (`Test-IsAddonEnabled` → false) silently skips it. Then `Enable.ps1` calls `Sync-HeadlampPlugins` directly at line 125 (after `Add-AddonToSetupJson`), which is the actual activation. This is functionally correct but the skip in `Update.ps1` is silent and might be confusing during debugging. Not a bug, but worth a comment.

---

## Files Reviewed
- ✅ addons/dashboard/Disable.ps1
- ✅ addons/dashboard/Enable.ps1
- ✅ addons/dashboard/README.md
- ✅ addons/dashboard/Update.ps1
- ✅ addons/dashboard/addon.manifest.yaml
- ✅ addons/dashboard/build/Build-HeadlampPluginImages.Methods.ps1
- ✅ addons/dashboard/build/Build-HeadlampPluginImages.ps1
- ✅ addons/dashboard/build/README.md
- ✅ addons/dashboard/build/headlamp-plugin-images.unit.tests.ps1
- ✅ addons/dashboard/build/headlamp-plugins.lock.json
- ✅ addons/dashboard/build/headlamp-plugins.lock.json.license
- ✅ addons/dashboard/build/plugins/cert-manager-0.1.0.tar.gz.license
- ✅ addons/dashboard/build/plugins/flux-0.6.0.tar.gz.license
- ✅ addons/dashboard/build/plugins/prometheus-0.8.2.tar.gz.license
- ✅ addons/dashboard/dashboard.module.psm1
- ✅ addons/dashboard/dashboard.module.unit.tests.ps1
- ✅ addons/ingress/nginx-gw/Disable.ps1
- ✅ addons/ingress/nginx-gw/Enable.ps1
- ✅ addons/ingress/nginx/Disable.ps1
- ✅ addons/ingress/nginx/Enable.ps1
- ✅ addons/ingress/traefik/Disable.ps1
- ✅ addons/ingress/traefik/Enable.ps1
- ✅ addons/monitoring/Disable.ps1
- ✅ addons/monitoring/Enable.ps1
- ✅ addons/monitoring/Update.ps1
- ✅ addons/rollout/fluxcd/Disable.ps1
- ✅ addons/rollout/fluxcd/Enable.ps1
- ✅ addons/rollout/fluxcd/Update.ps1
- ✅ addons/security/Disable.ps1
- ✅ addons/security/Update.ps1

krotz-dieter (commented):
Now I have enough context for a thorough review. Let me check one more thing:

Now I have everything I need for a complete review.

---

## Code Review: PR #2583 — Headlamp plugin integration changes

Overall this is a well-structured feature: capability-based plugin activation via init-containers, a solid supply-chain (lock file + build script + offline packaging), extensive unit tests, and good documentation. A few issues deserve attention.

---

### Bug (Medium): `Sync-HeadlampPlugins` called without error handling in all callers except `dashboard/Enable.ps1`

`dashboard/Enable.ps1` explicitly wraps the call in try/catch with a clear rationale:

> *"A transient sync failure (e.g. a kubectl patch error) must NOT leave the dashboard in a registered-but-failed state."*

But every other caller in this PR propagates the error unchecked:

| File | Consequence of kubectl patch failure |
|---|---|
| `addons/dashboard/Update.ps1` (line added in this PR) | Entire dashboard update fails |
| `addons/ingress/nginx/Enable.ps1` L192 | Nginx enable fails |
| `addons/ingress/nginx/Disable.ps1` L101 | Nginx disable fails |
| `addons/ingress/traefik/Enable.ps1` L211 | Traefik enable fails |
| `addons/ingress/traefik/Disable.ps1` L102 | Traefik disable fails |
| `addons/ingress/nginx-gw/Enable.ps1` L267 | NginxGW enable fails |
| `addons/ingress/nginx-gw/Disable.ps1` L107 | NginxGW disable fails |
| `addons/monitoring/Enable.ps1` L208 | Monitoring enable fails |
| `addons/monitoring/Disable.ps1` L78 | Monitoring disable fails |
| `addons/monitoring/Update.ps1` L173 | Monitoring update fails |
| `addons/rollout/fluxcd/Enable.ps1` L129 | FluxCD enable fails |
| `addons/rollout/fluxcd/Disable.ps1` L94 | FluxCD disable fails |
| `addons/rollout/fluxcd/Update.ps1` L32 | FluxCD update fails |
| `addons/security/Enable.ps1` L448 | Security enable fails |
| `addons/security/Disable.ps1` L185 | Security disable fails |
| `addons/security/Update.ps1` L30 | Security update fails |

A transient API-server hiccup or RBAC issue during the plugin patch would cascade into a failure of an otherwise-successful addon operation. The same try/catch pattern from `Enable.ps1` should be applied to all these callers, or `Sync-HeadlampPlugins` itself should catch and warn internally instead of throwing.

---

### Bug (Low): `Remove-HeadlampPluginPatch` is exported dead code

`dashboard.module.psm1` exports `Remove-HeadlampPluginPatch` with the docstring *"Called when the dashboard addon is being disabled."* It is never called from any production script. `Disable.ps1` explicitly explains (correctly) why it skips plugin cleanup before Helm uninstall. The exported function is unreachable in the production path and will confuse future maintainers who expect it to be part of the disable lifecycle.

Either call it where appropriate, or remove the misleading docstring and consider keeping it unexported.

---

### Minor: Prometheus lock-file URL uses inconsistent filename pattern

In `addons/dashboard/build/headlamp-plugins.lock.json`, the `prebuilt.url` filenames are:
- `flux`: `.../headlamp-k8s-flux-0.6.0.tar.gz`
- `cert-manager`: `.../headlamp-k8s-cert-manager-0.1.0.tar.gz`
- `prometheus`: `.../prometheus-0.8.2.tar.gz` ← no `headlamp-k8s-` prefix

This is likely just the actual upstream release asset name and is safe in practice (the vendored tarball is committed and sha256-pinned), but the inconsistency is worth a comment in the lock file or a note in the URL field to clarify it's intentional.

---

### Observation: Parity test does not verify YAML key path

`headlamp-plugin-images.unit.tests.ps1` validates lock ↔ manifest parity by grepping `addon.manifest.yaml` for lines containing `shsk2s.azurecr.io/headlamp-plugin-`. This catches missing or extra image refs but would also match images placed under an incorrect YAML key (e.g., `windows` instead of `linux`, or a comment). It won't catch a structural misplacement. This is low-risk given the manifest is currently correct, but worth knowing.

---

## Files Reviewed

- ✅ addons/dashboard/Disable.ps1
- ✅ addons/dashboard/Enable.ps1
- ✅ addons/dashboard/README.md
- ✅ addons/dashboard/Update.ps1
- ✅ addons/dashboard/addon.manifest.yaml
- ✅ addons/dashboard/build/Build-HeadlampPluginImages.Methods.ps1
- ✅ addons/dashboard/build/Build-HeadlampPluginImages.ps1
- ✅ addons/dashboard/build/README.md
- ✅ addons/dashboard/build/headlamp-plugin-images.unit.tests.ps1
- ✅ addons/dashboard/build/headlamp-plugins.lock.json
- ✅ addons/dashboard/build/headlamp-plugins.lock.json.license
- ✅ addons/dashboard/build/plugins/cert-manager-0.1.0.tar.gz.license
- ✅ addons/dashboard/build/plugins/flux-0.6.0.tar.gz.license
- ✅ addons/dashboard/build/plugins/prometheus-0.8.2.tar.gz.license
- ✅ addons/dashboard/dashboard.module.psm1
- ✅ addons/dashboard/dashboard.module.unit.tests.ps1
- ✅ addons/ingress/nginx-gw/Disable.ps1
- ✅ addons/ingress/nginx-gw/Enable.ps1
- ✅ addons/ingress/nginx/Disable.ps1
- ✅ addons/ingress/nginx/Enable.ps1
- ✅ addons/ingress/traefik/Disable.ps1
- ✅ addons/ingress/traefik/Enable.ps1
- ✅ addons/monitoring/Disable.ps1
- ✅ addons/monitoring/Enable.ps1
- ✅ addons/monitoring/Update.ps1
- ✅ addons/rollout/fluxcd/Disable.ps1
- ✅ addons/rollout/fluxcd/Enable.ps1
- ✅ addons/rollout/fluxcd/Update.ps1
- ✅ addons/security/Disable.ps1
- ✅ addons/security/Update.ps1

krotz-dieter (commented):
Now I have enough context for a thorough review. Let me analyze all findings.

Based on my thorough review, here is my analysis:

---

## Code Review: PR #2583 — Headlamp Plugin Integration

### Summary
This is a well-structured PR that introduces capability-based Headlamp plugin injection via Kubernetes init-containers. The implementation is thoughtful (offline-first, idempotent, gracefully-degrading). I found no critical bugs, but there are a few issues worth addressing.

---

### Issues Found

#### 1. Misleading log message in `Apply-HeadlampPluginPatch` (`dashboard.module.psm1`)

```powershell
Write-Log "[Dashboard][Plugin] Patching headlamp deployment: $($toAddOrUpdate.Count) add/update, $($toRemove.Count) remove" -Console

$patchJson = Build-PluginPatchJson -K2sInitContainers $InitContainers -NamesToRemove $toRemove
```

`$toAddOrUpdate` is calculated but **never passed** to `Build-PluginPatchJson`. The patch always contains the full `$InitContainers` set (all desired plugins, not just the changed ones). The log line says e.g. "1 add/update, 0 remove" but the actual patch contains all 3 plugins. Operationally harmless (strategic merge is idempotent for unchanged entries), but the log claim is false and could mislead during incident diagnosis.

#### 2. `Wait-ForHeadlampRollout` blocks all capability-touching addon operations for up to 300s (`dashboard.module.psm1`)

`Sync-HeadlampPlugins` calls `Wait-ForHeadlampRollout` (300s default) synchronously whenever a patch is issued. This means any addon enable/disable that changes a registered capability (ingress nginx/traefik/nginx-gw, monitoring, fluxcd, security) will stall for up to 5 minutes if a plugin init-container image pull is slow or fails. The affected scripts (`Enable.ps1`/`Disable.ps1` of 6 addons) call `Sync-HeadlampPlugins` synchronously inside a `try/catch`, so the rollout wait is inline with those operations. Consider whether the default timeout should be shorter, or whether the rollout wait should be opt-in rather than always-on.

#### 3. `Remove-HeadlampPluginPatch` is exported but has no callers (`dashboard.module.psm1`)

`Disable.ps1` intentionally skips it (Helm deletes the deployment). The 6 other addon `Disable.ps1` scripts all call `Sync-HeadlampPlugins` instead. `Remove-HeadlampPluginPatch` is dead API surface. It should either be removed from `Export-ModuleMember` (keeping it internal) or have a documented intended caller.

#### 4. `Build-PluginPatchJson` constructs JSON via string concatenation without escaping (`dashboard.module.psm1`)

```powershell
$icParts += '{\"name\":\"' + $n + '\",\"image\":\"' + $img + '\",...}'
```

If any plugin `Name` or `Image` ever contains `"` or `\` (not the case today, but a future maintenance hazard), the produced JSON is silently malformed and the `kubectl patch` call will fail with a cryptic parse error. Given the validated inputs (OCI refs, semver-style names), this is safe today, but the function is a maintenance trap. Low priority but worth noting for future-proofing.

#### 5. Minor: `$dashboardModule` is not imported in nginx-gw/Enable.ps1's initial `Import-Module` call (expected, but `Import-Module $dashboardModule -Force` without error handling if the file is corrupted)

```powershell
if (Test-Path $dashboardModule) {
    Import-Module $dashboardModule -Force
    if (Get-Command Sync-HeadlampPlugins -ErrorAction SilentlyContinue) {
        ...Sync-HeadlampPlugins...
    }
```

The `Get-Command` guard is good, but a parse error in `dashboard.module.psm1` would cause `Import-Module -Force` to throw, propagating to the outer `try/catch` in each addon's Enable/Disable. The existing `try/catch` around `Sync-HeadlampPlugins` does NOT wrap the `Import-Module` call itself. A corrupted dashboard module would silently abort the addon's Enable/Disable with no specific error. Wrapping the entire block in `try/catch` would be safer.

---

### Minor Observations (not bugs)
- SHA256 hashes in `headlamp-plugins.lock.json` are all correctly 64-char lowercase hex. ✅
- `pluginsDir: "/tmp/headlamp/plugins"` in `values.yaml` matches `$script:HeadlampPluginsDir` in the module. ✅
- Strategic merge patch delete directives use correct merge keys (`name` for volumes, `mountPath` for volumeMounts). ✅
- The `dashboardModule` variable is correctly defined before use in all Enable.ps1 files. ✅
- `Sync-HeadlampPlugins` correctly short-circuits via `Test-IsAddonEnabled` during fresh installs (before `Add-AddonToSetupJson` runs). ✅

---

## Files Reviewed
- ✅ addons/dashboard/Disable.ps1
- ✅ addons/dashboard/Enable.ps1
- ✅ addons/dashboard/README.md
- ✅ addons/dashboard/Update.ps1
- ✅ addons/dashboard/addon.manifest.yaml
- ✅ addons/dashboard/build/Build-HeadlampPluginImages.Methods.ps1
- ✅ addons/dashboard/build/Build-HeadlampPluginImages.ps1
- ✅ addons/dashboard/build/README.md
- ✅ addons/dashboard/build/headlamp-plugin-images.unit.tests.ps1
- ✅ addons/dashboard/build/headlamp-plugins.lock.json
- ✅ addons/dashboard/build/headlamp-plugins.lock.json.license
- ✅ addons/dashboard/build/plugins/cert-manager-0.1.0.tar.gz.license
- ✅ addons/dashboard/build/plugins/flux-0.6.0.tar.gz.license
- ✅ addons/dashboard/build/plugins/prometheus-0.8.2.tar.gz.license
- ✅ addons/dashboard/dashboard.module.psm1
- ✅ addons/dashboard/dashboard.module.unit.tests.ps1
- ✅ addons/ingress/nginx-gw/Disable.ps1
- ✅ addons/ingress/nginx-gw/Enable.ps1
- ✅ addons/ingress/nginx/Disable.ps1
- ✅ addons/ingress/nginx/Enable.ps1
- ✅ addons/ingress/traefik/Disable.ps1
- ✅ addons/ingress/traefik/Enable.ps1
- ✅ addons/monitoring/Disable.ps1
- ✅ addons/monitoring/Enable.ps1
- ✅ addons/monitoring/Update.ps1
- ✅ addons/rollout/fluxcd/Disable.ps1
- ✅ addons/rollout/fluxcd/Enable.ps1
- ✅ addons/rollout/fluxcd/Update.ps1
- ✅ addons/security/Disable.ps1
- ✅ addons/security/Update.ps1

PR #2587, opened 2026-06-15
Add headless support in webhook, log enhancement and readme

### LINKED ISSUE #2560: Move `clusterip-webhook` certificate generation into the Deployment

GermanGLLopez:
Fixes #2560

### DIFF
--- a/k2s/cmd/clusterip-webhook/README.md
+++ b/k2s/cmd/clusterip-webhook/README.md
@@ -21,7 +21,15 @@ The webhook runs as a container inside the Kubernetes cluster. The image is buil
 From the repository root:
 
 ```console
-k2s image build --input-folder k2s --dockerfile k2s/cmd/clusterip-webhook/Dockerfile --image-name shsk2s.azurecr.io/clusterip-webhook --image-tag v1.2.0
+k2s image build --input-folder k2s --dockerfile k2s/cmd/clusterip-webhook/Dockerfile --image-name shsk2s.azurecr.io/clusterip-webhook --image-tag v1.3.0
+```
+
+### Push to registry
+
+To push the built image to the container registry:
+
+```console
+k2s image push -n shsk2s.azurecr.io/clusterip-webhook:v1.3.0
 ```
 
 ### Version Bump Checklist
--- a/k2s/cmd/clusterip-webhook/README.md
+++ b/k2s/cmd/clusterip-webhook/README.md
@@ -1,0 +1,67 @@
+<!--
+SPDX-FileCopyrightText: © 2026 Siemens Healthineers AG
+
+SPDX-License-Identifier: MIT
+-->
+
+# ClusterIP Webhook
+
+The ClusterIP webhook is a Kubernetes mutating admission webhook that assigns ClusterIPs from OS-specific subnets (Linux vs Windows) to Services created in the K2s cluster.
+
+## Building the Container Image
+
+The webhook runs as a container inside the Kubernetes cluster. The image is built from the `Dockerfile` in this directory.
+
+### Prerequisites
+
+- A running K2s cluster
+
+### Build
+
+From the repository root:
+
+```console
+k2s image build --input-folder k2s --dockerfile k2s/cmd/clusterip-webhook/Dockerfile --image-name shsk2s.azurecr.io/clusterip-webhook --image-tag v1.2.0
+```
+
+### Version Bump Checklist
+
+When changing the webhook code, update the image tag in:
+
+1. `lib/manifests/clusterip-webhook/deployment.yaml` — both `init-cert` and `webhook` containers
+2. `lib/modules/k2s/k2s.node.module/linuxnode/distros/common-setup.module.psm1` — image pull during cluster setup
+
+## Operating Modes
+
+The webhook binary supports two modes:
+
+### Webhook Server (default)
+
+Runs the admission webhook HTTP server:
+
+```
+/clusterip-webhook --addr=:8443 --tls-cert=/certs/tls.crt --tls-key=/certs/tls.key
+```
+
+### Certificate Init (`--init-cert`)
+
+Generates a self-signed TLS certificate and patches the webhook configuration. Used as an init container:
+
+```
+/clusterip-webhook --init-cert --tls-cert=/certs/tls.crt --tls-key=/certs/tls.key \
+    --service-name=clusterip-webhook --namespace=k2s-webhook --webhook-name=k2s-webhook
+```
+
+This mode:
+- Generates a CA certificate and server certificate (ECDSA P-256, 1-year validity)
+- Writes `tls.crt` and `tls.key` to the shared emptyDir volume
+- Patches the `caBundle` field in the MutatingWebhookConfiguration
+- Exits after completion
+
+## Architecture
+
+- The Deployment uses an **emptyDir** volume for `/certs`, avoiding a dependency on a pre-existing Secret
+- The **init container** generates fresh TLS certificates on every Pod creation
+- Certificate renewal is triggered by recreating the Pod (e.g., `kubectl rollout restart`)
+- `k2s system certificate renew` includes the webhook in its renewal flow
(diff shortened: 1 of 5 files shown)


krotz-dieter (commented):
Working from the provided diff. Let me analyze all changed files carefully.

---

## PR Review: Add headless support in webhook, log enhancement and readme

### `k2s/cmd/clusterip-webhook/main.go`

**Headless service detection logic** — The dual-field check is correct and meaningful. Since this is a **mutating admission webhook**, it processes requests *before* the API server normalizes `ClusterIP`/`ClusterIPs` fields into a consistent state. A client could legitimately submit a Service specifying only `clusterIPs: ["None"]` without explicitly setting `clusterIP: None`, or vice versa. The `||` condition handles both cases properly.

`containsNone` is safe: iterating a nil slice in Go is a no-op and returns `false` — no panic risk.

**One real concern — missing test coverage for the new branch:** The PR adds a new logical path (`containsNone(service.Spec.ClusterIPs)`) with no accompanying unit test. If there are existing unit tests for `mutateService`, they should be updated to cover the `ClusterIPs`-only headless case to prevent future regressions.

**Version in startup log** — The `ve.GetVersion().String()` addition is straightforward and correct.

### `k2s/cmd/clusterip-webhook/README.md`

Accurate and well-structured. One note: the build/push examples hardcode `v1.3.0`:
```
k2s image build ... --image-tag v1.3.0
```
This will silently become outdated after the next version bump. Consider using a placeholder like `<version>` so readers know to substitute the current tag.

### `lib/manifests/clusterip-webhook/deployment.yaml`

Version bumped consistently from `v1.2.0` → `v1.3.0` in both `init-cert` and `webhook` containers. Correct.

### `lib/modules/k2s/k2s.node.module/linuxnode/distros/common-setup.module.psm1`

Version bumped consistently from `v1.2.0` → `v1.3.0`. Correct.

---

### Summary

The core fix is logically sound. The only actionable issue is the **absence of a unit test** for the `containsNone(service.Spec.ClusterIPs)` branch — if tests exist in this package, this case should be covered. The README version hardcoding is a minor maintenance concern worth addressing.

## Files Reviewed
- ✅ k2s/cmd/clusterip-webhook/main.go
- ✅ k2s/cmd/clusterip-webhook/README.md
- ✅ lib/manifests/clusterip-webhook/deployment.yaml

PR #2594, opened 2026-06-16
#2250 Keep kubeadm as source of truth for pause image

### LINKED ISSUE #2550: Some of e2e tests are failing sporadically in v1.9.0
Some of e2e tests are failing sporadically in v1.8.0

krotz-dieter:
#2250 Keep kubeadm as source of truth for pause image

### DIFF
--- a/cfg/nodeextension/debian12/scripts/install-k8s-packages.sh
+++ b/cfg/nodeextension/debian12/scripts/install-k8s-packages.sh
@@ -145,14 +145,16 @@ else
     echo "[InstallK8s] unqualified-search-registries already configured in /etc/containers/registries.conf, skipping"
 fi
 
-KUBEADM_PAUSE_IMAGE="$(kubeadm config images list | grep '/pause:' | tail -n 1 || true)"
+KUBEADM_VERSION="$(kubeadm version -o short)"
+KUBEADM_PAUSE_IMAGE="$(kubeadm config images list --kubernetes-version "$KUBEADM_VERSION" | grep '/pause:' | tail -n 1 || true)"
 if [ -n "$KUBEADM_PAUSE_IMAGE" ]; then
-    echo "[InstallK8s] Configuring CRI-O pause image from kubeadm: $KUBEADM_PAUSE_IMAGE"
+    echo "[InstallK8s] Configuring CRI-O pause image from kubeadm $KUBEADM_VERSION: $KUBEADM_PAUSE_IMAGE"
     sudo mkdir -p /etc/crio/crio.conf.d
     {
         echo '[crio.image]'
         echo "pause_image = \"$KUBEADM_PAUSE_IMAGE\""
     } | sudo tee /etc/crio/crio.conf.d/20-k2s-kubeadm-pause.conf > /dev/null
+    sudo systemctl restart crio || true
 else
     echo "[InstallK8s] WARNING: Could not resolve pause image from kubeadm; keeping CRI-O package default"
 fi
--- a/cfg/nodeextension/debian12/scripts/install-k8s-packages.sh
+++ b/cfg/nodeextension/debian12/scripts/install-k8s-packages.sh
@@ -145,6 +145,18 @@ else
     echo "[InstallK8s] unqualified-search-registries already configured in /etc/containers/registries.conf, skipping"
 fi
 
+KUBEADM_PAUSE_IMAGE="$(kubeadm config images list | grep '/pause:' | tail -n 1 || true)"
+if [ -n "$KUBEADM_PAUSE_IMAGE" ]; then
+    echo "[InstallK8s] Configuring CRI-O pause image from kubeadm: $KUBEADM_PAUSE_IMAGE"
+    sudo mkdir -p /etc/crio/crio.conf.d
+    {
+        echo '[crio.image]'
+        echo "pause_image = \"$KUBEADM_PAUSE_IMAGE\""
+    } | sudo tee /etc/crio/crio.conf.d/20-k2s-kubeadm-pause.conf > /dev/null
+else
+    echo "[InstallK8s] WARNING: Could not resolve pause image from kubeadm; keeping CRI-O package default"
+fi
+
 # ---------------------------------------------------------------------------
 # Hold kubelet, kubeadm, kubectl
 # ---------------------------------------------------------------------------
--- a/cfg/nodeextension/debian13/scripts/install-k8s-packages.sh
+++ b/cfg/nodeextension/debian13/scripts/install-k8s-packages.sh
@@ -145,14 +145,16 @@ else
     echo "[InstallK8s] unqualified-search-registries already configured in /etc/containers/registries.conf, skipping"
 fi
 
-KUBEADM_PAUSE_IMAGE="$(kubeadm config images list | grep '/pause:' | tail -n 1 || true)"
+KUBEADM_VERSION="$(kubeadm version -o short)"
+KUBEADM_PAUSE_IMAGE="$(kubeadm config images list --kubernetes-version "$KUBEADM_VERSION" | grep '/pause:' | tail -n 1 || true)"
 if [ -n "$KUBEADM_PAUSE_IMAGE" ]; then
-    echo "[InstallK8s] Configuring CRI-O pause image from kubeadm: $KUBEADM_PAUSE_IMAGE"
+    echo "[InstallK8s] Configuring CRI-O pause image from kubeadm $KUBEADM_VERSION: $KUBEADM_PAUSE_IMAGE"
     sudo mkdir -p /etc/crio/crio.conf.d
     {
         echo '[crio.image]'
         echo "pause_image = \"$KUBEADM_PAUSE_IMAGE\""
     } | sudo tee /etc/crio/crio.conf.d/20-k2s-kubeadm-pause.conf > /dev/null
+    sudo systemctl restart crio || true
 else
     echo "[InstallK8s] WARNING: Could not resolve pause image from kubeadm; keeping CRI-O package default"
 fi
--- a/cfg/nodeextension/debian13/scripts/install-k8s-packages.sh
+++ b/cfg/nodeextension/debian13/scripts/install-k8s-packages.sh
@@ -145,6 +145,18 @@ else
     echo "[InstallK8s] unqualified-search-registries already configured in /etc/containers/registries.conf, skipping"
 fi
 
+KUBEADM_PAUSE_IMAGE="$(kubeadm config images list | grep '/pause:' | tail -n 1 || true)"
+if [ -n "$KUBEADM_PAUSE_IMAGE" ]; then
+    echo "[InstallK8s] Configuring CRI-O pause image from kubeadm: $KUBEADM_PAUSE_IMAGE"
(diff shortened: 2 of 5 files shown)


krotz-dieter (commented):
## PR Review: #2594 — Keep kubeadm as source of truth for pause image

Overall the approach is sound, but there are several bugs and inconsistencies across the changed files.

---

### 🐛 Bug: Missing CRI-O restart in `install-k8s-packages.sh`

**Files:** `cfg/nodeextension/debian12/scripts/install-k8s-packages.sh`, `cfg/nodeextension/debian13/scripts/install-k8s-packages.sh`

The new block writes `/etc/crio/crio.conf.d/20-k2s-kubeadm-pause.conf` but never restarts CRI-O. By contrast, `apply-debian-delta.sh` correctly calls `systemctl restart crio` after writing the same config file. If CRI-O is already running when the install script executes (e.g., the package post-install scriptlet started it), the new pause image config will be silently ignored until the next restart.

```bash
# Missing after the tee:
sudo systemctl restart crio || true
```

---

### 🐛 Bug: `kubeadm config images list` missing `--kubernetes-version` in install scripts

**Files:** `cfg/nodeextension/debian12/scripts/install-k8s-packages.sh`, `cfg/nodeextension/debian13/scripts/install-k8s-packages.sh`

Line 148 calls `kubeadm config images list` with no version flag, relying on kubeadm's own embedded default. `apply-debian-delta.sh` consistently uses `--kubernetes-version "v${KUBE_VERSION}"`. If these install scripts are ever run with a kubeadm binary that defaults to a different version than the one being installed, the wrong pause image tag will be written.

---

### 🐛 Bug: Kubeadm warning lines counted as missing images in verification loop

**File:** `lib/scripts/k2s/system/package/scripts/apply-debian-delta.sh`, lines ~130–142

The `while IFS= read -r required_image` loop reads directly from `kubeadm config images list` output without filtering non-image lines. Kubeadm regularly emits `[WARNING ...]` or `W<timestamp> ...` lines. Each such line passes the `[[ -z "$required_image" ]] && continue` guard (they're non-empty), is then parsed as a fake `image_repo`/`image_tag`, fails the `crictl images` check, and increments `MISSING_IMAGES`. This produces spurious warnings on essentially every run.

The PowerShell scripts correctly guard with `Where-Object { $_ -match '^[^\s]+/[^\s]+:[^\s]+$' }`; the same filter is needed here:

```bash
done < <(kubeadm config images list --kubernetes-version "v${KUBE_VERSION}" | grep -E '^[^[:space:]]+/[^[:space:]]+:[^[:space:]]+$')
```

---

### ⚠️ Missing validation: `$K8sVersion` could be empty in `common-setup.module.psm1`

**File:** `lib/modules/k2s/k2s.node.module/linuxnode/distros/common-setup.module.psm1`, new `K8sVersion` parameter

```powershell
[string] $K8sVersion = $kubernetesVersion
```

`$kubernetesVersion` is a module-scoped variable. If it's unset or empty when this function is called, `kubeadm config images list --kubernetes-version ` (with an empty value) will either error ambiguously or fall back to kubeadm's default. There's no guard before the command is executed. A defensive check would prevent a confusing failure:

```powershell
if ([string]::IsNullOrWhiteSpace($K8sVersion)) {
    throw 'K8sVersion is required but was not provided and $kubernetesVersion is not set.'
}
```

---

### ⚠️ `systemctl restart crio` without error handling in `apply-debian-delta.sh`

**File:** `lib/scripts/k2s/system/package/scripts/apply-debian-delta.sh`, ~line 122

If the new config file is malformed or CRI-O fails to restart for any reason, the script silently continues into `kubeadm upgrade apply`. Kubernetes upgrade will then fail with a container runtime error rather than a clear message about why. Consider at minimum logging the restart failure:

```bash
systemctl restart crio || { echo "[debian-delta][error] CRI-O failed to restart after pause image config update"; exit 1; }
```

---

### ℹ️ Minor: `kubeadm config images list` called twice in `apply-debian-delta.sh`

The command is invoked once to get the pause image (line ~111) and again for the verification loop (line ~140), both with the same `--kubernetes-version` argument. In an air-gapped environment this is harmless (kubeadm resolves locally), but capturing the output to a variable and reusing it would be cleaner and avoids any edge-case divergence.

---

## Files Reviewed
- ✅ cfg/nodeextension/debian12/scripts/install-k8s-packages.sh
- ✅ lib/scripts/k2s/system/package/New-K2sNodePackage.ps1

krotz-dieter (commented):
## PR Review: #2594 — Keep kubeadm as source of truth for pause image

The overall approach is sound: using `kubeadm config images list` as the authoritative source for K8s images (especially pause) is cleaner than scraping `crictl` output. However, there are a few issues worth addressing.

---

### 🐛 Bug — `apply-debian-delta.sh`: False-positive "all images verified" when `kubeadm` fails

```bash
done < <(kubeadm config images list --kubernetes-version "v${KUBE_VERSION}")
```

If `kubeadm config images list` fails or produces no output, the `while` loop simply never executes, `MISSING_IMAGES` stays `0`, and the script prints **"All required kubeadm images verified"** — a false positive. This could mask a broken air-gapped environment, which is exactly what this check exists to detect. The command's exit code should be captured and checked before trusting `MISSING_IMAGES`.

---

### 🐛 Bug — `apply-debian-delta.sh`: Unguarded `systemctl restart crio` aborts upgrade on failure

```bash
systemctl restart crio
```

The equivalent lines in both `install-k8s-packages.sh` files use `|| true` to treat CRI-O restart failure as non-fatal. Here it is unguarded, so a transient CRI-O restart failure (which can happen during upgrades) will abort the entire delta application. This inconsistency looks unintentional. At minimum, a failure here should log a warning and continue (matching install-script behavior), or the rationale for making it fatal during upgrade should be documented.

---

### ⚠️ Risk — `common-setup.module.psm1`: `K8sVersion` validation fires after scriptblock definition but has an unusual placement

```powershell
Function Copy-KubernetesImagesFromControlPlaneNodeToWindowsHost {
    param (
        [string] $TargetPath = $(throw 'Argument missing: TargetPath'),
        [string] $K8sVersion = $kubernetesVersion
    )

    $executeRemoteCommand = { ... }   # scriptblock definition only

    if ([string]::IsNullOrWhiteSpace($K8sVersion)) {
        throw 'K8sVersion is required...'
    }
```

This is technically safe — the scriptblock is only defined here, not invoked — but it's easy for a future contributor to add a command call above the validation. Consider moving the guard to the top of the function body, before `$executeRemoteCommand` is defined.

---

### Minor — `install-k8s-packages.sh` (both): No error guard on `kubeadm version -o short`

```bash
KUBEADM_VERSION="$(kubeadm version -o short)"
```

If `kubeadm` is not yet on `PATH` at this point, the script dies with a cryptic error. Given the block is after the k8s packages are installed, this is low risk in practice, but a `|| { echo "[InstallK8s] ERROR: kubeadm not found"; exit 1; }` would give a clearer failure message.

---

## Files Reviewed
- ✅ cfg/nodeextension/debian12/scripts/install-k8s-packages.sh
- ✅ lib/scripts/k2s/system/package/New-K2sNodePackage.ps1

krotz-dieter (commented):
## Code Review: PR #2594 — Keep kubeadm as source of truth for pause image

---

### `cfg/nodeextension/debian12/scripts/install-k8s-packages.sh` & `cfg/nodeextension/debian13/scripts/install-k8s-packages.sh`

**Bug: No guard for empty `KUBEADM_VERSION`**

```sh
KUBEADM_VERSION="$(kubeadm version -o short)"
KUBEADM_PAUSE_IMAGE="$(kubeadm config images list --kubernetes-version "$KUBEADM_VERSION" | ...)"
```

If `kubeadm version -o short` returns empty output (or produces an unexpected format), `KUBEADM_VERSION` will be an empty string. The subsequent `kubeadm config images list --kubernetes-version ""` call will then either fail loudly or produce garbage output. Because there is no `|| true` on the first line, if `set -e` is active and that command fails, the script aborts entirely. Add a guard:

```sh
if [ -z "$KUBEADM_VERSION" ]; then
    echo "[InstallK8s] WARNING: Could not determine kubeadm version; skipping pause image configuration"
else
    ...
fi
```

**Minor: `kubeadm version -o short` is deprecated** in newer kubeadm versions. Using `kubeadm version -o json | jq -r .clientVersion.gitVersion` (or similar) would be more forward-compatible.

---

### `lib/modules/k2s/k2s.node.module/linuxnode/distros/common-setup.module.psm1`

**Potential Bug: `-IgnoreErrors` parameter may not be defined in `$executeRemoteCommand`**

```powershell
$additionalImagesResult = $(&$executeRemoteCommand -Command $retrieveAdditionalImagesCmd -ReturnCommandOutput -IgnoreErrors)
```

The diff only shows `Command` and `ReturnCommandOutput` as parameters of the `$executeRemoteCommand` script block. If `-IgnoreErrors` is not a declared parameter of that script block, PowerShell without `[CmdletBinding()]` will silently ignore it, meaning errors from the flannel `crictl` command are *not* actually ignored — they could surface as terminating errors if the command fails. This should be verified against the full script block definition.

**Missing `v` prefix guard for `$K8sVersion`**

```powershell
$kubeadmImagesCmd = "sudo kubeadm config images list --kubernetes-version $K8sVersion | ..."
```

The version string is interpolated unquoted into a shell command. More critically, `apply-debian-delta.sh` explicitly prepends `v` (`"v${KUBE_VERSION}"`), but here it's used bare. If `$K8sVersion` lacks the `v` prefix (e.g., `1.29.0`), kubeadm may reject it or return an error, causing the function to throw and halt the entire copy operation.

---

### `lib/scripts/k2s/system/package/New-K2sNodePackage.ps1`

**Same `v`-prefix inconsistency as above:**

```powershell
"sudo kubeadm config images list --kubernetes-version $k8sVersion"
```

If `$k8sVersion` is `1.29.0` rather than `v1.29.0`, this fails. The `apply-debian-delta.sh` in the same PR explicitly adds the `v`, suggesting the format is not guaranteed upstream. Should be `"v$k8sVersion"` (or verify that `$k8sVersion` always includes the prefix).

**Possible export failure for unloaded images:**

`kubeadm config images list` returns images that *should* be used — not necessarily those *already pulled* on the node. If any listed image isn't in the container runtime cache, the subsequent `buildah push` will fail. The original code used `crictl images` (actual state), which was safer against this. Consider filtering kubeadm output against `crictl images` output, or handling `buildah push` failures per-image rather than allowing one failure to abort the entire export.

---

### `lib/scripts/k2s/system/package/scripts/apply-debian-delta.sh`

**Inefficiency that can cause correctness issues: `kubeadm config images list` called twice**

The script calls `kubeadm config images list --kubernetes-version "v${KUBE_VERSION}"` once for the pause image and once again for verification (stored in `KUBEADM_IMAGES_OUTPUT`). In addition to the redundant network/subprocess cost, if the first call succeeds but the second fails (transient error, timeout, air-gapped race), the verification silently increments `MISSING_IMAGES` even though the pause image was already correctly configured. Reuse `KUBEADM_IMAGES_OUTPUT` for both operations.

---

## Files Reviewed
- ✅ cfg/nodeextension/debian12/scripts/install-k8s-packages.sh
- ✅ lib/scripts/k2s/system/package/New-K2sNodePackage.ps1

PR #2595, opened 2026-06-16
#2550 tests failing according to delta upgrades

krotz-dieter:
#2550 tests failing according to delta upgrades


krotz-dieter (commented):
Here is my review of PR #2595:

---

## Code Review

**Bug: `Invoke-Ctr` called with `-CtrExePath` parameter that likely doesn't exist**

In the image import section (phase 7 / `ImageImport`):
```powershell
$importSuccess = Invoke-Ctr -Arguments '-n', 'k8s.io', 'images', 'import', $img.FullName -CtrExePath $ctrExe
```
The old code passed no `-CtrExePath` to `Invoke-Ctr`. This PR adds `-CtrExePath $ctrExe` but does **not** modify the `Invoke-Ctr` function definition anywhere in this diff. If `Invoke-Ctr` doesn't already have this parameter, PowerShell will throw `"A parameter cannot be found that matches parameter name 'CtrExePath'"`, causing Windows image import to fail entirely. This needs to be verified or the `Invoke-Ctr` definition needs to be updated in this PR.

---

**Bug: `$Manifest.WholeDirectories` — field may not exist in delta-manifest.json**

In `Copy-MissingFilesFromOldInstall`:
```powershell
$wholesaleManifestDirs = @($Manifest.WholeDirectories) | Where-Object { $_ -and ($_ -ne '') }
```
The manifest's field name is referenced nowhere else as `WholeDirectories`. If the delta manifest uses a different name (e.g., `WholeSaleDirectories`, `WholeDirectoriesReplaced`, etc.), this silently returns an empty array. The consequence: files from the old installation inside wholesale directories would not be skipped by the `$skippedWholesaleCount` path. They'd still be skipped by the `Test-Path` existence check if those files are already present in the delta, but any file that exists in the old install's wholesale directory but is absent from the delta package would be copied in — potentially bringing back stale binaries that were intentionally excluded from the delta.

---

**`Set-ConfigInstalledKubernetesVersion` — silent failure if function doesn't exist**

```powershell
Set-ConfigInstalledKubernetesVersion -Value $targetKubernetesVersion
```
This function is called in the version update phase but is not defined in this diff, and it's not a common function visible in the existing code surface. The call is inside a try/catch that only warns on failure, so if the function is missing, setup.json's Kubernetes version is silently not updated. Given that fixing Kubernetes version mismatches after delta upgrade is one of the stated goals of this PR (see the added tips in docs), this silent failure would leave the root problem unresolved.

---

**`Set-ConfigInstallFolder` called twice**

`Set-ConfigInstallFolder -Value $targetInstallPath` is called both after phase 6 (WindowsArtifacts / PATH switch) and again inside phase 14 (UpdateVersion). The second call is harmless but indicates the phase 14 call is dead code — it should be removed to avoid confusion.

---

**Phase count mismatch**

```powershell
$script:totalPhases = 14
```
The function docstring enumerates 16 phases (1–16). Counting the actual `_phase` calls in the code also yields more than 14. The progress percentage shown to the user will exceed 100% before the final phase completes. The original count was 13 and only +1 was added, but multiple new phases were actually inserted.

---

### `k2s/cmd/k2s/cmd/system/upgrade/upgrade.go`

**Mixed tab/space indentation breaks `--help` output**

In `upgradeCommandLongDescription` and `upgradeCommandExample`, the new lines use hard tab characters (`\t`) while the surrounding lines use spaces:

```
// spaces:
  The following tasks will be executed:
  1. Detect delta package root
  2. Detect existing installation folder
// tabs (new lines):
	3. Update all Debian packages from delta...
	4. Stop the cluster if it was running
```

And in the example block:
```
// tabs:
	Expand-Archive k2s-delta-v1.5.0-to-v1.6.0.zip -Destination C:\k2s-v1.6.0
	cd C:\k2s-v1.6.0
// spaces:
  .\k2s.exe system upgrade
```

This renders as misaligned output in `k2s system upgrade --help`, making the instructions hard to read. All indentation in these raw string literals should use consistent spaces.

---

### `lib/scripts/k2s/system/package/scripts/apply-debian-delta.sh`

The final fallback in `read_target_kubernetes_version`:
```bash
version=$(kubelet --version 2>/dev/null | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' || echo "")
```
This fallback runs **after** apt has already upgraded kubelet to the target version, so at the point `kubeadm upgrade apply` runs, `kubelet --version` would reflect the new version. The fallback is therefore safe, but it should be noted in a comment that this relies on apt having already upgraded kubelet before this function is called, since the intent of the whole function is to discover the TARGET version.

---

### Remaining files

- **`docs/op-manual/delta-packages.md`**: New step 2 code block uses 2-space indent instead of 4-space, inconsistent with other blocks in the file. Not a functional issue.
- **`docs/op-manual/upgrading-k2s.md`**: Accurate, consistent with behavioral changes.
- **`lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1`**: `TargetKubernetesVersion` injection into the Debian manifest looks correct.
- **`lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1`**: Variable rename from `$targetInstallPath` → `$existingInstallPath` is clean and accurate.
- **`lib/scripts/k2s/system/upgrade/Start-ClusterUpgrade.ps1`**: Log message change only, no issues.

---

## Files Reviewed
- ✅ docs/op-manual/delta-packages.md
- ✅ docs/op-manual/upgrading-k2s.md
- ✅ k2s/cmd/k2s/cmd/system/upgrade/upgrade.go
- ✅ lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1
- ✅ lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1
- ✅ lib/scripts/k2s/system/upgrade/Start-ClusterUpgrade.ps1

krotz-dieter (commented):
## PR Review: #2595 — Delta upgrade: extract-to-final-dir model

This PR changes the delta upgrade model so the extracted delta package directory *is* the final installation directory, rather than files being copied into the existing install location. The approach is conceptually sound, but there are real bugs in the implementation.

---

### 🐛 Bug: `Invoke-Ctr -CtrExePath` parameter may not exist in the old installation

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1` (~line 1145)

```powershell
$importSuccess = Invoke-Ctr -Arguments '-n', 'k8s.io', 'images', 'import', $img.FullName -CtrExePath $ctrExe
```

`Invoke-Ctr` is **not** defined in the delta package's `update.module.psm1`; it comes from `k2s.cluster.module` loaded from the **existing (old) installation** (`$existingInstallPath`), as set up in `Start-ClusterUpdate.ps1`. If the old installation's `Invoke-Ctr` does not have a `-CtrExePath` parameter, PowerShell will throw `"A parameter cannot be found that matches parameter name 'CtrExePath'"`, causing the entire Windows image-import phase to fail.

After the `Set-ConfigInstallFolder` switch in phase 6, `Get-KubeBinPath` now resolves to the new installation directory, so `Invoke-Ctr` without `-CtrExePath` would already use the correct `ctr.exe`. The explicit path parameter is the right *intent*, but it should be guarded:

```powershell
# Safe approach: invoke ctr.exe directly rather than via the old Invoke-Ctr wrapper
$importResult = & $ctrExe '-n' 'k8s.io' 'images' 'import' $img.FullName
```

---

### 🐛 Bug: `Set-ConfigInstalledKubernetesVersion` silently fails if absent in old install

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1` (~line 1281)

```powershell
Set-ConfigInstalledKubernetesVersion -Value $targetKubernetesVersion
```

This function is loaded from the existing installation's infra module. If it was introduced in the same release as this PR (i.e., it doesn't exist in older `k2s.infra.module`), the call will throw, be caught by the phase-14 `catch` block, and logged as a `[Warn]`. The `setup.json` will then have a stale Kubernetes version — the very issue this PR is trying to fix. The tip in the docs ("inspect the Debian delta log for the `Target Kubernetes version` line") would become the only mitigation, but setup.json being wrong is a systemic problem.

**Fix:** Check if the function exists before calling it:
```powershell
if (Get-Command -Name Set-ConfigInstalledKubernetesVersion -ErrorAction SilentlyContinue) {
    Set-ConfigInstalledKubernetesVersion -Value $targetKubernetesVersion
} else {
    Write-Log '[Update][Warn] Set-ConfigInstalledKubernetesVersion not available in existing installation; manual setup.json update may be required.' -Console
}
```

---

### ⚠️ Logic gap: `containerd` root not in new PATH entries

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`, `Set-K2sMachinePathEntries`

`$oldEntries` includes `Join-Path $oldRoot 'containerd'`, but `$newEntries` does not include the equivalent for the new root:

```powershell
$oldEntries = @(
    ...
    (Join-Path $oldRoot 'containerd'),       # <-- in old
    (Join-Path $oldRoot 'bin\containerd')
)
$newEntries = @(
    ...
    # (Join-Path $newRoot 'containerd')      # <-- missing for new
    (Join-Path $newRoot 'bin\containerd')
)
```

If any existing installation had `<old_root>\containerd` in `PATH` (rather than `<old_root>\bin\containerd`), it will be removed from the machine `PATH` and not re-added under the new root. This would silently leave containerd binaries unreachable in a new shell. If `containerd` at the root level is deprecated intentionally, there should be a comment to that effect.

---

### ⚠️ `Start-ClusterNode` removed from restart path without fallback

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1` (~line 1086)

```powershell
# BEFORE (truncated):
if (Get-Command -Name Start-ClusterNode ...) {
    Start-ClusterNode -SetupName $setupInfo.Name -ShowLogs:$ShowLogs
} else {
    # fallback to k2s.exe start
}

# AFTER:
$k2sExe = Join-Path $targetInstallPath 'k2s.exe'
if (Test-Path -LiteralPath $k2sExe) {
    & $k2sExe @argsCall
} else {
    throw "k2s.exe not found at $k2sExe"
}
```

Using the new `k2s.exe` directly is the right call since modules are still loaded from the old install. However, `k2s start` via the new binary may invoke the same `Start-ClusterNode` internally, so this is likely fine. That said, the `-o` flag (mapped from `$ShowLogs`) should be validated — if `k2s start` doesn't accept `-o`, this will silently fail or error. The original `Start-ClusterNode -ShowLogs` parameter name was unambiguous.

---

### ℹ️ Docs: MkDocs admonition indentation

**File:** `docs/op-manual/delta-packages.md`

The `!!! warning` and `!!! tip` content blocks use 2-space indentation instead of the 4-space indentation required by MkDocs material admonitions:

```markdown
!!! warning "Keep the extracted directory"
  After a successful delta upgrade...   ← 2 spaces, should be 4
```

These will likely render as plain paragraphs rather than styled admonition boxes.

---

## Files Reviewed
- ✅ docs/op-manual/delta-packages.md
- ✅ docs/op-manual/upgrading-k2s.md
- ✅ k2s/cmd/k2s/cmd/system/upgrade/upgrade.go
- ✅ lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1
- ✅ lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1
- ✅ lib/scripts/k2s/system/upgrade/Start-ClusterUpgrade.ps1

krotz-dieter (commented):

---

## PR Review: #2595 — Delta upgrade flow rework

This PR fundamentally changes the delta update model from "copy delta files **into** the existing installation directory" to "complete the delta directory **from** the existing installation and switch `InstallFolder` to it." The approach is sound but there are several concrete issues.

---

### 🐛 Bug: `k2s.exe start --output` is likely an invalid flag

```powershell
$argsCall = @('start')
if ($ShowLogs) { $argsCall += '--output' }
& $k2sExe @argsCall
if ($LASTEXITCODE -ne 0) { throw "k2s.exe start returned exit code $LASTEXITCODE" }
```

The previous code called `Start-ClusterNode -ShowLogs:$ShowLogs` (a PowerShell function) or `& $k2sExe start` without any log flag. This PR now passes `--output` to `k2s.exe start` when `$ShowLogs` is true. If `k2s start` does not have an `--output` flag (nowhere else in this diff is such a flag referenced), the restart will fail with a non-zero exit code whenever `$ShowLogs` is set, causing the update to abort after the cluster has already been stopped and all files replaced. The old fallback at least didn't pass any flag.

---

### 🐛 Bug: Partial upgrade state when `wait_for_kubernetes_server_version` fails

**File:** `lib/scripts/k2s/system/package/scripts/apply-debian-delta.sh`

```bash
if kubeadm upgrade apply "v${KUBE_VERSION}" ... 2>&1; then
    echo "[debian-delta] kubeadm upgrade completed successfully"
    if ! wait_for_kubernetes_server_version "$KUBE_VERSION"; then
        if [[ -d "$IMAGES_DIR" ]]; then
            rm -rf "$IMAGES_DIR"
        fi
        exit 1   # <-- new hard exit
    fi
```

When `kubeadm upgrade apply` succeeds but `wait_for_kubernetes_server_version` times out (120 s / 24 × 5 s), the script exits 1. At this point:
- The Linux VM is fully upgraded (Debian packages, kubeadm config migration)
- `PerformClusterUpdate` on the Windows side will treat the debian-delta phase as failed and may abort before completing Windows artifacts / switching `InstallFolder`

The system is left in an irrecoverable partially-upgraded state: Linux at the new Kubernetes version, Windows still at the old version. No rollback path is documented or implemented. Before this PR the script continued past version-check warnings; now it hard-fails. Either add a rollback mechanism or at minimum document the manual recovery steps when this happens.

---

### 🐛 Bug: `kubelet` Windows service `Application` path not retargeted

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`, function `Set-K2sWindowsServiceInstallPathEntries`

`kubelet` is included in the `AppParameters` update loop:
```powershell
foreach ($serviceName in @('containerd', 'dnsproxy', 'flanneld', 'httpproxy', 'kubelet', 'kubeproxy', 'windows_exporter')) {
```
but it is **absent** from `$serviceApplications` (the map that updates the `Application` property — the actual executable path). If kubelet is managed by nssm with an `Application` pointing into the old install root, it will continue launching from the old directory after `InstallFolder` is switched. The `AppParameters` update alone is insufficient for a path change.

---

### ⚠️ Risk: `Set-ConfigInstallFolder` is called but not defined in this diff

```powershell
Set-ConfigInstallFolder -Value $targetInstallPath
```

This is the single most critical function call in the PR (it finalises the switch to the new install directory in `setup.json`). It is not defined anywhere in the diff. The parallel `Set-ConfigProductVersion` follows the same pattern so presumably it exists in the infra module — but if `Set-ConfigInstallFolder` does not exist or was never implemented in the infra module, the try/catch block will throw and the update will fail after files have already been copied and the cluster stopped.

---

### ⚠️ Risk: `Copy-MissingFilesFromOldInstall` silently captures `$consoleSwitch` from outer scope

```powershell
function Copy-MissingFilesFromOldInstall {
    param([string]$SourceRoot, [string]$DestinationRoot, [object]$Manifest)
    ...
    Write-Log (...) -Console:$consoleSwitch   # $consoleSwitch not a param
}
```

`$consoleSwitch` is not a parameter; it relies on PowerShell's dynamic scope lookup from `PerformClusterUpdate`. While this works today, it is fragile: if the function is ever extracted or `$consoleSwitch` is renamed, the `-Console:` switch silently resolves to `$false` (null coercion) rather than erroring. `$consoleSwitch` should be passed as a parameter.

---

### ℹ️ Minor: Phase 7 ("Clean deprecated kubelet flags") listed in description but counter may be off

The `.DESCRIPTION` block lists 14 phases with a distinct "7. Clean deprecated Windows kubelet flags" phase, and `$script:totalPhases` was bumped from 13 to 14. Without seeing the complete function body it cannot be confirmed that exactly 14 `_phase` calls exist; if the deprecated-flags cleanup runs inside phase 6 without its own `_phase` call the progress indicator will be off.

---

## Files Reviewed

- ✅ docs/op-manual/delta-packages.md
- ✅ docs/op-manual/upgrading-k2s.md
- ✅ k2s/cmd/k2s/cmd/system/upgrade/upgrade.go
- ✅ lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1
- ✅ lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1
- ✅ lib/scripts/k2s/system/upgrade/Start-ClusterUpgrade.ps1

PR #2604, opened 2026-06-18
#2550 New delta package implementation to solve test failures

krotz-dieter:
#2550 New delta package implementation to solve test failures


krotz-dieter (commented):

---

## PR #2604 Code Review

**1. Services left stopped on seeding failure (no pre-re-home recovery)**

In the relocation branch inside the `try` block, K2s-managed services are explicitly stopped before `Copy-UnchangedInstallationFiles` is called. If seeding fails (`throw 'Failed to seed new installation folder from previous installation'`), `$relocationDone` is still `$false`, so the `catch` block skips the rollback entirely and just re-throws:

```powershell
# services stopped here
foreach ($svcName in (Get-K2sManagedServiceName)) { Stop-Service ... }
Stop-Process -Name 'containerd' ...

# if this throws → catch block does nothing (relocationDone = false)
if (-not (Copy-UnchangedInstallationFiles ...)) {
    throw 'Failed to seed new installation folder from previous installation'
}
```

The services are now stopped, the installation still points to the old path (nothing was re-homed yet), and there is no automatic restart. If any earlier update phase leaves the cluster running into phase 6, this leaves K2s in a broken stopped state with no recovery attempt. At minimum, a `finally` or a pre-check guard should restart the old services if `$relocationDone` is false but services were stopped.

**2. `$ctrExe` passed to `Invoke-Ctr` as `-CtrExePath` — needs verification**

```powershell
$importSuccess = Invoke-Ctr -Arguments '-n', 'k8s.io', 'images', 'import', $img.FullName -CtrExePath $ctrExe
```

This adds a `-CtrExePath` parameter to `Invoke-Ctr` that wasn't previously used. I was unable to fetch the `Invoke-Ctr` function definition to verify this parameter exists, because the repository is on an internal GHE instance. **Please confirm `Invoke-Ctr` accepts `-CtrExePath`** — if it doesn't, PowerShell will silently ignore the parameter (or error, depending on CmdletBinding), meaning the relocation case will call `ctr.exe` from the wrong path (falling back to whatever was resolved previously).

**3. `Set-K2sInstallationHome`: Containerd config regeneration ignores `$FromPath`/`$ToPath` substitution**

The function does a regex path-replacement for nssm services and `StartKubelet.ps1`, but for containerd it instead calls:

```powershell
Set-RootPathForImagesInConfig $containerdTomlPath | Out-Null
Set-InstallationDirectory $containerdTomlPath | Out-Null
Set-UserTokenForRegistryInConfig $containerdTomlPath | Out-Null
```

These functions presumably derive the path from the current K2s setup state (e.g., `Get-KubePath`/`Get-ConfigInstallFolder`). The problem is that this is called **from within** `Set-K2sInstallationHome` mid-update, at a point where `setup.json` `InstallFolder` has already been updated to `$ToPath` by step 5 (just above, in the same function). If these containerd helper functions internally read the install folder from `setup.json` after step 5 updates it, they will correctly regenerate the config for `$ToPath`. But if the rollback path calls this function (swapping `$FromPath`/`$ToPath`) and by then `setup.json` already points to the rollback target, this should also work. This is subtle enough to warrant a comment confirming the ordering invariant.

---

### `lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1`

**4. Default `$NssmPath` references an unguarded module-level variable**

```powershell
[string] $NssmPath = "$kubeBinPath\nssm.exe"
```

`$kubeBinPath` is a module-scope variable. When `services.module.psm1` is freshly imported (e.g., from `Set-K2sInstallationHome` in a new PS session context), `$kubeBinPath` might be empty/null if the module initialiser didn't run or ran before K2s paths were configured. In that case the default resolves to `\nssm.exe`. All current callers pass `-NssmPath` explicitly, so this isn't currently broken, but the footgun should be guarded:

```powershell
[string] $NssmPath = $(if ($kubeBinPath) { "$kubeBinPath\nssm.exe" } else { 'nssm.exe' })
```

**5. `$OldPath` not trimmed in `Update-NssmServiceInstallPath`**

Unlike `Set-K2sInstallationHome` which does `$FromPath = $FromPath.TrimEnd('\')`, `Update-NssmServiceInstallPath` receives `$OldPath` as-is. Since the function is now exported (callable by external scripts), a caller passing a path with trailing backslash will silently fail to match any registry values. The function should trim its inputs, or the doc comment should explicitly state the precondition.

---

### `lib/modules/k2s/k2s.cluster.module/update/update.module.relocation.unit.tests.ps1`

The unit tests for `Copy-UnchangedInstallationFiles` and `Remove-DeltaPackageArtifact` are solid and correctly validate the key behaviors. The `Copy-UnchangedInstallationFiles` tests correctly use `$TestDrive` for isolation.

**Minor gap**: `Set-K2sInstallationHome` has zero test coverage. The rollback symmetry claim (`calling it with swapped FromPath/ToPath reverses the re-home`) is entirely untested.

---

### `lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1`

The test `'returns an empty result when the nssm registry parameters are missing'` mocks `Test-Path` to globally return `$false` within the module scope. This also intercepts the `-LiteralPath $regKey` check but it could unexpectedly intercept `Test-Path` calls in module initialization if any exist. Consider scoping the mock more narrowly or using a `-ParameterFilter`.

---

### `lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1` / `docs/op-manual/delta-packages.md`

Documentation and comment-only changes; look accurate and consistent with the implementation.

---

## Files Reviewed

- ✅ docs/op-manual/delta-packages.md
- ✅ lib/modules/k2s/k2s.cluster.module/update/update.module.relocation.unit.tests.ps1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1
- ✅ lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1

krotz-dieter (commented):

---

## Review of PR #2604 — Delta Package Re-home Implementation

This is a substantial change implementing installation relocation during delta updates. The overall design and rollback logic are sound. I found one confirmed bug and two unverifiable concerns that should be double-checked before merge.

---

### 🐛 Confirmed Bug: `Update-NssmServiceInstallPath` silently ignores nssm failure

**File:** `lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1`

```powershell
$newValue = [regex]::Replace($current, $oldPathPattern, $NewPath, ...)
Write-Log ("Re-pointing service '{0}' parameter '{1}' to new installation path" -f $Name, $parameter)
& $NssmPath set $Name $parameter $newValue | Out-Null
$previousValues[$parameter] = $current   # <-- recorded unconditionally
```

`$LASTEXITCODE` is not checked after `& $NssmPath set`. If nssm returns a non-zero exit code (e.g. service parameters registry key is ACL-protected, nssm version mismatch, or service in a locked state), the function silently records `$parameter` in `$previousValues` as if the update succeeded. The caller (`Set-K2sInstallationHome`) then believes all service parameters have been re-pointed, while the affected service still points to the old installation path. After a relocation, that service will fail to start from the new folder.

Fix: check `$LASTEXITCODE` after the nssm call and skip recording / log a warning:
```powershell
& $NssmPath set $Name $parameter $newValue | Out-Null
if ($LASTEXITCODE -ne 0) {
    Write-Log ("[Update][Warn] nssm failed (exit {0}) re-pointing '{1}' parameter '{2}'" -f $LASTEXITCODE, $Name, $parameter)
    continue
}
$previousValues[$parameter] = $current
```

---

### ⚠️ Concern: `Invoke-Ctr -CtrExePath` parameter may not exist

```powershell
$importSuccess = Invoke-Ctr -Arguments '-n', 'k8s.io', 'images', 'import', $img.FullName -CtrExePath $ctrExe
```

The PR adds `-CtrExePath $ctrExe` to the `Invoke-Ctr` call. The original code called `Invoke-Ctr` without this parameter, implying the function previously resolved the ctr.exe path internally. If `Invoke-Ctr` does **not** accept `-CtrExePath`, this change will break the image-import phase for **both** the relocation and in-place code paths (the parameter is passed unconditionally). Please verify `Invoke-Ctr`'s signature includes `-CtrExePath`.

---

### ⚠️ Concern: Containerd config helper call signatures

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1` — `Set-K2sInstallationHome`

```powershell
Import-Module $containerdModule -Force
Set-RootPathForImagesInConfig $containerdTomlPath | Out-Null
Set-InstallationDirectory $containerdTomlPath | Out-Null
Set-UserTokenForRegistryInConfig $containerdTomlPath | Out-Null
```

These three functions each receive `$containerdTomlPath` as a positional argument. If they do not accept a config-file path as their first (or only) positional parameter — for example, if they derive the path internally from module-level state — then passing `$containerdTomlPath` here would be silently ignored or cause an error. The comment explains the ordering invariant correctly, but if the functions' actual signatures differ from the assumed ones, containerd config regeneration will silently produce wrong results. Please confirm the signatures of these helpers.

---

### Minor: No test for nssm failure path in `Update-NssmServiceInstallPath`

**File:** `lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1`

There is no test for the case where nssm itself returns a non-zero exit code. Given the confirmed bug above, adding such a test would help prevent regressions.

---

## Files Reviewed
- ✅ docs/op-manual/delta-packages.md
- ✅ lib/modules/k2s/k2s.cluster.module/update/update.module.relocation.unit.tests.ps1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1
- ✅ lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1

krotz-dieter (commented):
Now I have enough information to write a thorough review.

Here is my review:

---

## Review: PR #2604 — New delta package implementation (installation relocation)

This is a substantial, well-structured PR. The relocation logic is carefully designed with clear forward/rollback paths. A few concrete issues are noted below.

---

**`& $oldK2sExe start | Out-Null` silently swallows the exit code in both rollback/recovery paths**

In the `$relocate && $relocationDone` rollback branch (around line 1380):
```powershell
if (Test-Path -LiteralPath $oldK2sExe) { & $oldK2sExe start | Out-Null }
```
And in the `$relocateServicesStopped` recovery branch (around line 1396):
```powershell
if (Test-Path -LiteralPath $oldK2sExe) { & $oldK2sExe start | Out-Null }
```

`| Out-Null` discards stdout/stderr *and* does not capture `$LASTEXITCODE`, so a failed restart is entirely invisible. The success message `'Windows-side installation reverted to the previous folder.'` / `'Previous installation restarted.'` is then unconditionally printed, creating a false impression of a successful recovery. The fix is to check `$LASTEXITCODE` after the call:
```powershell
& $oldK2sExe start 2>&1 | Out-Null
if ($LASTEXITCODE -ne 0) {
    Write-Log ("[Update][Rollback][Warn] k2s start returned exit code $LASTEXITCODE" ...) -Console
}
```

**Rollback after a successful phase-7 restart doesn't stop the already-running cluster before re-pointing configs**

When a failure happens in phases 8–12 (after the cluster was restarted from `$newInstallPath` in phase 7), the catch block calls `Set-K2sInstallationHome -FromPath $newInstallPath -ToPath $oldInstallPath` — which re-points nssm registry entries, containerd config, KUBECONFIG and setup.json — and then immediately calls `$oldK2sExe start`. At this point the cluster's services are already running (from `$newInstallPath` binaries). There is no `k2s stop` call before `k2s start`, which may result in `k2s start` failing because services are already registered as running, leaving the system with services running from `$newInstallPath` but all config pointing at `$oldInstallPath`. Consider adding an explicit stop before the restart in the `$relocationDone` rollback path.

---

### `lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1`

**Single quote in `$Proxy` breaks the embedded bash `printf` command**

In `Initialize-DebAcquisitionEnvironment`:
```powershell
$proxyFmt = "Acquire::http::Proxy `"$Proxy`";\nAcquire::https::Proxy `"$Proxy`";\n"
$proxyLine = "printf '$proxyFmt' | sudo tee /etc/apt/apt.conf.d/proxy.conf > /dev/null || true"
```
`$proxyFmt` is embedded inside single quotes in the bash `printf` call. If `$Proxy` contains a single quote (e.g. a URL with an unusual character or a corporate proxy with embedded auth), the generated bash script is syntactically invalid. Because the script runs without `-e`, this silent failure means apt is never configured to use the proxy, causing the package download to fail with a non-obvious error. Either validate `$Proxy` at the call site or escape single quotes before embedding: `$Proxy.Replace("'", "'\''")`.

---

### `lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1`

**The nssm stub file uses a Windows batch shebang on a potentially non-Windows CI runner**

```powershell
Set-Content -Path $nssmStub -Value "@echo off`r`nexit /b 0" -NoNewline
```
The stub is a `.cmd` batch file. If the unit tests are run on a Linux CI agent (which the tags `'unit', 'ci'` suggest is possible), invoking a `.cmd` file will fail. The test is platform-specific without any `[System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform` guard or a platform-appropriate stub. This would cause the `'trims trailing backslashes'` and `'non-zero exit code'` tests to fail on Linux CI.

---

## Files Reviewed
- ✅ docs/op-manual/delta-packages.md
- ✅ k2s/cmd/k2s/cmd/system/package/package.go
- ✅ lib/modules/k2s/k2s.cluster.module/update/update.module.relocation.unit.tests.ps1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.HyperV.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1
- ✅ lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1

krotz-dieter (commented):
Based on the full diff, here is my review analysis:

---

## Review: PR #2604 — New delta package implementation

This is a substantial, well-thought-out change. The overall design (relocation vs. in-place, rollback, seeding) is coherent. Below are issues found.

---

### 🔴 Potential Runtime Bug: `Invoke-Ctr -CtrExePath` parameter

```diff
-$importSuccess = Invoke-Ctr -Arguments '-n', 'k8s.io', 'images', 'import', $img.FullName
+$importSuccess = Invoke-Ctr -Arguments '-n', 'k8s.io', 'images', 'import', $img.FullName -CtrExePath $ctrExe
```

`-CtrExePath` is added to the `Invoke-Ctr` call, but the function definition is not modified anywhere in this PR. If `Invoke-Ctr` (defined elsewhere in `update.module.psm1`) does **not** accept a `-CtrExePath` parameter, PowerShell will throw `A parameter cannot be found that matches parameter name 'CtrExePath'` and Windows image import will fail entirely for relocating updates. This must be verified against the existing `Invoke-Ctr` definition.

---

### 🔴 Partial Re-home Leaves Services in Inconsistent State

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`, `Set-K2sInstallationHome`

`Set-K2sInstallationHome` updates nssm services (step 1), `StartKubelet.ps1` (step 2), containerd config (step 3), `KUBECONFIG` (step 4), and `setup.json` (step 5) sequentially. If it fails partway through — e.g., nssm successfully re-points `containerd` and `kubelet` but `Set-ConfigInstallFolder` throws — the function returns `$false` and `$relocationDone` stays `$false`.

The catch block then takes the `$relocateServicesStopped` recovery path and attempts to restart the old `k2s.exe`. But at that point the services already point to the new folder while `setup.json` still points at the old folder. This inconsistency can cause the recovered cluster to fail to start with hard-to-diagnose errors, without any warning that nssm configs were partially re-pointed.

A minimal mitigation: log which service parameters were successfully re-pointed before the failure so an operator knows what needs manual attention. The `Update-NssmServiceInstallPath` already returns a `$previousValues` map for each service — but that return value is discarded with `| Out-Null` in `Set-K2sInstallationHome`.

---

### 🟡 Containerd Config Regeneration Ordering Assumption

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`, `Set-K2sInstallationHome` step 3

The comment states:

> the containerd regen helpers derive the installation path from the module-level `$kubePath = Get-KubePath` captured when containerd.module is imported, which resolves from the module's own `$PSScriptRoot` (i.e. `$ToPath` here), NOT from setup.json

`Set-InstallationDirectory` and `Set-RootPathForImagesInConfig` are called before `Set-ConfigInstallFolder` (step 5). If this assumption is wrong — i.e., if any of those functions call `Get-KubePath` which reads `setup.json` — `config.toml` would be generated with the **old** installation path, silently producing a broken containerd config. This should be verified by reviewing `containerd.module.psm1`, specifically whether `$kubePath` is initialized from `$PSScriptRoot` or from `Get-KubePath`.

---

### 🟡 `set -euo pipefail` → `set -uo pipefail`: Silent Directory Failure

**File:** `lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1`

```diff
-'set -euo pipefail',
+'set -uo pipefail',
```

With `-e` removed, if `mkdir -p $RemoteDir` fails (e.g., permissions), execution continues. The subsequent `cd $RemoteDir` then silently fails, and all download paths resolve relative to the current working directory instead of the intended remote dir. This would produce confusing diagnostics with no clear indication of the root cause. At minimum, the `mkdir` line should retain `|| exit 1`.

---

### 🟡 `Update-NssmServiceInstallPath` Return Value Discarded

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`, `Set-K2sInstallationHome`

```powershell
Update-NssmServiceInstallPath -Name $svc -OldPath $FromPath -NewPath $ToPath -NssmPath $nssmPath | Out-Null
```

`Update-NssmServiceInstallPath` returns a `$previousValues` hashtable documenting which parameters were successfully changed (used for rollback). This is discarded. While the current rollback calls `Set-K2sInstallationHome` with swapped paths (which regenerates everything), discarding the previous values makes it impossible to detect partial nssm failures silently and removes a useful recovery audit trail. At a minimum, consider logging the count of re-pointed services.

---

### 🟢 Minor: No Trailing Newline at End of `services.module.psm1`

**File:** `lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1`

Both the old and new version end without a trailing newline (`\ No newline at end of file`). Should be fixed.

---

## Files Reviewed

- ✅ docs/op-manual/delta-packages.md
- ✅ k2s/cmd/k2s/cmd/system/package/package.go
- ✅ lib/modules/k2s/k2s.cluster.module/update/update.module.relocation.unit.tests.ps1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.HyperV.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1
- ✅ lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1

krotz-dieter (commented):
Now I have enough information to write a thorough review.

---

## PR Review: #2550 New delta package implementation

### Overall Assessment
This is a substantial PR implementing installation relocation ("re-homing") for delta package upgrades. The architecture is sound but there is one confirmed bug in the containerd config regeneration path.

---

### Bug: `Set-K2sInstallationHome` — `containerd config.toml` regenerated with old installation path

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`, lines 547–556

```powershell
Import-Module $containerdModule -Force
Set-RootPathForImagesInConfig $containerdTomlPath | Out-Null
Set-InstallationDirectory $containerdTomlPath | Out-Null
```

`Set-InstallationDirectory` (in `containerd.module.psm1`) uses the module-scope variable `$kubePath = Get-KubePath` evaluated at import time. `Get-KubePath` is defined in `path.module.psm1` and reads `$PSScriptRoot` to navigate up to the install root. However:

1. `path.module.psm1` from the **old** installation is already loaded into the PowerShell session (imported at startup via `$infraModule` in `Start-ClusterUpdate.ps1`).
2. The nested `Import-Module $pathModule` inside `containerd.module.psm1` is called **without** `-Force`. PowerShell matches by module name (`path.module`) and skips the reimport.
3. Therefore `Get-KubePath` uses the **old** `$PSScriptRoot`, returns the old install path, and `$kubePath` in `containerd.module.psm1`'s scope is set to the old path.

The consequence: `Set-InstallationDirectory` writes the **old installation path** into the new installation's `config.toml`. Concretely, `cfg/containerd/config.toml` ends up with:
```toml
bin_dir = "C:\\k\\bin\\cni"      # should be C:\\k2s\\1.9.0\\bin\\cni
conf_dir = "C:\\k\\cfg\\containerd\\cni\\conf"
```
after regeneration from the template (`cfg/containerd/config.toml.template` uses `%INSTALLATION_DIRECTORY%`). This would break CNI / pod networking after relocation.

The inline comment at lines 538–544 asserts that `Get-KubePath` resolves from the module's own `$PSScriptRoot` (`$ToPath`) — this claim is only true if `path.module.psm1` is NOT already cached. In an update run it always is.

**Fix:** Before importing `$containerdModule`, explicitly force-reimport the path module from `$ToPath`:
```powershell
$pathModule = Join-Path $ToPath 'lib\modules\k2s\k2s.infra.module\path\path.module.psm1'
if (Test-Path -LiteralPath $pathModule) { Import-Module $pathModule -Force }
Import-Module $containerdModule -Force
```

---

### Minor: `$LASTEXITCODE` logged after `Invoke-Ctr` call is unreliable

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`, line 1265

```powershell
$importSuccess = Invoke-Ctr -Arguments '...', $img.FullName -CtrExePath $ctrExe
if ($importSuccess) { ... }
if ($attempt -lt $maxRetries) {
    Write-Log ("...Attempt {0} failed (exit code {1})..." -f $attempt, $LASTEXITCODE, ...)
```

`Invoke-Ctr` captures the ctr exit code internally in `$ctrExitCode` but doesn't expose it. By the time the log line runs, `$LASTEXITCODE` may reflect a different command. The logged exit code will be misleading for debugging. Not a functional bug, but worth noting.

---

### Minor: `services.module.psm1` — module-scope `$kubeBinPath` default in `Update-NssmServiceInstallPath`

**File:** `lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1`, line ~175

```powershell
[string] $NssmPath = $(if ($kubeBinPath) { "$kubeBinPath\nssm.exe" } else { 'nssm.exe' })
```

`$kubeBinPath` is set at module load time from `Get-KubeBinPath`, which (due to the same module caching issue as above) resolves to the old install path when the module is imported from `$ToPath`. The default value would be wrong. In practice this doesn't matter since all callers in this PR pass `$NssmPath` explicitly — but if this function is called without `$NssmPath` in future code, it will silently use the wrong nssm.exe.

---

### Observation: Duplicate cluster-stop in relocation branch

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`, lines 885–897

Phase 2 (stop cluster) already calls `k2s.exe stop` which stops all K2s-managed services. Phase 6 (relocation branch) then iterates `Get-K2sManagedServiceName` and calls `Stop-Service` on each. This is harmless and intentionally defensive, but the comment at line 883 implies this is the first stop — which might be confusing during debugging.

---

## Files Reviewed

- ✅ docs/op-manual/delta-packages.md
- ✅ k2s/cmd/k2s/cmd/system/package/package.go
- ✅ lib/modules/k2s/k2s.cluster.module/update/update.module.relocation.unit.tests.ps1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.HyperV.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1
- ✅ lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1

krotz-dieter (commented):
Now let me check the Go package.go proxy handling and the containerd module:

Now I have enough information to write the review.

---

## Code Review: PR #2604 — New delta package implementation

### Bug 1: Proxy value not escaped in `package.go`

**File:** `k2s/cmd/k2s/cmd/system/package/package.go`, line 279

```go
params = append(params, " -Proxy "+proxy)
```

Every other user-supplied string in the same function (e.g., `-InputPackageOne`, `-InputPackageTwo`, `-CertificatePath`, `-Password`) is wrapped with `utils.EscapeWithSingleQuotes(...)`. The proxy value is not, so a proxy URL containing a space or a PowerShell-special character (e.g., `http://proxy.corp.com:8080`) would silently break the generated PowerShell command. The full-package path on line 289 has the same pre-existing issue, but that is not a reason to repeat it. Should be:

```go
params = append(params, " -Proxy "+utils.EscapeWithSingleQuotes(proxy))
```

---

### Bug 2: Rollback silently ignores failure of `Set-K2sInstallationHome`

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`, catch block (line ~1403 in the PR branch)

```powershell
Set-K2sInstallationHome -FromPath $newInstallPath -ToPath $oldInstallPath -ShowLogs:$ShowLogs | Out-Null
```

`Set-K2sInstallationHome` signals failure by returning `$false` (not by throwing). Piping to `Out-Null` silently discards that signal. If the re-home-back fails (e.g., services module is gone, `Set-ConfigInstallFolder` throws internally and is caught inside the function), the code continues unconditionally to `& $oldK2sExe start`, attempting to restart the cluster in a state where:

- Services may still point to the new folder
- `setup.json` `InstallFolder` may still point to the new folder
- `KUBECONFIG` may still point to the new folder

The user is then given no indication that the rollback itself failed, and the subsequent `k2s start` is likely to silently diverge. At minimum, the return value should be checked and a `[Rollback][Error]` entry logged before proceeding:

```powershell
$rollbackOk = Set-K2sInstallationHome -FromPath $newInstallPath -ToPath $oldInstallPath -ShowLogs:$ShowLogs
if (-not $rollbackOk) {
    Write-Log '[Update][Rollback][Error] Set-K2sInstallationHome returned false; installation state may be inconsistent. Manual recovery may be required.' -Console
}
```

---

## Files Reviewed

- ✅ docs/op-manual/delta-packages.md
- ✅ k2s/cmd/k2s/cmd/system/package/package.go
- ✅ lib/modules/k2s/k2s.cluster.module/update/update.module.relocation.unit.tests.ps1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.HyperV.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1
- ✅ lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1

krotz-dieter (commented):
Now I have enough context to write the review. Let me summarize my findings:

I have a complete picture now. Let me write the review.

---

## Code Review: PR #2604 — Delta package relocation

Overall this is a well-designed, carefully implemented PR. The relocation logic is thoughtful, the rollback path is considered, and the test coverage is meaningful. I have one confirmed bug and a few minor observations.

---

### 🐛 Bug: Wrong recovery branch when `Set-K2sInstallationHome` fails at step 5

`Set-K2sInstallationHome` returns `$false` in three cases. Two are safe early-exits (services module or nssm.exe not found – no changes made). The third is when `Set-ConfigInstallFolder` (step 5) throws after steps 1–4 have **already been applied**:

- nssm service parameters re-pointed to `$newInstallPath`
- `StartKubelet.ps1` patched
- containerd `config.toml` regenerated
- `KUBECONFIG` machine env var updated

Since the `throw` at the call site fires before `$relocationDone = $true` is set, the catch block falls into the `elseif ($relocate -and $relocateServicesStopped)` branch and attempts to **restart the old installation**. But nssm services (and `KUBECONFIG`) now point to the new folder — so `$oldK2sExe start` will likely fail or produce inconsistent behaviour.

```powershell
# update.module.psm1 – caller of Set-K2sInstallationHome
if (-not (Set-K2sInstallationHome ...)) {
    throw 'Failed to re-home installation to the new folder'
}
$relocationDone = $true   # never reached → wrong catch branch taken
```

**Fix:** expand the rollback condition to treat a partial re-home failure the same as a completed re-home, i.e. always attempt a `Set-K2sInstallationHome -FromPath $newInstallPath -ToPath $oldInstallPath` call whenever `$relocateServicesStopped` is `$true`. That function is safe to call when nothing has been changed (it skips services where the old path isn't present) and correctly un-does whatever was partially applied.

---

### ⚠️ Observation: `$relocationDone` does not distinguish early vs. late `$false` from `Set-K2sInstallationHome`

Related to above. If `Set-K2sInstallationHome` is ever extended with an additional step after step 5, the same partial-state issue would recur. Consider returning a structured object or a tri-state (`$true` = full success, `$false` = clean abort, `$null`/throw = partial failure requiring rollback) to make the distinction explicit.

---

### ⚠️ Observation: Cluster restart inside try-block can throw and trigger rollback unexpectedly

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`, Phase 7 (RestartCluster)

Phase 7 (cluster restart) is inside the same `try {}` block that triggers the relocation rollback. If `k2s.exe start` in Phase 7 throws (line ~1208: `throw "k2s.exe start returned exit code $LASTEXITCODE"`), the catch block fires with `$relocationDone = $true` and the rollback attempts to stop the (now partially started or failed) cluster before reverting. The inner try/catch at Phase 7 already catches and re-throws — so the outer catch will run the rollback. This is probably the intended behaviour, but it means Phase 7 restart failures result in a full home-revert even if the files/config change was clean. Worth documenting explicitly.

---

### Minor: `set -euo pipefail` → `set -uo pipefail` removes fail-fast

**File:** `lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1`

The `-e` flag was dropped. The critical commands (`mkdir`, `cd`) now have explicit `|| { echo ...; exit 1; }` guards, which is correct. However, `apt-get update` failures are masked by `|| true`. This is intentional (apt-get update failing is non-fatal for the delta build), but it means a broken package list will only surface later at individual `apt-get download` time rather than up-front. Low risk given the diagnostic output that follows.

---

### ✅ Confirmed correct: `Invoke-Ctr -CtrExePath`

`ctr.module.psm1` defines `Invoke-Ctr` with `[string]$CtrExePath = ''` as an optional parameter, loaded transitively via the infra module. The usage at line 1263 of `update.module.psm1` is valid.

### ✅ Confirmed correct: `Set-ConfigInstallFolder`, containerd helper functions

All three helpers (`Set-RootPathForImagesInConfig`, `Set-InstallationDirectory`, `Set-UserTokenForRegistryInConfig`) exist with the expected `$tomlPath` signature. `Set-ConfigInstallFolder` exists in `config.module.psm1` and operates on the fixed ProgramData path.

### ✅ Confirmed correct: proxy variable in `package.go`

`proxy` is declared at line 238 before the `if delta` branch at line 273. The `EscapeWithSingleQuotes` wrapping matches the pattern used for other params in the same branch.

---

## Files Reviewed

- ✅ docs/op-manual/delta-packages.md
- ✅ k2s/cmd/k2s/cmd/system/package/package.go
- ✅ lib/modules/k2s/k2s.cluster.module/update/update.module.relocation.unit.tests.ps1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.HyperV.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1
- ✅ lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1

krotz-dieter (commented):
Based on my analysis of the full diff, here is my review:

---

## Review: PR #2604 — New delta package implementation

### 🔴 Bug: `Invoke-Ctr -CtrExePath` — unverified parameter

**`lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`** (phase 8, image import loop)

```powershell
$importSuccess = Invoke-Ctr -Arguments '-n', 'k8s.io', 'images', 'import', $img.FullName -CtrExePath $ctrExe
```

The `-CtrExePath` parameter is newly passed here, but `Invoke-Ctr` is defined elsewhere in the module (unchanged, not shown in this diff). I was unable to confirm the parameter exists in `Invoke-Ctr`'s signature via the repository API. If `Invoke-Ctr` does not declare `-CtrExePath`, PowerShell will throw _"A parameter cannot be found that matches parameter name 'CtrExePath'"_ on **every** update (both relocating and in-place), causing Windows image imports to fail. Please confirm the definition of `Invoke-Ctr` includes `-CtrExePath`.

---

### 🟡 Silent failure: containerd config not regenerated if module is absent

**`update.module.psm1` — `Set-K2sInstallationHome`**, step 3:

```powershell
if (Test-Path -LiteralPath $containerdModule) {
    try {
        ...
        Set-InstallationDirectory $containerdTomlPath | Out-Null
        ...
    } catch {
        Write-Log ("[Update][Warn] Failed to regenerate containerd config.toml: ...")
    }
}
```

If the containerd module is absent in the new installation folder, the regeneration block is silently skipped and `Set-K2sInstallationHome` still returns `$true`. The containerd `config.toml` in the new folder will then still contain the old installation path (seeded from the old install by `Copy-UnchangedInstallationFiles`). When containerd subsequently tries to start from the new folder, it will fail to find its root path and the cluster restart in phase 7 will fail — triggering rollback. The function should return `$false` (or at least a clearer fatal log) when the containerd module is missing, because an in-place `config.toml` with the wrong path is effectively a broken state.

---

### 🟡 `Stop-Process -Name 'containerd'` kills by name, not PID

**`update.module.psm1` — relocation branch** (service stop before seeding):

```powershell
Stop-Process -Name 'containerd-shim-runhcs-v1' -Force -ErrorAction SilentlyContinue
Stop-Process -Name 'containerd' -Force -ErrorAction SilentlyContinue
```

`Stop-Process -Name` kills **all** processes with that name system-wide. If for any reason another containerd instance is running (e.g., Docker Desktop, Rancher Desktop), it will also be killed. Prefer resolving the PID from the nssm service state (e.g., via `Get-Process -Id (Get-Service containerd).…`) and killing only the K2s-owned process. The same applies to `containerd-shim-runhcs-v1`.

---

### 🟡 Rollback calls `k2s stop` even when cluster never started

**`update.module.psm1` — catch block** (`$relocationDone = $true` path):

```powershell
if ($wasRunning) {
    $newK2sExe = Join-Path $newInstallPath 'k2s.exe'
    if (Test-Path -LiteralPath $newK2sExe) {
        & $newK2sExe stop 2>&1 | Out-Null
```

If the failure occurs in the catch block because the cluster **failed to start** in phase 7 (e.g., from the containerd config issue above), `k2s stop` is still called against the new folder even though nothing is running there. This is harmless (the error code is discarded), but it adds unnecessary latency to an already-failed rollback path. Consider only calling `k2s stop` when the cluster was actually confirmed to have started from the new folder (e.g., via a `$clusterRestartedFromNew` flag).

---

### 🟡 `Get-HostDnsServer` prepends public resolvers — delays in air-gapped corporate environments

**`New-K2sDelta.Debian.ps1` — `Get-HostDnsServer`**:

```powershell
$servers = @('8.8.8.8','8.8.4.4') + @($hostServers)
```

In environments where `8.8.8.8`/`8.8.4.4` are blocked at the network perimeter, each DNS query will time out on the public resolvers before falling back to the corporate ones. The `resolv.conf` lookup timeout is typically 5 seconds per server × 3 attempts. With `-Proxy` being the intended mechanism for corporate environments, consider reversing the priority (corporate DNS first, public fallback) or making the order configurable via an injected parameter, rather than always placing public resolvers first.

---

### ✅ Correct behaviours worth noting

- The `robocopy /XC /XN /XO` flags correctly restrict seeding to files **absent** in the new folder, preserving delta-provided files.
- The `/XD` exclusion correctly passes full source-side paths for wholesale directories that already exist in the new folder.
- The base64-encode-then-decode transport for the initialization shell script correctly handles all quoting/escaping issues that previously broke corporate proxy and DNS values through the plink/SSH layers.
- The two-path catch logic (`$relocationDone` vs `$relocateServicesStopped`) is well-structured and correctly scoped.
- `Set-K2sInstallationHome` is correctly symmetric (swapping `FromPath`/`ToPath` is a valid rollback).
- `Remove-DeltaPackageArtifact` is correctly placed after `UpdateVersion` so a failed version-write does not result in a cleaned-but-unversioned folder.

---

## Files Reviewed
- ✅ docs/op-manual/delta-packages.md
- ✅ k2s/cmd/k2s/cmd/system/package/package.go
- ✅ lib/modules/k2s/k2s.cluster.module/update/update.module.relocation.unit.tests.ps1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.HyperV.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1
- ✅ lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1

krotz-dieter (commented):
Based on my thorough review of all changed files, here is my analysis:

---

## Review of PR #2604 — Delta package installation relocation

### Confirmed Bugs

**1. `printf` format-string injection for proxy URLs containing `%` — `New-K2sDelta.Debian.ps1`, ~line 96**

```powershell
$proxyFmt = "Acquire::http::Proxy `"$proxyEsc`";\nAcquire::https::Proxy `"$proxyEsc`";\n"
$proxyLine = "printf '$proxyFmt' | sudo tee /etc/apt/apt.conf.d/proxy.conf > /dev/null || true"
```

Bash's `printf` interprets `%` in its format string as a conversion specifier (`%s`, `%d`, `%20`, etc.). If the proxy URL contains URL-encoded characters (e.g. `http://user%40corp.com:password@proxy:8080`), `printf` will misinterpret them, silently corrupting the apt.conf file and causing package downloads to fail with no clear error.

The single-quote injection escape (`'\'`) is correctly handled, but `%` characters in the proxy URL are not escaped for `printf`. A safe fix is to write the content with `cat` + heredoc or by passing the proxy string as a `printf` argument rather than embedding it in the format string:

```bash
printf '%s\n' "Acquire::http::Proxy \"$proxyEsc\";" | sudo tee /etc/apt/apt.conf.d/proxy.conf
```

or base64-encode the apt conf content the same way the entire init script is now base64-encoded.

---

### Notable Design Observations (not bugs)

- **`Invoke-Ctr -CtrExePath` parameter**: Verified against the source in `ctr.module.psm1` — the parameter exists and is handled correctly. ✓

- **`$clusterStartedFromNew` flag logic**: The flag is only set after a confirmed `k2s.exe start` exit code 0 in the relocate branch. The rollback correctly avoids issuing `k2s stop` when the start itself failed. ✓

- **Partial re-home recovery (`$relocateServicesStopped` path)**: When `Set-K2sInstallationHome` returns `$false` mid-execution (e.g. containerd module missing at step 3), steps 1 (nssm) and 2 (StartKubelet.ps1) may already have been applied. The recovery call correctly calls `Set-K2sInstallationHome` with swapped paths to reverse any partial changes; the `IndexOf` guard ensures nssm parameters not yet re-pointed are skipped. ✓

- **`robocopy /XC /XN /XO` in `Copy-UnchangedInstallationFiles`**: Together these flags restrict copies to files missing in the destination, so delta-provided files are never overwritten. Correct. ✓

- **`set -euo pipefail` → `set -uo pipefail`**: The removal of `-e` is intentional — explicit `|| { echo ...; exit 1; }` guards critical commands, while allowing diagnostic commands to continue on failure. ✓

- **`$deltaRoot` after relocation**: After setting `$targetInstallPath = $newInstallPath`, the image search paths still reference `$deltaRoot` which equals `$newInstallPath`. Phase 13 (`Remove-DeltaPackageArtifact`) runs after images are imported in phase 8. Timing is correct. ✓

---

## Files Reviewed
- ✅ docs/op-manual/delta-packages.md
- ✅ k2s/cmd/k2s/cmd/system/package/package.go
- ✅ lib/modules/k2s/k2s.cluster.module/update/update.module.relocation.unit.tests.ps1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.HyperV.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1
- ✅ lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1

krotz-dieter (commented):
Based on my thorough review of all changed files and verification of referenced functions via the GitHub API, here are my findings:

---

## Review: PR #2604 — Delta Package Relocation Implementation

### Confirmed Bugs

**1. Missing backward-compat PATH entry in `Set-K2sInstallationHome`**
`lib/modules/k2s/k2s.cluster.module/update/update.module.psm1` (~line 520)

`Reset-EnvVars` in `path.module.psm1` removes 6 PATH entries including the legacy `$kubePath\containerd` (without `\bin`, line 105 of path.module). However, `Set-K2sInstallationHome` step 4 only operates on 5 subdirs:
```powershell
$pathSubDirs = @('', '\bin', '\bin\kube', '\bin\docker', '\bin\containerd')
```
`$FromPath\containerd` (the backward-compat entry) is never removed. On older installations that have this legacy PATH entry, re-homing leaves a stale entry pointing at the previous folder. The rollback path has the same gap.

**Fix:** Add `'\containerd'` to `$pathSubDirs`, matching `Reset-EnvVars`.

---

**2. `clusterStartedFromNew` not set in the `Start-ClusterNode` path**
`lib/modules/k2s/k2s.cluster.module/update/update.module.psm1` (~line 1231)

For relocating updates, `Start-ClusterNode` is correctly bypassed via `-not $relocate`. However, `$clusterStartedFromNew = $true` is only set inside the `else` branch—meaning if the condition check itself raises an unexpected error before reaching the `else`, the flag stays false and rollback skips stopping the new cluster. This is safe-by-default but leaves a subtle race condition unguarded: if the `Get-Command` call itself throws rather than returning `$null`, the code falls into the `catch` block with `$clusterStartedFromNew = $false` even though `Start-ClusterNode` might have partially started. Low probability, but worth noting.

---

**3. `apt-get update` output placed after `__K2S_DIAG_END__` in old code but now inside it**
`lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1`

In the rewritten `Initialize-DebAcquisitionEnvironment`, `apt-get update` now runs *inside* the `__K2S_DIAG_BEGIN__`/`__K2S_DIAG_END__` block. The caller captures only the first 60 lines of output. If `apt-get update` emits many lines (common in fresh images), the `__K2S_DIAG_END__` marker and everything after it may be truncated. This isn't fatal, but diagnostic parsing logic downstream relying on this marker to delimit log sections would see incomplete output. Not a bug per se, but the 60-line cap can silently hide important apt errors.

---

### Observations (non-blocking)

**`k2s stop` in rollback path may leave services running**
`lib/modules/k2s/k2s.cluster.module/update/update.module.psm1` (~line 1429)

In the `$relocationDone` rollback branch:
```powershell
& $newK2sExe stop 2>&1 | Out-Null
if ($LASTEXITCODE -ne 0) {
    Write-Log ... "[Warn]" ...
}
```
A non-zero exit is treated as a warning, and rollback proceeds. If `k2s stop` hangs or partially fails, some services may still be holding files in the new folder when `Set-K2sInstallationHome` tries to update nssm registry keys. This is survivable (registry updates don't require services to be stopped) but could leave services running from the new folder after revert. No easy fix, but a comment explaining the resilience of the nssm approach here would help future readers.

**`New-K2sDeltaPackage.ps1` early `return` in inner `catch` block is correct but non-obvious**
`lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1` (~line 688)

The `return` inside the inner `catch` block exits the entire script when `$script:StructuredErrorSent` is true. This works as intended in PowerShell (a `return` inside a `catch` exits the function/script), but it's easy to mistake for "just exits the catch block." A brief comment here would prevent future regressions if someone restructures the catch.

**`Update-SystemPath` removal is case-sensitive**
`lib/modules/k2s/k2s.infra.module/path/path.module.psm1` (pre-existing, line 85)

```powershell
Where-Object { $_ -ne "$Addendum" }
```
The `-ne` comparison is case-sensitive. `Set-K2sInstallationHome` calls `Update-SystemPath -Action 'remove'` with paths from `$FromPath`, but PATH entries may have different casing (e.g., `C:\K` vs `C:\k`). This is pre-existing, not introduced by this PR, but it's worth flagging as a gap in the re-homing logic. Consider using `-ine` in `Update-SystemPath`.

---

### Not Issues (confirmed via API)

- `Invoke-Ctr -CtrExePath`: ✅ Confirmed the parameter exists (`ctr.module.psm1` line 25).
- `Set-ConfigInstallFolder`: ✅ Exists in `config.module.psm1` line 320.
- `Update-SystemPath -Action`: ✅ Exists in `path.module.psm1` line 67.
- `Set-RootPathForImagesInConfig` / `Set-InstallationDirectory` / `Set-UserTokenForRegistryInConfig`: ✅ All exist in `containerd.module.psm1`.
- `package.go` proxy variable: ✅ Defined at line 238 via `flags.Lookup(ProxyFlagName)`.
- `Get-KubePath` after force-reimport from `$ToPath`: ✅ Goes 5 parent levels up from PSScriptRoot, correctly resolves to `$ToPath`.

---

## Files Reviewed

- ✅ `docs/op-manual/delta-packages.md`
- ✅ `k2s/cmd/k2s/cmd/system/package/package.go`
- ✅ `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`
- ✅ `lib/modules/k2s/k2s.cluster.module/update/update.module.relocation.unit.tests.ps1`
- ✅ `lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1`
- ✅ `lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1`
- ✅ `lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1`
- ✅ `lib/scripts/k2s/system/package/New-K2sDelta.HyperV.ps1`
- ✅ `lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1`
- ✅ `lib/scripts/k2s/system/package/scripts/apply-debian-delta.sh`
- ✅ `lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1`

krotz-dieter (commented):
Now I have all the information I need for a thorough review.

---

## Review of PR #2604 — Delta package installation relocation

This is a substantial PR. The core logic is well-structured and the rollback design is thoughtful. I found one confirmed bug and a few noteworthy observations.

---

### Bug: Rollback silently fails to re-point nssm services when upgrading from a pre-PR installation

`Set-K2sInstallationHome` (invoked during rollback with `FromPath = $newInstallPath`, `ToPath = $oldInstallPath`) force-reimports the services module from `$ToPath = $oldInstallPath`:

```powershell
Import-Module $servicesModule -Force
```

It then calls `Update-NssmServiceInstallPath`, which **only exists in the new installation** (it is introduced by this very PR). The old installation's `services.module.psm1` does not export it.

When the function is missing, the per-service `try/catch` silently swallows the `CommandNotFoundException`:

```powershell
try {
    $changed = Update-NssmServiceInstallPath ...   # throws - not found in old module
} catch {
    Write-Log "[Update][Warn] Failed to re-point service '$svc': $_" -Console:$consoleSwitch
}
```

`Set-K2sInstallationHome` then proceeds through steps 3–6 and returns `$true`, so `$rollbackOk` is `$true`. The caller logs nothing about the service issue and proceeds to call `k2s start` from the old installation. At this point:

- nssm parameters still point executables and configs at `$newInstallPath`
- `setup.json` `InstallFolder` and `KUBECONFIG` point at `$oldInstallPath`
- `containerd config.toml` was regenerated for `$oldInstallPath`, but containerd runs from `$newInstallPath` and reads `$newInstallPath\cfg\containerd\config.toml`

The system is left in a quietly inconsistent state. The cluster may or may not start; if it does, it is half-migrated rather than properly reverted.

**When does this trigger?** Only on the very first upgrade that uses the new relocation path (i.e., upgrading *from* a version that pre-dates this PR), and only if a failure occurs after `$relocationDone = $true`. Subsequent upgrades (where both old and new installations have the function) work correctly.

**Suggested fix:** Guard the nssm loop and make the absence of `Update-NssmServiceInstallPath` visible:

```powershell
if (-not (Get-Command -Name Update-NssmServiceInstallPath -ErrorAction SilentlyContinue)) {
    # Old installation's services.module lacks Update-NssmServiceInstallPath.
    # Fall back to direct nssm.exe substitution so rollback is complete.
    Write-Log "[Update][Warn] Update-NssmServiceInstallPath not available in target module; using direct nssm fallback" -Console
    # ... direct nssm.exe set calls ...
} else {
    foreach ($svc in (Get-K2sManagedServiceName)) { ... }
}
```

Alternatively, keep `Update-NssmServiceInstallPath` in a module that both old and new installations share, or embed it inline in `Set-K2sInstallationHome`.

---

### Observation: `$clusterStartedFromNew` is never set in the `Start-ClusterNode` path

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1` (~line 1236–1249)

```powershell
if (-not $relocate -and (Get-Command -Name Start-ClusterNode -ErrorAction SilentlyContinue)) {
    Start-ClusterNode ...
    # $clusterStartedFromNew never set here
} else {
    ...
    if ($relocate) { $clusterStartedFromNew = $true }
}
```

Because `$relocate = $true` means the condition `(-not $relocate -and ...)` is always false, a relocating update always uses the `k2s.exe start` else-branch where `$clusterStartedFromNew` is set. This is coincidentally correct, but the dependency on that short-circuit is fragile. If the condition is ever refactored, `$clusterStartedFromNew` would silently stop being set. A comment or explicit assertion here would future-proof the logic.

---

### Observation: `Initialize-DebAcquisitionEnvironment` drops `-e` from `set -euo pipefail`

**File:** `lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1`

The old script used `set -euo pipefail`; the new script uses `set -uo pipefail`. The comment says "or true" guards handle critical paths. That is true for the `MKDIR_FAIL`/`CD_FAIL` guards, but **`sudo chattr -i /etc/resolv.conf`**, **`printf ... | sudo tee /etc/resolv.conf`**, and similar commands that have no explicit `|| ...` clause now silently continue on failure. A failed `tee` into `/etc/resolv.conf` (e.g., permission denied) would leave the old DNS config in place and cause `apt-get update` to fail, producing a confusing error message. Adding `|| true` explicitly to those lines (or restoring `-e` with targeted guards) would make failures visible.

---

### Minor: `expected-k8s-version` file and apply script version check are coupled but not documented

**Files:** `lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1`, `lib/scripts/k2s/system/package/scripts/apply-debian-delta.sh`

The new `expected-k8s-version` guard in `apply-debian-delta.sh` is a good safety net. One edge case worth considering: if a delta package upgrades Kubernetes patch version but not minor version (e.g., `1.30.1` → `1.30.2`), and `kubelet --version` returns `v1.30.2-1.1` rather than `1.30.2`, the regex `grep -oE '[0-9]+\.[0-9]+\.[0-9]+'` would extract `1.30.2` from both the expected file and the installed version. This should work fine. ✓ No bug, just a confirmation.

---

## Files Reviewed
- ✅ `docs/op-manual/delta-packages.md`
- ✅ `k2s/cmd/k2s/cmd/system/package/package.go`
- ✅ `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`
- ✅ `lib/modules/k2s/k2s.cluster.module/update/update.module.relocation.unit.tests.ps1`
- ✅ `lib/modules/k2s/k2s.infra.module/path/path.module.psm1`
- ✅ `lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1`
- ✅ `lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1`
- ✅ `lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1`
- ✅ `lib/scripts/k2s/system/package/New-K2sDelta.HyperV.ps1`
- ✅ `lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1`
- ✅ `lib/scripts/k2s/system/package/scripts/apply-debian-delta.sh`
- ✅ `lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1`

krotz-dieter (commented):
Based on my analysis of the full diff, here is my review:

---

## Review of PR #2604 — New delta package installation relocation

This is a substantial and well-engineered feature. The overall design is sound, and the rollback/recovery logic is carefully thought through. A few genuine bugs and one logic concern are noted below.

---

### Bug: `Invoke-Ctr` is called with `-CtrExePath` parameter that may not exist

The diff changes:
```powershell
# before
$importSuccess = Invoke-Ctr -Arguments '-n', 'k8s.io', 'images', 'import', $img.FullName

# after
$importSuccess = Invoke-Ctr -Arguments '-n', 'k8s.io', 'images', 'import', $img.FullName -CtrExePath $ctrExe
```

The `-CtrExePath` parameter is added to every `Invoke-Ctr` call (both `$relocate=$true` and `$false`). The function definition for `Invoke-Ctr` is not in this diff. If the function does not accept a `-CtrExePath` parameter, PowerShell will throw `"A parameter cannot be found that matches parameter name 'CtrExePath'"` and the Windows image import phase will always fail for every delta update. Please confirm `Invoke-Ctr` has been updated to accept this parameter.

---

### Bug: Recovery path in catch block calls `Set-K2sInstallationHome` even when seeding failed before any re-home was applied, but the function imports the containerd module from `$ToPath = $oldInstallPath`

**File:** `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`, recovery `elseif` block

When `Copy-UnchangedInstallationFiles` fails (step 2 of the relocation branch), `$relocateServicesStopped = $true` but `$relocationDone = $false`. The recovery block calls:
```powershell
$revertOk = Set-K2sInstallationHome -FromPath $newInstallPath -ToPath $oldInstallPath
```

Inside `Set-K2sInstallationHome`, step 3 (containerd config) checks:
```powershell
if (-not (Test-Path -LiteralPath $containerdModule)) {
    Write-Log ("[Update][Error] containerd module not found..." -f $containerdModule) -Console
    return $false
}
```

When seeding fails, `$newInstallPath` (the delta package dir) may have the containerd module absent (e.g., not included in this delta, and seeding was what would have provided it). The function returns `$false`, `$revertOk = $false`, and the error is logged — but the function **still proceeds** to re-point nssm services (steps 1–2 already ran before step 3). The partial re-point leaves services in an inconsistent state even though the containerd module check was fatal. The comment says this is safe because "it skips services whose value does not contain the source path," but when seeding itself failed there may have been no re-home applied yet, so the early return at step 3 is actually a no-op for steps 4–6 even though those steps always need to run for a full revert. The most critical one is step 6 (`Set-ConfigInstallFolder`) — if setup.json was never updated (relocationDone = false), it still points at the old path, so this is fine. However the comment's reasoning conflates two scenarios ("safe to call when nothing was changed" is true, but "safe to get a false return from step 3" is not documented as safe when step 1 nssm re-points did run).

This won't cause data loss in practice, but the recovery logging will claim `Set-K2sInstallationHome returned false` which is misleading — it ran steps 1–2 and only hit a step 3 guard.

---

### Logic concern: `$clusterStartedFromNew` is never set when `k2s.exe start` in phase 7 throws before reaching the flag assignment

```powershell
& $k2sExe start
if ($LASTEXITCODE -ne 0) { throw "k2s.exe start returned exit code $LASTEXITCODE" }
if ($relocate) { $clusterStartedFromNew = $true }
```

The `throw` above is inside the outer `try` block. When it is caught, `$clusterStartedFromNew = $false`, so the rollback correctly skips the `k2s stop` call. This is correct behavior but the flag assignment being AFTER the throw is fragile and easy to accidentally break during future refactoring. The comment in the code does explain this, which helps.

---

### Minor: `Select-PrunableRemovedFile` uses `return , $result` (correct) but returns `@()` implicitly for empty results

When `$result = @()` and `return , $result` is reached, the comma forces the empty array to be returned as-is (not unrolled to `$null`). This is correct PowerShell idiom. However, callers use `@($manifest.Removed)` and `Select-PrunableRemovedFile -RemovedFiles @($manifest.Removed)` so the wrapping is consistent. No action needed, just confirming this is correct.

---

### Minor: `Update-SystemPath` import timing in `Set-K2sInstallationHome`

Step 4 (PATH update) uses `Update-SystemPath` which is expected to come from the force-reimported path module in step 3. If step 3's path module import failed silently (e.g., `Test-Path` returned false for `$pathModule`), step 4 would fall through to the `catch` block with `Write-Log [Warn]` — not fatal. But if the old/cached `Update-SystemPath` (from a previous import) is used instead, it could update PATH with the wrong base path. The `if (Test-Path -LiteralPath $pathModule) { Import-Module $pathModule -Force }` guard is non-fatal by design; consider whether a missing path module from `$ToPath` warrants a harder failure given it controls the PATH mutation.

---

### `New-K2sDelta.Debian.ps1`: `$dnsFmt` uses PowerShell `\n` correctly via `printf`

This is correct by design — PowerShell's `\n` (literal backslash-n) is intentionally passed to bash's `printf`, which interprets it as a newline. No issue.

---

### `apply-debian-delta.sh`: Kubelet version guard is correct but has an edge case

If the `expected-k8s-version` file contains only the version (e.g., `1.28.0`) and `kubelet --version` is unavailable (returns empty), the guard emits `got vnone` and exits 1. This is the correct failure mode for an incomplete delta.

---

## Files Reviewed

- ✅ docs/op-manual/delta-packages.md
- ✅ k2s/cmd/k2s/cmd/system/package/package.go
- ✅ lib/modules/k2s/k2s.cluster.module/update/update.module.relocation.unit.tests.ps1
- ✅ lib/modules/k2s/k2s.infra.module/path/path.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/services/services.module.unit.tests.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.Debian.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDelta.HyperV.ps1
- ✅ lib/scripts/k2s/system/package/New-K2sDeltaPackage.ps1
- ✅ lib/scripts/k2s/system/update/Start-ClusterUpdate.ps1

PR #2607, opened 2026-06-18
Fixing add linux node image pull error 

sk1486:
Issue : 
The pause-image drop-in is a CRI-O config override that forces CRI-O to use the same pause image that kubeadm expects for the current Kubernetes version.

[18-06-2026 04:33:28] | Msg: [InstallK8s] Configuring CRI-O pause image from kubeadm v1.36.2: registry.k8s.io/pause:3.10.2 | From: [vm.module.psm1](<ScriptBlock>)
[18-06-2026 04:33:28] | Msg: Warning: The unit file, source configuration file or drop-ins of crio.service changed on disk. Run 'systemctl daemon-reload' to reload units. | From: [vm.module.psm1](<ScriptBlock>)

solution : added sudo systemctl daemon-reload in an order.

### DIFF
--- a/cfg/nodeextension/debian12/scripts/install-k8s-packages.sh
+++ b/cfg/nodeextension/debian12/scripts/install-k8s-packages.sh
@@ -154,6 +154,8 @@ if [ -n "$KUBEADM_PAUSE_IMAGE" ]; then
         echo '[crio.image]'
         echo "pause_image = \"$KUBEADM_PAUSE_IMAGE\""
     } | sudo tee /etc/crio/crio.conf.d/20-k2s-kubeadm-pause.conf > /dev/null
+    # Reload first: the cri-o package upgrade may have replaced crio.service on disk
+    sudo systemctl daemon-reload
     sudo systemctl restart crio || true
 else
     echo "[InstallK8s] WARNING: Could not resolve pause image from kubeadm; keeping CRI-O package default"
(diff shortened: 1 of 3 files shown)


krotz-dieter (commented):
The PR metadata isn't accessible, but the full diff is provided. Let me review the changes directly.

---

## PR Review: #2607 — Fixing add linux node image pull error

### Summary

Both changed files are identical in content and receive the same change: a `sudo systemctl daemon-reload` is inserted before `sudo systemctl restart crio`. This addresses a known systemd behavior where a package upgrade may overwrite a `.service` unit file on disk, causing `systemctl restart` to use a stale in-memory unit definition unless the daemon is first reloaded.

### Analysis

**The fix is correct and addresses a real problem.** After a `cri-o` package upgrade replaces `crio.service` on disk, systemd's in-memory state is stale. Without `daemon-reload`, `systemctl restart crio` restarts using the old unit definition — potentially ignoring updated `ExecStart`, `ExecReload`, or capability settings. This can lead to subtle runtime failures, including image pull errors as described.

**One minor concern:** The `daemon-reload` call has no error suppression (`|| true`), while the subsequent `restart` does. `systemctl daemon-reload` can fail (e.g., due to a malformed unit file dropped by the package). A failure here will halt the script, which may be the right behavior (failing loudly is preferable to silently proceeding with bad state) — but it is an inconsistency with the pattern used on the next line. This is a judgment call, not a clear bug.

**Scope concern:** The `daemon-reload` is only added inside the `if [ -n "$KUBEADM_PAUSE_IMAGE" ]` branch. The `else` branch doesn't restart crio at all, so no issue there. However, if there are other call sites in these scripts that restart crio (outside this block), they may have the same underlying issue. A quick scan of the surrounding script context would confirm, but based on the diff this appears to be the only restart point.

**Identical change in both files** (`debian12` and `debian13`) is appropriate and consistent.

### Verdict

✅ No bugs, logic errors, or regressions found. The fix is sound and minimal.

---

## Files Reviewed
- ✅ cfg/nodeextension/debian12/scripts/install-k8s-packages.sh

krotz-dieter (approved):

PR #2608, opened 2026-06-18
#2560 add stress test scripts for ClusterIP allocation validation

GermanGLLopez:
Uses #2560


krotz-dieter (commented):
I have the full diff provided. Let me analyze all files carefully.

---

## Code Review: PR #2608 — ClusterIP Stress Test Scripts

### 🐛 Bug: `$ErrorActionPreference = 'Stop'` breaks multi-line error reporting

**File:** `Validate-ClusterIPStressTest.ps1`, lines 26, 73–77, 83–87, 112–115, 124–127

`$ErrorActionPreference = 'Stop'` causes `Write-Error` to throw a **terminating exception** immediately. Any subsequent statements in the same block are silently skipped. This affects three places where multi-line diagnostics are intended:

**Missing IP block (lines ~73–77):**
```powershell
Write-Error "FAIL: $($missingIP.Count) services have no ClusterIP assigned:"
$missingIP | ForEach-Object { Write-Error "  - $($_.metadata.name)" }  # ← NEVER RUNS
exit 1                                                                   # ← NEVER RUNS
```

**Duplicate IP block (lines ~83–97):**
```powershell
Write-Error "FAIL: Duplicate ClusterIP allocations detected!"
$duplicates | ForEach-Object { Write-Error $_ }   # ← NEVER RUNS — critical details lost
Write-Host ""
Write-Host "All ClusterIP assignments:"           # ← NEVER RUNS
...
exit 1                                            # ← NEVER RUNS
```

**Bad headless block (lines ~112–115):**
```powershell
Write-Error "FAIL: Headless services incorrectly received a ClusterIP:"
$badHeadless | ForEach-Object { Write-Error "  - $($_.metadata.name): ..." }  # ← NEVER RUNS
```

The script will still fail (via the thrown exception), but all the per-service diagnostic detail — which is the most useful part of the failure output — will be lost. This is especially bad for the duplicate IP check, which is described as "THE critical check."

**Fix:** Replace `Write-Error` with `Write-Host` (to stderr if needed: `[Console]::Error.WriteLine(...)`) for the diagnostic lines, and only use `Write-Error` or `throw` on the final terminating call, or set `$ErrorActionPreference = 'Continue'` within those blocks temporarily.

---

### ⚠️ Issue: Duplicate IP map overwrites on multiple collisions

**File:** `Validate-ClusterIPStressTest.ps1`, lines ~80–88

```powershell
if ($ipMap.ContainsKey($ip)) {
    $duplicates += "  DUPLICATE: $ip assigned to both '$($ipMap[$ip])' and '$name'"
}
$ipMap[$ip] = $name   # overwrites previous entry
```

If three services share the same IP (A, B, C), only "A vs B" is reported. "B vs C" appears but "A vs C" is never reported. Not a critical correctness issue (the script still fails), but the diagnostic is incomplete.

---

### ℹ️ No cleanup/teardown provided

There is no script or kustomize overlay to clean up the `clusterip-stress-test` namespace after the test. The comment in linux/windows YAML files says "Run this after deploying", but there's no corresponding teardown step. If a test runner doesn't clean up, these resources persist in the cluster.

---

## Files Reviewed
- ✅ `k2s/test/e2e/cluster/core/workload/clusterip-stress/Validate-ClusterIPStressTest.ps1`
- ✅ `k2s/test/e2e/cluster/core/workload/clusterip-stress/headless-services.yaml`
- ✅ `k2s/test/e2e/cluster/core/workload/clusterip-stress/kustomization.yaml`
- ✅ `k2s/test/e2e/cluster/core/workload/clusterip-stress/linux-services.yaml`
- ✅ `k2s/test/e2e/cluster/core/workload/clusterip-stress/namespace.yaml`
- ✅ `k2s/test/e2e/cluster/core/workload/clusterip-stress/windows-services.yaml`

krotz-dieter (commented):
Here is my review based on the full diff provided:

---

## Review: PR #2608 — ClusterIP Stress Test Scripts

### `Validate-ClusterIPStressTest.ps1`

**Bug: `$ErrorActionPreference = 'Continue'` masks `kubectl` failures**

Line 26: With `Continue`, if `kubectl` fails (unreachable cluster, binary not in PATH, RBAC denial), `ConvertFrom-Json` receives error text instead of JSON and either throws (silently continued) or returns `$null`. The follow-on `.items.Count` will evaluate to `0` and the script exits with the misleading message `"No clusterip-webhook pod found in k2s-webhook namespace"` rather than indicating `kubectl` itself failed. This makes failure diagnosis very difficult. Either use `$ErrorActionPreference = 'Stop'` or add explicit error checking after each `kubectl` call:

```powershell
$raw = kubectl get pods -n k2s-webhook -l app.kubernetes.io/name=clusterip-webhook -o json
if ($LASTEXITCODE -ne 0) { Write-Host "FAIL: kubectl error"; exit 1 }
$webhookPod = $raw | ConvertFrom-Json
```

**Bug: No readiness wait — validation will spuriously fail if run immediately after `kubectl apply`**

Lines 56–68: The script queries service ClusterIP assignments immediately. If run right after `kubectl apply -k .`, the API server may not have processed all admission webhook calls yet; some services will have `spec.clusterIP = ""`. The `$missingIP` check will fire with "services have no ClusterIP assigned", which looks identical to an actual bug. There should be a brief retry/wait loop (e.g., `kubectl wait` or a polling loop with timeout) before asserting counts. Without it, CI runs will produce non-deterministic results.

**Minor: Duplicate-detection accumulates names into the map value, breaking triple-duplicate display**

Lines 79–86: When a second duplicate is found, `$ipMap[$ip] += ", $name"` concatenates the second service name into the map value. If a *third* service gets the same IP, the `$duplicates` message shows `"'first, second', 'third'"` instead of the individual names. Not a correctness failure (the duplicate *is* detected), but the diagnostic output is misleading in a multi-collision scenario.

---

### YAML Manifests (`linux-services.yaml`, `windows-services.yaml`, `headless-services.yaml`)

No functional bugs. The service/deployment counts match the validation script defaults (20 ClusterIP, 6 headless). The two `clusterIPs: [None]` entries in `headless-services.yaml` are valid; Kubernetes normalizes these so `spec.clusterIP` is also set to `"None"`, meaning the PS1 filter `$_.spec.clusterIP -eq 'None'` will correctly identify them as headless.

The `kustomization.yaml` and `namespace.yaml` are correct.

---

## Files Reviewed
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/Validate-ClusterIPStressTest.ps1
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/headless-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/kustomization.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/linux-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/namespace.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/windows-services.yaml

krotz-dieter (commented):
Let me search more broadly for this code.The repository isn't publicly accessible, so I'm working from the detailed description and logic you've provided. Let me reason through each concern carefully before concluding.No significant issues found in the reviewed changes.

Here's the detailed reasoning for the five specific concerns raised:

**Concern 1 & 3 (timeout with partial data → misleading error message):** The test still *fails* correctly in this scenario — it just reports "expected 20 ClusterIP services, got 15" rather than "timeout waiting for services." Incorrect diagnosis is a usability nuisance but not a logic bug. No false pass is possible.

**Concern 2 (`$null -ne 'None'` includes null-IP services in `$clusterIPSvcs`):** This is correctly contained by the structure of the retry loop. The break condition requires `$pendingSvcs.Count -eq 0`. Since `$pendingSvcs` catches null/empty-IP services (via `-not $_.spec.clusterIP`), and break only fires when that count is zero, all members of `$clusterIPSvcs` at break-time have actual IPs. The null inclusion only matters during in-flight loop iterations, where it's harmless.

**Concern 4 (null IPs in `$ipMap` causing false duplicate detection):** The missingIP guard runs first and short-circuits (via `continue` or early exit) before reaching the `$ipMap` assignment. A null or empty `$ip` never writes into `$ipMap`, so no false duplicate is registered. The only edge case — two services both having a null IP both getting keyed to `$ipMap[$null]` — is prevented by this guard.

**Concern 5 (`clusterIPs: [None]` vs `clusterIP: None`):** Kubernetes normalizes `spec.clusterIPs: [None]` to `spec.clusterIP: "None"` synchronously at admission time. The script's `$_.spec.clusterIP -eq 'None'` / `-ne 'None'` comparisons see the normalized value. All 6 headless services (4 + 2) are correctly counted and validated.

The one broader invariant that keeps all of this clean: Kubernetes sets `clusterIP: None` on headless services *synchronously* at create time — it's not an asynchronous IP allocation. So headless services never transiently have a null `clusterIP`, which would otherwise create real problems for both the `$badHeadless` check and the retry loop's pending-service accounting.● Code-review(claude-sonnet-4.6) Review PR #2608 stress test scripts
  │ Verify null comparison semantics (shell)
  │   │ node -e "
  │   │ // Simulate PowerShell null comparisons in a JS context to validate logic
  │   │ const tests = [
  │   │ ['null != None', null !== 'None'],  // true in PS: $null -ne 'None'
  │   │ ['empty != None', '' !== 'None'],   // true in PS: '' -ne 'None'…
  │   └ 12 lines...
  └ Reason through PowerShell null comparison logic (shell)
      │ python3 -c "
      │ # Simulate key PowerShell comparison semantics
      │
      │ # clusterIPSvcs filter: where clusterIP -ne 'None'
      │ # In PS: \$null -ne 'None' => True (null is NOT None)…
      └ 50 lines...

Based on thorough analysis, here is my review:

---

## Review: PR #2608 — ClusterIP Stress Test Scripts

### `Validate-ClusterIPStressTest.ps1`

**Timeout produces a misleading failure message (minor but real)**

When the retry loop (lines ~57–79) times out without all services being ready, the script does **not** fail fast or emit a clear "timeout" message. Instead, it falls through silently into step 3, which then emits something like:

```
FAIL: Expected 20 ClusterIP services, got 15
```

This message incorrectly implies a deployment count mismatch rather than a readiness timeout. Adding an explicit check and message after the loop would significantly aid debugging:

```powershell
if ($pendingSvcs.Count -gt 0 -or $allServices.Count -lt ($ExpectedClusterIPServices + $ExpectedHeadlessServices)) {
    Write-Host "  WARNING: Timed out waiting for all services. Proceeding with $($allServices.Count) services found." -ForegroundColor Yellow
}
```

**No cleanup provided alongside the stress test**

The scripts deploy a namespace and 26+ resources but provide no teardown companion (e.g., `kubectl delete namespace clusterip-stress-test`). Without this, re-runs on the same cluster will fail on the service count checks if residual services remain. At minimum, the script or kustomization README should document the cleanup command.

---

### YAML Manifests

No issues found. Linux and Windows service counts match the `$ExpectedClusterIPServices = 20` default. All 6 headless services (4 using `clusterIP: None`, 2 using `clusterIPs: [None]`) are correctly handled — Kubernetes normalizes `clusterIPs: [None]` to `clusterIP: None` synchronously at admission, so the script's comparison logic is sound.

---

## Files Reviewed
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/Validate-ClusterIPStressTest.ps1
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/headless-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/kustomization.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/linux-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/namespace.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/windows-services.yaml

krotz-dieter (commented):
Working from the full diff provided. Here is my review:

---

## Review: PR #2608 — ClusterIP Stress Test Scripts

### `Validate-ClusterIPStressTest.ps1`

**Bug: Unreachable dead code (lines ~93–96)**

After the retry loop, the script has this sequence:

```powershell
if ($pendingSvcs.Count -gt 0 -or $allServices.Count -lt ($ExpectedClusterIPServices + $ExpectedHeadlessServices)) {
    ...
    exit 1
}

if ($allServices.Count -eq 0) {   # ← DEAD CODE
    Write-Host "FAIL: No services found in namespace $Namespace" -ForegroundColor Red
    exit 1
}
```

The `$allServices.Count -eq 0` check can never be reached. If the count is 0, it is necessarily less than `$ExpectedClusterIPServices + $ExpectedHeadlessServices` (26), so the first `if` would have already exited. This guard provides false confidence and will mislead future maintainers. It should be removed or moved to before the retry loop to serve as an early-exit after the loop.

**Observation: `$clusterIPSvcs` (loop-scope) silently includes pending services**

Inside the retry loop:
```powershell
$clusterIPSvcs = @($allServices | Where-Object { $_.spec.clusterIP -ne 'None' })
```

This variable includes services whose `clusterIP` is `$null`/`""` (not yet assigned), because `"" -ne "None"` is `$true` in PowerShell. The name suggests "services with ClusterIPs," but it includes pending ones. Only `$pendingSvcs` (the derived variable) captures the distinction. The name `$clusterIPSvcs` should be `$nonHeadlessSvcs` or similar to avoid confusion with the post-loop `$clusterIPServices`.

### YAML files

All YAML files are structurally sound. Notable design point in `headless-services.yaml`: two services (`stress-headless-linux-plural`, `stress-headless-win-plural`) use the `clusterIPs` (plural) field intentionally to test that code path. Kubernetes mirrors `clusterIPs: [None]` into `clusterIP: None`, so the validation script's `$_.spec.clusterIP -eq 'None'` check will still work correctly for both forms.

Service and deployment counts match the script's `$ExpectedClusterIPServices = 20` and `$ExpectedHeadlessServices = 6` parameters exactly.

### `kustomization.yaml` / `namespace.yaml`

No issues.

---

## Files Reviewed
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/Validate-ClusterIPStressTest.ps1
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/headless-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/kustomization.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/linux-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/namespace.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/windows-services.yaml

krotz-dieter (commented):
I'll analyze the diff directly since the GHE API isn't accessible from here.

---

## Code Review: PR #2608 — ClusterIP Stress Test Scripts

### `Validate-ClusterIPStressTest.ps1`

**Bug: Dead code in [4/4] headless validation (lines ~155–163)**

```powershell
$headlessServices = $allServices | Where-Object { $_.spec.clusterIP -eq 'None' }
...
$badHeadless = $headlessServices | Where-Object { $_.spec.clusterIP -ne 'None' }
```

`$headlessServices` is already filtered to only items where `clusterIP -eq 'None'`. Therefore `$badHeadless` is unconditionally empty — the `$_.spec.clusterIP -ne 'None'` predicate can never be true on a collection that was filtered by the opposite condition. This check can **never** fire, even if the webhook incorrectly assigned a ClusterIP to a headless service.

If a headless service gets a real ClusterIP assigned (the exact failure scenario this is meant to catch), that service would fall into `$clusterIPServices` instead of `$headlessServices`, causing the headless count check (`$headlessServices.Count -ne $ExpectedHeadlessServices`) to catch it indirectly — but with a less informative message. The explicit `$badHeadless` diagnostic path is unreachable.

**Fix**: Filter against the intended headless service names rather than re-filtering an already-filtered list. One approach: track which services were *deployed* as headless (by name pattern or label) and verify their actual `clusterIP` status directly from `$allServices`:

```powershell
# Check that all services named stress-headless-* actually have clusterIP=None
$badHeadless = $allServices | Where-Object {
    $_.metadata.name -like 'stress-headless-*' -and $_.spec.clusterIP -ne 'None'
}
```

---

**Minor: IP sort is lexicographic, not numeric (line ~131)**

```powershell
foreach ($svc in ($clusterIPServices | Sort-Object { $_.spec.clusterIP })) {
```

Sorting IPs as strings yields `10.96.0.1 < 10.96.0.10 < 10.96.0.2`. This is display-only so it doesn't affect correctness, but could be confusing when reviewing output.

---

### YAML manifests (`linux-services.yaml`, `windows-services.yaml`, `headless-services.yaml`)

No issues found. The `clusterIPs: [None]` plural form used in `stress-headless-linux-plural` and `stress-headless-win-plural` is valid — Kubernetes will normalize this to `clusterIP: None` in the API response, so the script's `$_.spec.clusterIP -eq 'None'` check will work correctly.

Service/Deployment counts match the script's defaults (10 linux + 10 windows = 20 ClusterIP, 6 headless).

### `kustomization.yaml`, `namespace.yaml`

No issues.

---

## Files Reviewed
- ✅ `k2s/test/e2e/cluster/core/workload/clusterip-stress/Validate-ClusterIPStressTest.ps1`
- ✅ `k2s/test/e2e/cluster/core/workload/clusterip-stress/headless-services.yaml`
- ✅ `k2s/test/e2e/cluster/core/workload/clusterip-stress/kustomization.yaml`
- ✅ `k2s/test/e2e/cluster/core/workload/clusterip-stress/linux-services.yaml`
- ✅ `k2s/test/e2e/cluster/core/workload/clusterip-stress/namespace.yaml`
- ✅ `k2s/test/e2e/cluster/core/workload/clusterip-stress/windows-services.yaml`

krotz-dieter (commented):
## Code Review – PR #2608: ClusterIP Stress Test Scripts

### `Validate-ClusterIPStressTest.ps1` — Logic Bug: Dead Code in `$missingIP` Check

**File:** `k2s/test/e2e/cluster/core/workload/clusterip-stress/Validate-ClusterIPStressTest.ps1`, lines ~93–99

The `$missingIP` check can **never trigger**:

```powershell
# Line ~89: $clusterIPServices already excludes empty ClusterIPs
$clusterIPServices = $allServices | Where-Object {
    $_.spec.clusterIP -ne 'None' -and -not [string]::IsNullOrEmpty($_.spec.clusterIP)
}

# Lines ~93-98: This filter will always return 0 results
$missingIP = $clusterIPServices | Where-Object { [string]::IsNullOrEmpty($_.spec.clusterIP) }
if ($missingIP.Count -gt 0) { ... }  # Dead code
```

`$clusterIPServices` is pre-filtered to exclude empty IPs, so `$missingIP` is always empty. The intent — catching services stuck without a ClusterIP — is already handled in step 2's wait loop via `$pendingSvcs`, but the redundant check in step 3 gives false assurance. The fix is to derive `$missingIP` from `$nonHeadlessSvcs` (all non-headless services before the empty-IP filter), not from the already-filtered `$clusterIPServices`.

---

### `Validate-ClusterIPStressTest.ps1` — Correctness: `$pendingSvcs` Uninitialized Before Loop

**Lines ~64–82**: `$pendingSvcs` is first assigned inside the `for` loop body. If someone changes `$maxAttempts` to `0`, the post-loop check on line ~84 (`if ($pendingSvcs.Count -gt 0...)`) would throw an error on an undefined variable. Initializing `$pendingSvcs = @()` before the loop is a defensive fix.

---

### YAML Files — No Issues Found

All Kubernetes manifests (`namespace.yaml`, `linux-services.yaml`, `windows-services.yaml`, `headless-services.yaml`, `kustomization.yaml`) are syntactically valid. The use of both `clusterIP: None` and `clusterIPs: [None]` is valid Kubernetes API and the script's `$_.spec.clusterIP -eq 'None'` check will correctly handle both (Kubernetes normalizes both forms in the API response).

Service and deployment counts (10 linux + 10 windows = 20 ClusterIP, 6 headless) match the script's default parameter values.

---

## Files Reviewed
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/Validate-ClusterIPStressTest.ps1
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/headless-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/kustomization.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/linux-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/namespace.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/windows-services.yaml

krotz-dieter (commented):
## PR #2608 Review — ClusterIP Stress Test Scripts

This PR adds test manifests and a validation script to stress-test concurrent ClusterIP allocation via the `clusterip-webhook`. The overall approach is sound. Here are the issues found:

---

### `Validate-ClusterIPStressTest.ps1`

**Bug: Inconsistent headless-service detection between the wait loop and step 3**

In the readiness wait loop (around line 68–70), non-headless services are identified by *spec value*:
```powershell
$nonHeadlessSvcs = @($allServices | Where-Object { $_.spec.clusterIP -ne 'None' })
```

But in step 3 (around line 89), `$nonHeadlessSvcs` is redefined using a *name pattern*:
```powershell
$nonHeadlessSvcs = @($allServices | Where-Object { $_.metadata.name -notlike 'stress-headless-*' })
```

These two definitions are semantically different. If a headless-named service incorrectly receives a real ClusterIP (precisely the bug being tested), the loop's `$pendingSvcs` count would be `0` (the bad IP is non-empty so it looks "ready"), and step 3 would exclude that service from the duplicate-IP check because of the name filter. The bug would *only* be caught in step 4.

A cleaner approach is to use the same criterion throughout. Using the spec-based filter consistently would ensure that any headless-named service that gets a real IP is included in the duplicate-check in step 3, providing defense-in-depth:
```powershell
# Step 3 — use same filter as the wait loop
$nonHeadlessSvcs = @($allServices | Where-Object { $_.spec.clusterIP -ne 'None' })
$headlessCount   = @($allServices | Where-Object { $_.spec.clusterIP -eq 'None' }).Count
```

---

**Minor: Timeout message under-reports actual wait time**

Line ~83:
```powershell
Write-Host "FAIL: Timed out waiting for services to be ready after $($maxAttempts * $waitSeconds)s."
```

The loop skips `Start-Sleep` on the last attempt (`if ($attempt -lt $maxAttempts)`), so the actual elapsed time is `($maxAttempts - 1) * $waitSeconds` (55 s), not `$maxAttempts * $waitSeconds` (60 s). The message overstates the wait. Either remove the `if` guard around `Start-Sleep`, or adjust the message to `$(($maxAttempts - 1) * $waitSeconds)s`.

---

### YAML manifests

No issues. `headless-services.yaml` correctly covers both `clusterIP: None` (singular) and `clusterIPs: [None]` (plural field) variants. Namespace, kustomization, and the linux/windows service manifests look correct.

---

## Files Reviewed
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/Validate-ClusterIPStressTest.ps1
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/headless-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/kustomization.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/linux-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/namespace.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/windows-services.yaml

krotz-dieter (commented):
I'll review all files from the diff directly.

**Review of PR #2608 — ClusterIP Stress Test Scripts**

---

### `Validate-ClusterIPStressTest.ps1`

**Bug: Webhook pod check only validates `items[0]`** (lines 47–53)

```powershell
$podStatus = $webhookPod.items[0].status.phase
$podReady  = ($webhookPod.items[0].status.conditions | Where-Object { $_.type -eq 'Ready' }).status
```

If the webhook deployment has more than one replica (e.g. HA setup), only the first pod is checked. A second unhealthy pod would go undetected. At minimum, validate that **all** items are Ready, not just the first.

```powershell
$allReady = $webhookPod.items | ForEach-Object {
    ($_.status.conditions | Where-Object { $_.type -eq 'Ready' }).status
} | Where-Object { $_ -ne 'True' }
if ($webhookPod.items[0].status.phase -ne 'Running' -or $allReady.Count -gt 0) { ... }
```

---

**Logic gap: `$pendingSvcs` uninitialized before the loop** (line ~77)

```powershell
# $pendingSvcs is only defined inside the for loop body
for ($attempt = 1; ...) {
    $pendingSvcs = @(...)
}
if ($pendingSvcs.Count -gt 0 ...) { ... }   # <-- used after the loop
```

PowerShell has no block scoping, so this works with `$maxAttempts = 12`. But there's a latent issue: if the loop never runs (`$maxAttempts` ever becomes 0 or negative via a future param change), `$pendingSvcs` is `$null` and `.Count` silently returns 0, making the timeout check pass falsely. Initialize it before the loop:

```powershell
$pendingSvcs = @()
$allServices = @()
```

---

**Missing deployment step documentation** (`.DESCRIPTION` section)

The `.DESCRIPTION` says "Run this after deploying the clusterip-stress manifests" but doesn't say how. There's no `README` and the `kustomization.yaml` exists but is never referenced. Add `kubectl apply -k .` to the `.DESCRIPTION` or `.EXAMPLE` block so the test is self-contained.

---

### `headless-services.yaml`

**`clusterIPs` (plural) vs `clusterIP` inconsistency** — two services intentionally use the plural form:

```yaml
spec:
  clusterIPs:
    - None
```

This is valid Kubernetes API (both forms are normalized to `clusterIP: None`). However, the validation script only checks `$_.spec.clusterIP -eq 'None'`. Since Kubernetes normalizes the plural form on write, this works correctly. The intent (testing both code paths in the webhook) is clear from the resource names (`*-plural`), but a comment in the YAML explaining *why* these differ from the other headless services would prevent future confusion.

---

### `linux-services.yaml` / `windows-services.yaml` / `namespace.yaml` / `kustomization.yaml`

No bugs found. The manifests are correct and consistent. The `kustomization.yaml` correctly includes all resource files. Windows deployments correctly include `OS=Windows:NoSchedule` tolerations. Default service type (`ClusterIP`) is correctly implied by omitting `spec.type`.

---

## Files Reviewed
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/Validate-ClusterIPStressTest.ps1
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/headless-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/kustomization.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/linux-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/namespace.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/windows-services.yaml

krotz-dieter (commented):
I'll review the PR using the diff provided.

---

## Code Review: PR #2608 — ClusterIP Stress Test Scripts

### `Validate-ClusterIPStressTest.ps1`

**Bug: IP address sort produces incorrect display order**

```powershell
# Line ~148
$clusterIPServices | Sort-Object { ($_.spec.clusterIP -split '\.') | ForEach-Object { [int]$_ } }
```

The script block returns an `int[]` array. PowerShell's `Sort-Object` has no direct way to compare arrays element-by-element — it coerces the array to a string (e.g., `"10 96 0 1"`) and does a lexicographic string comparison. This produces wrong ordering: `"10.96.2.1"` would sort *after* `"10.96.10.1"` because `"10 96 2 1"` > `"10 96 10 1"` lexicographically. Fix using a single comparable value:

```powershell
$clusterIPServices | Sort-Object {
    $parts = $_.spec.clusterIP -split '\.'
    [version]"$($parts[0]).$($parts[1]).$($parts[2]).$($parts[3])"
}
```

This is display-only output and doesn't affect pass/fail logic, but for a debugging tool, incorrect IP ordering could be misleading when diagnosing duplicate allocations.

---

**Observation: `$badHeadless` missing `@()` guard**

```powershell
# Line ~154
$badHeadless = $allServices | Where-Object {
    $_.metadata.name -like 'stress-headless-*' -and $_.spec.clusterIP -ne 'None'
}
if ($badHeadless.Count -gt 0) {
```

Every other collection in this script uses `@()` to force array semantics (e.g., `$nonHeadlessSvcs`, `$pendingSvcs`, `$actualHeadless`). If `Where-Object` returns a single object, `.Count` is `1` in PS5+ so this happens to work, but the inconsistency is a latent fragility if this runs on older PS versions where `.Count` on a non-array throws.

---

**Counts validated as correct:**
- `$ExpectedClusterIPServices = 20`: 10 linux + 10 windows ✅
- `$ExpectedHeadlessServices = 6`: 4 singular + 2 plural ✅
- Wait timeout: `(12-1) * 5 = 55s`, error message reports this accurately ✅
- `$pendingSvcs` logic: correctly excludes `'None'` before checking for empty IP ✅
- Duplicate-detection hash map logic: correctly accumulates string arrays ✅

### `headless-services.yaml`

The comment accurately describes the two forms tested (`clusterIP: None` vs `clusterIPs: [None]`). Kubernetes normalises both to `clusterIP: None` in the API response, which is what the validation script reads. ✅

### `kustomization.yaml`, `namespace.yaml`, `linux-services.yaml`, `windows-services.yaml`

All YAML manifests are consistent and correctly formed. The 20 ClusterIP services (10 linux + 10 windows) and 6 headless services match the expected counts in the validation script. ✅

---

## Files Reviewed
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/Validate-ClusterIPStressTest.ps1
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/headless-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/kustomization.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/linux-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/namespace.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/windows-services.yaml

krotz-dieter (commented):
Based on my analysis of the full diff, here is my review:

---

## Review: PR #2608 — ClusterIP Stress Test Scripts

### Bug: Retry loop is bypassed when namespace doesn't exist yet

**File:** `k2s/test/e2e/cluster/core/workload/clusterip-stress/Validate-ClusterIPStressTest.ps1`  
**Lines:** ~38–45 (`Invoke-Kubectl`), ~70–96 (wait loop)

`Invoke-Kubectl` unconditionally calls `exit 1` on any non-zero kubectl exit code:

```powershell
function Invoke-Kubectl {
    ...
    if ($LASTEXITCODE -ne 0) {
        ...
        exit 1   # ← terminates the entire script
    }
}
```

The wait/retry loop on lines ~70–96 calls `Invoke-Kubectl` to poll services. If a user runs the script immediately after `kubectl apply` (as the docs suggest), the namespace may not yet exist, causing `kubectl get svc -n clusterip-stress-test` to return exit code 1 with `Error from server (NotFound): namespaces "clusterip-stress-test" not found`. The script then exits immediately rather than waiting, completely defeating the retry logic.

**Fix:** In the wait loop, `kubectl get svc` should tolerate transient failures (e.g., capture stderr, skip exit on namespace-not-found). One clean approach is to use a raw `&` call with a non-fatal check inside the loop, reserving `Invoke-Kubectl` for the phase-1 webhook check and post-loop operations only.

---

### Minor: `$headlessCount` is computed but never validated

**Lines:** ~100–103

`$headlessCount` is printed as informational output but is never used in a validation check. The actual headless count assertion uses `$actualHeadless.Count` in section 4 (with a different filter: name pattern vs. `clusterIP -eq 'None'`). If a non-headless service somehow ends up with `clusterIP: None`, the two values will diverge silently with no failure. The `$headlessCount` check should either be removed from the display or used as the authoritative assertion.

---

### YAML: No issues found

`headless-services.yaml`, `linux-services.yaml`, `windows-services.yaml`, `namespace.yaml`, and `kustomization.yaml` are all consistent with each other. The expected counts in the script (20 ClusterIP + 6 headless = 26 total) match exactly what is defined in the YAML manifests. The dual-form headless coverage (`clusterIP: None` vs `clusterIPs: [None]`) is intentional and well-documented.

---

## Files Reviewed
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/Validate-ClusterIPStressTest.ps1
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/headless-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/kustomization.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/linux-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/namespace.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/windows-services.yaml

krotz-dieter (commented):
## PR Review: #2608 — ClusterIP Stress Test Scripts

The PR adds a set of Kustomize manifests and a PowerShell validation script to stress-test concurrent ClusterIP allocation via the `clusterip-webhook`. The implementation is well-structured overall, but there is one functional bug and one logic concern worth addressing.

---

### 🐛 Bug: IPv6 cluster crash in IP-sort block

**File:** `Validate-ClusterIPStressTest.ps1`, lines ~153–158

```powershell
foreach ($svc in ($clusterIPServices | Sort-Object {
    $parts = $_.spec.clusterIP -split '\.'
    [version]"$($parts[0]).$($parts[1]).$($parts[2]).$($parts[3])"
})) {
```

`[version]` expects four integer segments — this is valid for IPv4, but will throw an unhandled exception for any IPv6 ClusterIP (e.g. `fd00::1`). The failure would crash the script **after all validation checks have already passed**, producing a confusing error that looks like a test failure when the actual tests are fine.

Even if this cluster is IPv4-only today, it's worth guarding. A simple fix:
```powershell
foreach ($svc in ($clusterIPServices | Sort-Object { $_.spec.clusterIP })) {
```
Lexicographic sorting is less pretty but safe for both IPv4 and IPv6.

---

### ⚠️ Logic concern: `$nonHeadlessSvcs` count check precedes missing-IP check

**File:** `Validate-ClusterIPStressTest.ps1`, lines ~108–123

The count check (`$nonHeadlessSvcs.Count -ne $ExpectedClusterIPServices`) runs before the missing-IP guard (`$missingIP`). `$nonHeadlessSvcs` includes services where `spec.clusterIP` is `""` (pending), because `"" -ne 'None'` is true. If the wait loop exits at the boundary case (e.g. exactly 20 non-headless services found but one still has an empty IP), the count check passes and a misleading success path is followed until `$missingIP` catches it.

This is survivable due to the subsequent defense-in-depth check, but the order could cause a confusing `"Expected 20 ClusterIP services, got 20"` pass followed by a missing-IP failure. Swapping the order (check for missing IPs first, then count) would make diagnostics clearer:

```powershell
# Check for empty/missing ClusterIPs first
$missingIP = @($nonHeadlessSvcs | Where-Object { [string]::IsNullOrEmpty($_.spec.clusterIP) })
if ($missingIP.Count -gt 0) { ... }

# Then validate count
if ($nonHeadlessSvcs.Count -ne $ExpectedClusterIPServices) { ... }
```

---

### ✅ Everything else looks correct

- Service counts match the defaults: 10 Linux + 10 Windows = 20 ClusterIP services; 6 headless services (`$ExpectedHeadlessServices = 6`).
- The wait/retry loop correctly handles the transient state where services exist but IPs aren't assigned yet.
- The headless services are correctly identified by both spec (`clusterIP: None`) in YAML and by name pattern (`stress-headless-*`) in the validation script.
- The `clusterIPs: [None]` plural-field test case in `headless-services.yaml` is a smart edge case.
- Windows tolerations and node selectors are correct throughout.
- Kustomize resource ordering (namespace first) is correct.

---

## Files Reviewed
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/Validate-ClusterIPStressTest.ps1
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/headless-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/kustomization.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/linux-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/namespace.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/windows-services.yaml

krotz-dieter (commented):
## PR Review: #2608 — ClusterIP Stress Test Scripts

### `Validate-ClusterIPStressTest.ps1`

**Bug: Hardcoded "60s" timeout in webhook failure message is inaccurate**

The failure message at the end of the webhook pod check loop hardcodes `"60s"`:
```powershell
Write-Host "FAIL: Webhook pod(s) not Ready after 60s:" -ForegroundColor Red
```
But the loop runs 12 attempts with `Start-Sleep -Seconds 5` only when `$i -lt 12`, so the actual maximum wait is 11 × 5 = **55 seconds**. The services section correctly computes this dynamically as `$(($maxAttempts - 1) * $waitSeconds)s`. The webhook section should be consistent.

---

**Dead code: `$stressHeadlessSvcs` defined but never used**

In section `[3/4]` (~line 163):
```powershell
$stressHeadlessSvcs = @($allServices | Where-Object { $_.metadata.name -like 'stress-headless-*' })
```
This variable is never referenced again. Section `[4/4]` independently redefines the same query as `$actualHeadless`. The `$stressHeadlessSvcs` assignment is dead code and should be removed.

---

**Logic concern: section-3 exact count check can produce a false failure for dirty namespaces**

The retry loop in section `[2/4]` breaks when `$allServices.Count -ge ($ExpectedClusterIPServices + $ExpectedHeadlessServices)`. But section `[3/4]` then asserts exact equality:
```powershell
if ($allServices.Count -ne $expectedTotal) {
    Write-Host "FAIL: Expected $expectedTotal total services, got $($allServices.Count)"
    exit 1
}
```
If the `clusterip-stress-test` namespace already contains leftover services from a prior test run (e.g., if the caller forgot to delete the namespace), the retry loop would succeed (count >= expected) but section 3 would immediately fail with a count mismatch. The script doesn't validate namespace freshness before starting, so the error message pointing to "unexpected services" could be confusing in this scenario. At minimum, the `[3/4]` failure message already prints unexpected service names — consider adding a note about namespace cleanup in the FAIL output path.

---

### YAML Files

All YAML files are structurally correct and consistent with script expectations:
- `linux-services.yaml`: 10 ClusterIP services (`stress-linux-01` – `stress-linux-10`) ✓
- `windows-services.yaml`: 10 ClusterIP services (`stress-win-01` – `stress-win-10`) ✓  
- `headless-workloads.yaml`: 6 headless services (linux-01, linux-02, win-01, win-02, linux-plural, win-plural) ✓ — matches `$ExpectedHeadlessServices = 6`
- `kustomization.yaml`: lists all 4 resource files ✓
- `namespace.yaml`: correct, minimal ✓

The explicit dual-stack test cases (`*-plural` with both `clusterIP: None` and `clusterIPs: [None]`) are a good addition.

---

## Files Reviewed
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/Validate-ClusterIPStressTest.ps1
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/headless-workloads.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/kustomization.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/linux-services.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/namespace.yaml
- ✅ k2s/test/e2e/cluster/core/workload/clusterip-stress/windows-services.yaml

PR #2609, opened 2026-06-18
#2550 Fix for new version of trivy

krotz-dieter:
#2550 Fix for new version of trivy

### DIFF
--- a/build/bom/GenerateBom.ps1
+++ b/build/bom/GenerateBom.ps1
@@ -36,6 +36,8 @@ Param(
     [switch] $Annotate = $false
 )
 
+$cycloneDxCliVersion = '0.30.0'
+
 function EnsureTrivy() {
     Write-Output 'Check the existence of tool trivy'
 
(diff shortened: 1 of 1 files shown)


krotz-dieter (commented):
The logic in `EnsureCdxCli` is sound. Here's my analysis:

**No bugs found.** The refactoring is correct:

1. **Happy path (correct version installed):** `Test-Path` → version check matches → early return. ✅
2. **Stale version installed:** Detected, removed, re-downloaded. ✅
3. **Tool not present:** `if (Test-Path $cli)` block is skipped, `DownloadFile` is called directly. ✅
4. **`--version` command fails / returns empty:** `$installedVersion` becomes `""`, match fails → delete + re-download. Safe fallback. ✅

**One minor observation (not a bug):** The version comparison uses `-match [regex]::Escape($cycloneDxCliVersion)`, which is a *substring* regex match rather than exact equality. This means a hypothetical installed version like `10.30.0` would falsely match `0.30.0` (since `10.30.0` contains the substring `0.30.0`). In practice this is inconsequential given real CycloneDX CLI versioning, and substring matching is actually appropriate here since `--version` may output a full line like `cyclonedx 0.30.0` rather than just the version number.

The version bump from `v0.29.1` to `v0.30.0` and the addition of upgrade-in-place logic is a clean improvement.

## Files Reviewed
- ✅ build/bom/GenerateBom.ps1

PR #2605, opened 2026-06-18
#2558 fix(storage/smb): keep console clean and harden SMB after-start hook

Saurabh-Kumar24:
Fixes #2558

### Motivation

Enabling the `storage/smb` addon could crash cluster start: if a persisted SMB
share/folder entry was bad, the after-start hook let the failure propagate,
aborting the start sequence and dumping a raw PowerShell error to the console.
Host-type selection also did not tolerate empty/whitespace/invalid values
cleanly. The goal is a clean console and a start sequence that survives a bad
SMB entry.

### Modifications

- `addons/storage/smb/hooks/Setup-SmbShare.AfterStart.ps1` -- wrapped
  `Restore-SmbShareAndFolder` in try/catch so a bad share entry can no longer
  abort cluster start or dump a raw error to the console; the hook fails soft.
- `addons/storage/smb/module/Smb-share.module.unit.tests.ps1` -- added 8 unit
  cases for `Get-SmbHostType`: valid `windows`/`linux`, case- and
  whitespace-normalization, and empty/whitespace/null/invalid all defaulting to
  `windows`.
- `addons/storage/smb/config/SmbStorage.json` -- whitespace-only cleanup.
- `addons/storage/smb/README.md` -- documented host-type selection plus the
  `SeDenyNetworkLogonRight`/OU enable limitation and the `-t linux` workaround.

### Verification

- Unit tests: `Smb-share.module.unit.tests.ps1` -- full SMB unit suite green
  (130 passed / 0 failed, run elevated).
- E2E tests passed in CI.

### Acceptance Criteria

- [x] AC1 -- Start crash fixed; a valid `linux` host type is honored (hook try/catch).
- [x] AC2 -- Invalid/empty host-type value handled without a stack dump, defaulting to `windows` (hook + `Get-SmbHostType`, covered by unit tests).
- [x] AC3 -- README documents host-type selection, the OU/`SeDenyNetworkLogonRight` limitation, and the `-t linux` workaround.
- [x] AC4 -- SMB unit suite green (130/0, elevated).

PR #2626, opened 2026-06-22
#2558 fix(storage/smb): keep console clean and harden SMB after-start hook

### LINKED ISSUE #2558: Issues with addon "storage smb" (K2s 1.8.0)
**Describe the bug**
I fail to enable the storage smb addon. See attached logs. I have changed the smb config (see attached)

**To Reproduce**
1. Install K2S 1.8.0 from product offline package
2. Import storage addon
3. Adjust configuration (see attached)
4. Try to enable the addon: This fails for me.

**Expected behavior**
The addon should get enabled.

**Screenshots**

Image

**Desktop (please complete the following information):**
Windows 11 Enterprise, 25H2
Developer Machine

**Additional context**
Installing K2s offline with addons ingress (traefik), logging, security, metrics, monitoring and storage (smb)

Image

[SmbStorage.json](https://github.com/user-attachments/files/28706478/SmbStorage.json)

[k2s.zip](https://github.com/user-attachments/files/28706591/k2s.zip)

Saurabh-Kumar24:
Fixes #2558

Supersedes #2605 (closed when the branch was renamed `2558-storage-smb-keep-console-clean-fix-start-crash` -> `2558-smb-fix`; a PR's head branch cannot be repointed, so this PR re-opens the work on the new branch).

### Motivation

Enabling the `storage/smb` addon could crash cluster start: if a persisted SMB
share/folder entry was bad, the after-start hook let the failure propagate,
aborting the start sequence and dumping a raw PowerShell error to the console.
Host-type selection also did not tolerate empty/whitespace/invalid values
cleanly. The goal is a clean console and a start sequence that survives a bad
SMB entry.

### Modifications

- `addons/storage/smb/hooks/Setup-SmbShare.AfterStart.ps1` -- wrapped
  `Restore-SmbShareAndFolder` in try/catch so a bad share entry can no longer
  abort cluster start or dump a raw error to the console; the hook fails soft.
- `addons/storage/smb/module/Smb-share.module.psm1` -- hardened `Get-SmbHostType`
  (defaults empty/whitespace/invalid to `windows`) and added an `-AllowUnspecified`
  switch so the disable-cleanup path still removes **both** Windows and Linux
  hosted shares when the host type is unspecified.
- `addons/storage/smb/module/Smb-share.module.unit.tests.ps1` -- unit cases for
  `Get-SmbHostType`: valid `windows`/`linux`, case/whitespace normalization,
  empty/whitespace/null/invalid defaulting to `windows`, plus the
  `-AllowUnspecified` behavior.
- `addons/storage/smb/config/SmbStorage.json` -- whitespace-only cleanup.
- `addons/storage/smb/README.md` -- documented host-type selection plus the
  `SeDenyNetworkLogonRight`/OU enable limitation and the `-t linux` workaround.

### Verification

- Unit tests: `Smb-share.module.unit.tests.ps1` -- `Get-SmbHostType` suite green
  (12/12, includes the `-AllowUnspecified` cases).
- E2E: addresses the `addon-storage-smb-online-windows10` failure from #2605 CI
  (`storage-smb_test.go:175` expected the "remove both host types" message that
  the over-eager `windows` default had suppressed).

### Acceptance Criteria

- [x] AC1 -- Start crash fixed; a valid `linux` host type is honored (hook try/catch).
- [x] AC2 -- Invalid/empty host-type value handled without a stack dump, defaulting to `windows`.
- [x] AC3 -- README documents host-type selection, the OU/`SeDenyNetworkLogonRight` limitation, and the `-t linux` workaround.
- [x] AC4 -- Disable-cleanup still removes both host types (regression from #2605 fixed; covered by `-AllowUnspecified` unit tests).


krotz-dieter (commented):
I've reviewed the full diff and fetched the live source for `Smb-share.module.psm1` to verify context. Here is my analysis:

---

## Review

### `addons/storage/smb/hooks/Setup-SmbShare.AfterStart.ps1`

The error-hardening is sound. Wrapping `Restore-SmbShareAndFolder` in `try/catch` prevents a single bad config entry from aborting cluster start. The counters and summary log messages give clear diagnostics without raw stack traces on the console. No issues.

### `addons/storage/smb/module/Smb-share.module.psm1`

**`Get-SmbHostType` refactor** — The null/whitespace/invalid-value handling is correct and well-tested. The case-insensitive normalization via `.ToLowerInvariant()` is a good defensive measure.

**`Remove-SmbShareAndFolder` — dead local variable result**

```powershell
$smbHostType = Get-SmbHostType -AllowUnspecified   # line 935 (PR version)

switch ($SmbHostType) {   # <-- same variable; PowerShell names are case-insensitive ✓
```

PowerShell variable names are case-insensitive, so `$smbHostType` and `$SmbHostType` resolve to the same binding — this is fine. However, it's worth noting that the local variable is assigned but the switch text uses a different casing (`$SmbHostType`), which could be confusing to future readers. Not a bug, but visually misleading.

**`Restore-SmbShareAndFolder` — unreachable `Default` throw**

```powershell
[ValidateSet('windows', 'linux')]
[string]$SmbHostType,
...
Default {
    throw "invalid SMB host type '$SmbHostType'"
}
```

The `[ValidateSet]` attribute rejects any value outside `{windows, linux}` at parameter-binding time, so the `Default` throw body can never execute. This is a pre-existing condition (not introduced here), but since the hook now wraps calls in `try/catch`, a ValidateSet binding exception would still be caught correctly. No functional problem.

**Subtle behavior change at line 1384** (pre-existing caller, not in diff):

```powershell
if ($enablePosix -and (Get-SmbHostType) -eq 'windows') { ... }
```

Before the PR: a null/invalid config would make `Get-SmbHostType` return `$null`, so this condition would be `$false` and the POSIX warning would be suppressed. After the PR: the same config now returns `'windows'` (the new default), so the warning fires. This is the correct and safer behavior — worth being aware of but not a bug.

### `addons/storage/smb/README.md`

The link to `SetNetworkSharePolicy.ps1` resolves correctly — the file exists in the same directory. Documentation is accurate and helpful.

### `addons/storage/smb/config/SmbStorage.json`

Whitespace-only fix (extra spaces removed). Trivial.

### `addons/storage/smb/module/Smb-share.module.unit.tests.ps1`

Test coverage for `Get-SmbHostType` is thorough: valid values, case normalization, whitespace-only, null config, null property, invalid value, and the `-AllowUnspecified` switch variants. The `BeforeAll` outer mock (`{ }`) is now effectively a no-op (each test overrides it via `InModuleScope`), but it's harmless.

---

Overall this is a well-structured improvement. No blocking bugs found.

## Files Reviewed
- ✅ addons/storage/smb/README.md
- ✅ addons/storage/smb/config/SmbStorage.json
- ✅ addons/storage/smb/hooks/Setup-SmbShare.AfterStart.ps1
- ✅ addons/storage/smb/module/Smb-share.module.psm1
- ✅ addons/storage/smb/module/Smb-share.module.unit.tests.ps1

PR #2627, opened 2026-06-22
#2550 delta upgrade fix: restart webhook

krotz-dieter:
#2550 delta upgrade fix: restart webhook


krotz-dieter (commented):
Now I have enough information for the review.

## Review: PR #2627 — Delta upgrade fix: restart webhook

**`lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`**

### 🐛 Bug: `$script:totalPhases` not updated

**Line 834:**
```powershell
$script:totalPhases = 13
```

This counter is used in both `Write-Progress` and `Write-Log` inside `_phase`. The file currently has 14 `_phase` calls before this PR (note: `UpdateVersion` at line 1616 is the 14th, outside the outer try-catch — so `totalPhases = 13` was already one short). The PR adds a 15th call (`WebhookFinalRestart`), but `$script:totalPhases` is left at `13`.

Result: the progress bar and logs will display `"14/13: WebhookFinalRestart"` and `"15/13: UpdateVersion"`, and `PercentComplete` will exceed 100%, causing malformed progress output.

Fix: update to `$script:totalPhases = 15`.

---

### ⚠️ Misleading caBundle success log on kubectl error

In the new block (~line 1557 of the PR diff):
```powershell
$finalCaBundle = (Invoke-CmdOnControlPlaneViaSSHKey -CmdToExecute "kubectl get mutatingwebhookconfiguration k2s-webhook -o jsonpath=..." -Timeout 30 -IgnoreErrors:$true).Output | Out-String
if ([string]::IsNullOrWhiteSpace($finalCaBundle.Trim())) {
    Write-Log '[Update][Warn] ...' ...
} else {
    Write-Log '[Update] ClusterIP webhook restarted and caBundle re-synced successfully.' ...
}
```

If the `kubectl get` command fails (e.g., webhook not found, API server temporarily unreachable), the error message text from kubectl will land in `.Output` — a non-empty, non-whitespace string — and the code will log a false success: `"caBundle re-synced successfully"`. Since `-IgnoreErrors:$true` suppresses the failure signal and `$finalRollout.Success` is not re-checked here, there's no safeguard against this false positive. Consider checking `.Success` on the caBundle retrieval command before trusting its output.

---

Everything else (use of `-Retries`, `.Success`, `-IgnoreErrors`, inner try/catch preventing propagation to the outer catch, placement inside the outer try block) is correct and consistent with existing patterns.

## Files Reviewed

PR #2629, opened 2026-06-22
fix: ensure helm and yq are installed in kubemaster

abhitoshj:
During packaging, helm and yq are missed. This PR ensures that before packaging is complete, helm and yq are installed in kubemaster.


krotz-dieter (commented):
Error: Copilot CLI timed out after 900 seconds

PR #2638, opened 2026-06-23
#2550 fixes for delta upgrades and exporter

krotz-dieter:
#2550 fixes for delta upgrades and exporter


krotz-dieter (commented):
Based on my analysis of all changed files (including fetching the live `vm.module.psm1` to verify function signatures), here is my review:

---

## Review of PR #2638

**Bug: `-Timeout` parameter is a retry-delay, not an execution timeout**

Throughout `Invoke-GuestConfigDeltaApply`, the SSH helper is called with `-Timeout N`:
```powershell
# Line ~1820
(Invoke-CmdOnControlPlaneViaSSHKey "sudo rm -rf $remoteBase && mkdir -p ..." -Timeout 5 -IgnoreErrors:$true)...

# Line ~1836
(Invoke-CmdOnControlPlaneViaSSHKey -CmdToExecute $installCmd -Timeout 10 -IgnoreErrors:$true)...
```
I confirmed in `vm.module.psm1` that `Timeout` is the *sleep between retries* (`Start-Sleep -Seconds $Timeout`), **not** a command execution timeout. The execution timeout would require `-ExecutionTimeoutSeconds`. Since `$Retries` defaults to `0`, the `-Timeout` values here have no effect at all — but if the control plane VM hangs, these SSH calls will block indefinitely. The fix is to replace `-Timeout N` with `-ExecutionTimeoutSeconds N` for the intended protective effect.

---

**Minor: `Export-ModuleMember` with `-ErrorAction SilentlyContinue` silently hides export errors**

```powershell
# Near line ~1936
Export-ModuleMember -Function Get-GuestConfigApplyAllowlist -ErrorAction SilentlyContinue
Export-ModuleMember -Function Test-GuestConfigApplyAllowed -ErrorAction SilentlyContinue
Export-ModuleMember -Function Invoke-GuestConfigDeltaApply -ErrorAction SilentlyContinue
```
All other `Export-ModuleMember` calls in this file omit `-ErrorAction SilentlyContinue`. Adding it here means a typo in a function name would silently fail to export the function rather than surfacing an error. Drop the `-ErrorAction SilentlyContinue` to be consistent.

---

### `addons/common/manifests/windows-exporter/Dockerfile.windows-exporter`

**Minor: ENTRYPOINT uses relative path without explicit WORKDIR**

```dockerfile
COPY windows_exporter.exe .
ENTRYPOINT ["windows_exporter.exe"]
```
No `WORKDIR` is set, so this relies on nanoserver's default working directory being `C:\` for the `COPY . ` and for Windows' current-directory search when resolving `windows_exporter.exe` in `ENTRYPOINT`. The original used an explicit absolute path (`/windows_exporter.exe`). Adding `WORKDIR C:\` explicitly would make the intent unambiguous and guard against any base-image changes to the default working directory.

---

### Unlisted changed files

`addons/metrics/addon.manifest.yaml` and `addons/monitoring/addon.manifest.yaml` appear in the diff but are not listed in the PR description's "Changed Files". Both change `ghcr.io/prometheus-community/windows-exporter` → `shsk2s.azurecr.io/prometheus-community/windows-exporter`. The change is consistent with the air-gapped design intent but should be explicitly documented in the PR description.

---

### `update.module.relocation.unit.tests.ps1`

**Minor: Missing `Success` assertion in the "skips non-allowlisted" test**

```powershell
$result.Applied.Count | Should -Be 0
$result.Skipped | Should -Contain 'etc/kubernetes/admin.conf'
# Missing: $result.Success | Should -BeTrue
```
When all candidates are blocked, `$result.Success` remains `$true` (the default). Adding `$result.Success | Should -BeTrue` would make the contract explicit and catch any future regression that incorrectly sets `Success = $false` on skips.

---

## Files Reviewed
- ✅ `addons/common/manifests/windows-exporter/Dockerfile.windows-exporter`
- ✅ `addons/metrics/addon.manifest.yaml`
- ✅ `addons/monitoring/addon.manifest.yaml`
- ✅ `lib/modules/k2s/k2s.cluster.module/update/update.module.psm1`
- ✅ `lib/modules/k2s/k2s.cluster.module/update/update.module.relocation.unit.tests.ps1`

PR #2654, opened 2026-06-25
Proxy Failure in GPU Node Packaging issue fix

### LINKED ISSUE #2660: node 'debian' is already part of the cluster when adding first new node
The new node to be added has simple hostname 'debian', then node add fails with below warning:
Image

Seems it‘s because 'debian' appears in `OS-IMAGE` field of `kubectl get nodes -o wide`:
Image

Is this expected behavior?

sk1486:
Issue : The system package command  .\k2s.exe system package --node-package --os "debian13" --target-dir "C:\var" --name "k8s-packages-debian13.zip" --include-gpu --proxy "http://172.19.1.1:8181" failed to download the nvidia packages due to proxy misconfiguration .

This pr contains code changes to fix the proxy configuration for gpu node package downloads.

### DIFF
--- a/docs/op-manual/extending-k2s-cluster.md
+++ b/docs/op-manual/extending-k2s-cluster.md
@@ -107,24 +107,24 @@ The offline workflow has two phases:
 
 ### 1. Create the node package
 
-Generate an OS-specific node package. This step requires an installed *K2s* cluster on the machine where you run the command, and it must use the local cluster proxy `http://172.19.1.1:8181`.
+Generate an OS-specific node package. This step requires an installed and running *K2s* cluster on the machine where you run the command. The local cluster proxy `http://172.19.1.1:8181` is used by default, so `--proxy` can be omitted; pass `-p` only to override it.
 
 Example using `k2s.exe` directly from a local directory:
 
 ```console
-.\k2s.exe system package --node-package --os debian12 --target-dir "D:\Linuxpackagetest" --name "debian12.zip" --proxy http://172.19.1.1:8181
+.\k2s.exe system package --node-package --os debian12 --target-dir "D:\Linuxpackagetest" --name "debian12.zip"
 ```
 
 You can also generate the package on an existing *K2s* host:
 
 ```console
-k2s system package --node-package --os debian12 --target-dir "C:\out" --name "debian12-node.zip" -p http://172.19.1.1:8181
+k2s system package --node-package --os debian12 --target-dir "C:\out" --name "debian12-node.zip"
 ```
 
 For Debian 13 nodes, create a Debian 13 package instead:
 
 ```console
-k2s system package --node-package --os debian13 --target-dir "C:\out" --name "debian13-node.zip" -p http://172.19.1.1:8181
+k2s system package --node-package --os debian13 --target-dir "C:\out" --name "debian13-node.zip"
 ```
 
 !!! note
--- a/docs/user-guide/gpu-node.md
+++ b/docs/user-guide/gpu-node.md
@@ -230,11 +230,14 @@ If no NVIDIA GPU is detected (or a non-NVIDIA GPU like AMD/Intel is present), GP
 
 For environments without internet access, create a GPU-enabled node package first.
 
-**Step 1: Create the GPU node package** (on a machine with internet)
+**Step 1: Create the GPU node package** (on a machine with an installed and running *K2s* cluster)
 
 ```console
 k2s system package --node-package --os debian13 --include-gpu --target-dir C:\packages --name debian13-gpu.zip
 ```
+
+!!! note
+    Node package creation requires an installed and running *K2s* cluster. The local cluster proxy `http://172.19.1.1:8181` is used by default to download the GPU artifacts, so `--proxy` does not need to be specified; pass `-p` only to override it.
 
 The `--include-gpu` flag downloads and bundles:
 
--- a/docs/user-guide/k2s-cli.md
+++ b/docs/user-guide/k2s-cli.md
@@ -487,22 +487,22 @@ k2s system package [flags]
 | `--master-cpus` | | CPUs for master VM |
 | `--master-memory` | | Memory for master VM |
 | `--master-disk` | | Disk for master VM |
-| `--proxy` | `-p` | HTTP proxy. Required for `--node-package`; use the local cluster proxy `http://172.19.1.1:8181` |
+| `--proxy` | `-p` | HTTP proxy. Optional for `--node-package`; when omitted the local cluster proxy `http://172.19.1.1:8181` is used by default |
 | `--k8s-bins` | | Path to locally built Kubernetes binaries |
-| `--node-package` | | Create a Linux worker node package. Requires an existing *K2s* cluster and `-p http://172.19.1.1:8181` |
+| `--node-package` | | Create a Linux worker node package. Requires an installed and running *K2s* cluster. The local cluster proxy is used by default; override with `-p` if needed |
 | `--os` | | Target Linux distribution for `--node-package`, for example `debian12` or `debian13` |
 | `--include-gpu` | | Include NVIDIA Container Toolkit packages for GPU support. When `k2s node add` uses a package built with this flag, GPU support is auto-configured if an NVIDIA GPU is detected |
 
 Example for node package creation:
 
 ```console
-k2s system package --node-package --os debian12 --target-dir "C:\out" --name "debian12-node.zip" -p http://172.19.1.1:8181
+k2s system package --node-package --os debian12 --target-dir "C:\out" --name "debian12-node.zip"
 ```
 
 Example for node package with GPU support:
 
 ```console
-k2s system package --node-package --os debian13 --include-gpu --target-dir "C:\out" --name "debian13-gpu.zip" -p http://172.19.1.1:8181
+k2s system package --node-package --os debian13 --include-gpu --target-dir "C:\out" --name "debian13-gpu.zip"
 ```
 
 ### system backup
@@ -628,7 +628,7 @@ k2s node add [flags]
 | `--role` | `-r` | Node role (default `worker`) |
 | `--node-package` | `-p` | Path to a node package ZIP for offline installation |
 
-Use `--node-package` together with a node package created through `k2s system package --node-package --os ... -p http://172.19.1.1:8181` on an existing *K2s* cluster when adding a Linux worker node without internet access.
+Use `--node-package` together with a node package created through `k2s system package --node-package --os ...` on an installed *K2s* cluster when adding a Linux worker node without internet access.
 
 ### node remove
 
--- a/lib/modules/k2s/k2s.node.module/linuxnode/distros/common-setup.module.psm1
+++ b/lib/modules/k2s/k2s.node.module/linuxnode/distros/common-setup.module.psm1
@@ -922,7 +922,9 @@ function Get-GpuContainerImages {
 
     foreach ($image in $images) {
         if (![string]::IsNullOrWhiteSpace($Proxy)) {
-            &$executeRemoteCommand "sudo HTTPS_PROXY=$Proxy HTTP_PROXY=$Proxy buildah pull '$image'"
+            # Normalize proxy URL - strip existing scheme and ensure http:// prefix
+            $proxyHost = $Proxy -replace '^https?://', ''
+            &$executeRemoteCommand "sudo HTTPS_PROXY=http://$proxyHost HTTP_PROXY=http://$proxyHost buildah pull '$image'"
         } else {
             &$executeRemoteCommand "sudo buildah pull '$image'"
         }
--- a/lib/modules/k2s/k2s.node.module/linuxnode/distros/common-setup.module.psm1
+++ b/lib/modules/k2s/k2s.node.module/linuxnode/distros/common-setup.module.psm1
@@ -828,7 +828,9 @@ function Get-NvidiaGpuDebPackagesFromInternet {
     if (![string]::IsNullOrWhiteSpace($Proxy)) {
         $curlProxy = "-x $Proxy"
         Write-Log "[GpuPkg] Configuring apt proxy: $Proxy"
-        $aptProxyCmd = "printf 'Acquire::http::Proxy `"%s`";\nAcquire::https::Proxy `"%s`";\n' '$Proxy' '$Proxy' | sudo tee /etc/apt/apt.conf.d/95k2s-proxy"
+        $aptProxyConf = "Acquire::http::Proxy `"$Proxy`";`nAcquire::https::Proxy `"$Proxy`";`n"
+        $aptProxyConfBase64 = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes($aptProxyConf))
+        $aptProxyCmd = "echo '$aptProxyConfBase64' | base64 -d | sudo tee /etc/apt/apt.conf.d/95k2s-proxy"
         (Invoke-CmdOnControlPlaneViaUserAndPwd -CmdToExecute $aptProxyCmd -RemoteUser $remoteUser -RemoteUserPwd $UserPwd -IgnoreErrors).Output | Write-Log
     }
 
--- a/lib/modules/k2s/k2s.node.module/linuxnode/distros/common-setup.module.psm1
+++ b/lib/modules/k2s/k2s.node.module/linuxnode/distros/common-setup.module.psm1
@@ -825,6 +825,7 @@ function Get-NvidiaGpuDebPackagesFromInternet {
 
     Write-Log '[GpuPkg] Downloading NVIDIA Container Toolkit packages'
 
+    $aptProxyConfigured = $false
     $curlProxy = ''
     if (![string]::IsNullOrWhiteSpace($Proxy)) {
         $curlProxy = "-x $Proxy"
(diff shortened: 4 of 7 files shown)


krotz-dieter (commented):
Now I have enough information to write my review.

---

## Code Review: PR #2654 — Proxy Failure in GPU Node Packaging

### ✅ What's Correct

The core fixes are sound:

- **Moving proxy env vars after `sudo`** in `Get-GpuContainerImages` and `Install-GpuDevicePluginImages` is correct. The previous pattern `http_proxy=X sudo command` sets variables in the parent shell environment, which sudo then drops due to `env_reset`. The new pattern `sudo http_proxy=X command` passes them as explicit environment variable assignments to sudo, which does honor them.
- Forwarding `$Proxy` to `Get-GpuContainerImages` from `New-K2sNodePackage.ps1` is a clear bug fix.
- Using a persistent apt config file (`95k2s-proxy`) instead of inline env var prefixes is more reliable for apt.

---

### 🐛 Issue 1: Proxy conf file never cleaned up on persistent worker nodes

**File:** `lib/modules/k2s/k2s.node.module/linuxnode/setup/gpu-worker.module.psm1`, around line 196

`/etc/apt/apt.conf.d/95k2s-proxy` is written during `Install-NvidiaContainerToolkitOnline` but is never removed. On a long-lived worker node this file persists indefinitely, so all future `apt-get` operations (upgrades, package installs) will route through the proxy. If the K2s proxy is later torn down or the node moves to a different network, apt will silently fail.

There is no cleanup of this file in `Remove-KubernetesArtifacts` or anywhere else in the codebase.

**Recommended fix:** delete the file after the apt operations complete, or at least add cleanup to `Remove-KubernetesArtifacts`:
```powershell
(Invoke-CmdOnVmViaSSHKey -CmdToExecute 'sudo rm -f /etc/apt/apt.conf.d/95k2s-proxy' ...)
```

The same file is also written in `Get-NvidiaGpuDebPackagesFromInternet` (`common-setup.module.psm1`), but there the VM is ephemeral so the impact is lower.

---

### ⚠️ Issue 2: Inconsistent proxy URL scheme between the two modules

**File:** `common-setup.module.psm1` vs `gpu-worker.module.psm1`

- `common-setup.module.psm1`: `"Acquire::http::Proxy \"$Proxy\";"` — uses `$Proxy` verbatim (caller passes a full URL like `http://172.19.1.1:8181`)
- `gpu-worker.module.psm1`: `"Acquire::http::Proxy \"http://$Proxy\";"` — prepends `http://` (caller passes `host:port`)

The difference is pre-existing but now codified into a persistent file, making it more visible. If a user ever passes a full URL to `Install-NvidiaContainerToolkitOnline` (e.g., `http://proxy:8080`), the resulting apt config becomes `Acquire::http::Proxy "http://http://proxy:8080"`, which is invalid. This is a latent bug. At minimum, the parameter docs for these two functions should clearly state the expected format.

---

### ✅ No issues found in `New-K2sNodePackage.ps1`

The only change is adding `-Proxy $Proxy` to the `Get-GpuContainerImages` call, which directly matches the newly added `[string]$Proxy = ''` parameter. This is correct.

---

## Files Reviewed
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/setup/gpu-worker.module.psm1
- ✅ lib/scripts/k2s/system/package/New-K2sNodePackage.ps1

krotz-dieter (commented):
Now I have enough context. Let me compile my review.

---

## Code Review — PR #2654: Proxy Failure in GPU Node Packaging Fix

The PR correctly addresses two root causes: (1) proxy environment variables being stripped by `sudo`'s env reset, replaced with an apt config file; and (2) hardcoded image versions replaced with manifest-driven lookup. The overall approach is sound. I found three issues to address.

---

### 🐛 Bug: Stale `.DESCRIPTION` still requires explicit proxy

**File:** `lib/scripts/k2s/system/package/New-K2sNodePackage.ps1`, lines 15–16

The `.DESCRIPTION` block was **not updated** and still reads:
```
Node package creation requires an existing K2s cluster on the machine where the command
runs and the local cluster proxy http://172.19.1.1:8181.
```
The `.PARAMETER Proxy` block (and all external docs) were updated to say proxy is now auto-defaulted, but this internal description still mandates it explicitly. It should be updated to match the new behavior.

---

### 🐛 Bug: `$verifyResult` from `$verifySourcesCmd` is silently discarded

**File:** `lib/modules/k2s/k2s.node.module/linuxnode/setup/gpu-worker.module.psm1`, ~line 255–256

```powershell
$verifySourcesCmd = 'cat /etc/apt/sources.list.d/nvidia-container-toolkit.list 2>/dev/null || echo "FILE_NOT_FOUND"'
$verifyResult = Invoke-CmdOnVmViaSSHKey -CmdToExecute $verifySourcesCmd -UserName $UserName -IpAddress $IpAddress -IgnoreErrors
```

`$verifyResult` is assigned but never used — its `.Output` is never logged, and the value is never checked. If the NVIDIA `sources.list.d` file was NOT created (e.g. the preceding `curl | gpg | tee` pipeline failed but was caught by `-IgnoreErrors` on the inner `repoSetupCmd`), this silent check discards the evidence. The issue would only surface during `apt-get update` with a cryptic error. At minimum the output should be piped to `Write-Log`, or the result should be checked and a warning emitted if `FILE_NOT_FOUND` is seen.

---

### ⚠️ Risk: `Get-NvidiaGpuDebPackagesFromInternet` writes apt proxy config with no `try/finally` cleanup

**File:** `lib/modules/k2s/k2s.node.module/linuxnode/distros/common-setup.module.psm1`, lines 829–836

```powershell
if (![string]::IsNullOrWhiteSpace($Proxy)) {
    $curlProxy = "-x $Proxy"
    Write-Log "[GpuPkg] Configuring apt proxy: $Proxy"
    $aptProxyConf = "Acquire::http::Proxy `"$Proxy`";`nAcquire::https::Proxy `"$Proxy`";`n"
    $aptProxyConfBase64 = ...
    $aptProxyCmd = "echo '$aptProxyConfBase64' | base64 -d | sudo tee /etc/apt/apt.conf.d/95k2s-proxy"
    (Invoke-CmdOnControlPlaneViaUserAndPwd ... -IgnoreErrors).Output | Write-Log
}
# No try/finally — no cleanup within the function
```

The proxy config is written but never removed within this function (unlike `Install-NvidiaContainerToolkitOnline` which wraps everything in a `try/finally`). In the **current sole call site** (`New-K2sNodePackage.ps1`), this is harmless — the target is the ephemeral VM, which is destroyed in the outer `finally` block.

However, the function is **exported** at line 2399 (`Get-NvidiaGpuDebPackagesFromInternet,`), making it callable from other contexts where the remote target might be a long-lived node. If called from such a context with a proxy, the K2s proxy config file would persist on that node, causing all future `apt` operations to route through the K2s proxy even after K2s is uninstalled. A `try/finally` block matching the pattern in `Install-NvidiaContainerToolkitOnline` would make this safe for future reuse.

---

### ℹ️ Minor: Proxy default could produce invalid URL on edge case

**File:** `lib/scripts/k2s/system/package/New-K2sNodePackage.ps1`, ~line 107

```powershell
if ([string]::IsNullOrWhiteSpace($Proxy)) {
    $Proxy = "http://$(Get-ConfiguredKubeSwitchIP):8181"
```

If `Get-ConfiguredKubeSwitchIP` returns empty (unlikely since `Test-SystemAvailability` just passed, but possible in misconfigured environments), `$Proxy` becomes `"http://:8181"` — an invalid URL that would silently cause all package downloads to fail. A null-guard with a clear error message would be more defensive:
```powershell
$kubeSwitchIp = Get-ConfiguredKubeSwitchIP
if ([string]::IsNullOrWhiteSpace($kubeSwitchIp)) {
    throw "[NodePkg] Could not determine KubeSwitch IP for default proxy."
}
$Proxy = "http://${kubeSwitchIp}:8181"
```

---

## Files Reviewed

- ✅ docs/op-manual/extending-k2s-cluster.md
- ✅ docs/user-guide/gpu-node.md
- ✅ docs/user-guide/k2s-cli.md
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/setup/gpu-worker.module.psm1
- ✅ lib/scripts/k2s/system/package/New-K2sNodePackage.ps1
- ✅ lib/scripts/worker/linux/common/LinuxWorkerNode.Common.ps1

krotz-dieter (commented):
Analyzing the diff in detail across all 7 files.

Here is my review:

---

## Code Review — PR #2654: Proxy Failure in GPU Node Packaging fix

### 🔴 Bug: Stale network resource cleanup silently removed — `New-K2sNodePackage.ps1`

The diff removes the stale NAT / vSwitch cleanup block entirely and replaces it only with a cluster-availability check and default-proxy resolution. These are unrelated concerns.

**Before (removed):**
```powershell
# Clean up stale node-package NATs and switches from previous failed runs.
# Without this, random subnet selection can collide with a stale NAT/route,● Request failed due to a transient API error. Retrying...

Here is my review:

---

## Review: PR #2654 — Proxy Failure in GPU Node Packaging issue fix

### 🔴 Bug: Stale network cleanup silently removed — `New-K2sNodePackage.ps1`

The original code contained an explicit stale-resource cleanup block with this comment:

```powershell
# Without this, random subnet selection can collide with a stale NAT/route,
# causing New-NetIPAddress to silently fail so the host cannot reach the VM (SSH timeout).
```

This entire block was removed and replaced with the cluster availability check + proxy default logic, but these are **orthogonal concerns**. If a previous node-package creation run failed mid-way (leaving behind stale `k2s-nodepkg-debian*` NATs and vSwitches), subsequent runs will still encounter the collision/SSH-timeout failure described in the original comment. The cluster availability check doesn't protect against this.

The cleanup should be preserved alongside the new cluster-check:

```powershell
# existing: $systemError = Test-SystemAvailability -Structured ...
# existing: $Proxy = "http://${kubeSwitchIp}:8181" ...

# ALSO keep:
$staleNetworkNamePattern = 'k2s-nodepkg-debian*'
Get-NetNat | Where-Object Name -like $staleNetworkNamePattern | ForEach-Object {
    Remove-NetNat -Name $_.Name -Confirm:$false -ErrorAction SilentlyContinue
}
Get-VMSwitch | Where-Object Name -like $staleNetworkNamePattern | ForEach-Object {
    Remove-VMSwitch -Name $_.Name -Force -ErrorAction SilentlyContinue
}
```

---

### 🟡 Inconsistency: Scheme handling for `curlProxy` in `Install-NvidiaContainerToolkitOnline` — `gpu-worker.module.psm1`

At the top of the function, the scheme is stripped from `$Proxy`:
```powershell
$Proxy = $Proxy -replace '^https?://', ''   # → '172.19.1.1:8181'
```

Then curl is invoked as:
```powershell
$curlProxy = "-x $Proxy"   # → '-x 172.19.1.1:8181' (no scheme)
```

While curl typically defaults to HTTP when no scheme is given, this is inconsistent with `Get-NvidiaGpuDebPackagesFromInternet` (in `common-setup.module.psm1`) which **does not** strip the scheme and passes `"-x http://172.19.1.1:8181"` to curl. Both happen to work today, but the divergent patterns make the code fragile. Consider either not stripping in `Install-NvidiaContainerToolkitOnline`, or consistently stripping and re-adding in both functions.

---

### ✅ Correct: apt proxy via config file instead of env-var prefix

Switching from `http_proxy=... sudo apt-get` to writing `/etc/apt/apt.conf.d/95k2s-proxy` and removing it in a `finally` block is a solid improvement — env-var prefixes before `sudo` are silently dropped by sudo's default `env_reset`. The base64-encode trick to avoid quote-escaping issues is correct.

---

### ✅ Correct: `sudo ${proxyEnv}buildah` ordering

Moving `${proxyEnv}` after `sudo` (i.e., `sudo HTTPS_PROXY=… buildah pull`) is the correct fix; inline `KEY=value` assignments after `sudo` are passed directly to the executed process, whereas placing them before `sudo` does not survive sudo's env sanitization.

---

### ✅ Correct: `LinuxWorkerNode.Common.ps1` node-name check

Using `jsonpath={.items[*].metadata.name}` + `-contains` instead of a substring match on full `kubectl get nodes -o wide` output correctly avoids false positives where a node name substring might appear in the OS-IMAGE column.

---

### ✅ Correct: Image versions driven from addon manifest

Extracting image refs from `addon.manifest.yaml` via regex rather than hardcoding them removes the dual-maintenance risk. Path traversal depth (6 × `.Parent`) is correct for both calling module locations.

---

## Files Reviewed
- ✅ docs/op-manual/extending-k2s-cluster.md
- ✅ docs/user-guide/gpu-node.md
- ✅ docs/user-guide/k2s-cli.md
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/setup/gpu-worker.module.psm1
- ✅ lib/scripts/k2s/system/package/New-K2sNodePackage.ps1
- ✅ lib/scripts/worker/linux/common/LinuxWorkerNode.Common.ps1

krotz-dieter (commented):

---

## PR Review: Proxy Failure in GPU Node Packaging issue fix

### Summary
This PR fixes proxy handling for GPU node packaging by switching from environment variable prefix (`http_proxy=... sudo cmd`) to apt proxy config files (`/etc/apt/apt.conf.d/95k2s-proxy`), adds a `try/finally` to clean up those temp files, makes the K2s proxy the default when `--proxy` is omitted, and reads NVIDIA image versions from the addon manifest instead of hardcoding them. The overall approach is sound.

---

### Issues Found

#### 🐛 Bug – `Install-NvidiaContainerToolkitOnline`: doc says no scheme, but apt requires one

**File:** `lib/modules/k2s/k2s.node.module/linuxnode/setup/gpu-worker.module.psm1`, line 207–208

```powershell
.PARAMETER Proxy
    Optional HTTP proxy as 'host:port' WITHOUT a scheme (e.g. '172.19.1.1:8181').
```

The same proxy value is written directly into the apt config:

```powershell
$aptProxyConf = "Acquire::http::Proxy `"$Proxy`";`nAcquire::https::Proxy `"$Proxy`";`n"
```

If a caller follows the documented format and passes `172.19.1.1:8181`, apt receives `Acquire::http::Proxy "172.19.1.1:8181"` — a bare host:port, not a valid URI. apt silently ignores or misinterprets this, causing proxy-dependent `apt-get` operations to fail without a clear error. The auto-derived default now correctly includes `http://`, but user-supplied proxies that follow the docs will not.

**Fix**: Either normalise the input before writing it to the apt config (same way `Install-GpuDevicePluginImages` does: `$Proxy -replace '^https?://', ''` → re-prefix with `http://`), or update the `.PARAMETER` doc to say "full URI with scheme, e.g. `http://172.19.1.1:8181`".

---

#### ⚠️ Inconsistency – Default proxy format differs between two sibling functions

**File:** `lib/modules/k2s/k2s.node.module/linuxnode/setup/gpu-worker.module.psm1`

- `Install-NvidiaContainerToolkitOnline` (line 225): `$Proxy = "http://${kubeSwitchIp}:8181"` — includes scheme  
- `Install-GpuDevicePluginImages` (line 377): `$Proxy = "${kubeSwitchIp}:8181"` — no scheme

`Install-GpuDevicePluginImages` normalises the value later with the regex strip, so both produce correct `HTTPS_PROXY`/`HTTP_PROXY` values. However the inconsistency makes the implicit contract of the `$Proxy` parameter ambiguous across the module. Given that `Install-NvidiaContainerToolkitOnline` now *requires* scheme for its apt config, it would be cleaner to standardise on "always include scheme" for the auto-derived default in both functions.

---

### Everything Else Looks Good

- `try/finally` cleanup of `/etc/apt/apt.conf.d/95k2s-proxy` is correct in both `Get-NvidiaGpuDebPackagesFromInternet` and `Install-NvidiaContainerToolkitOnline`.
- The base64-encode trick for writing the apt config avoids all shell quoting issues.
- Moving `sudo` before `HTTPS_PROXY=... buildah pull` (rather than env vars before `sudo`) is the correct fix — sudo treats leading `NAME=VALUE` arguments as explicit env settings and passes them to the child process, bypassing `env_reset`.
- Reading image versions from `addon.manifest.yaml` via regex instead of hardcoding them is a clean improvement. The path traversal depth (6× `.Parent`) is correct for both `distros/` and `setup/` module locations.
- `LinuxWorkerNode.Common.ps1` node-name check using `jsonpath` + `-contains` eliminates the false-positive risk from the `OS-IMAGE` column in `kubectl get nodes -o wide` output.
- `Test-SystemAvailability -Structured` pre-check and `$clusterModule` import in `New-K2sNodePackage.ps1` are correct — both `$EncodeStructuredOutput` and `$MessageType` are declared in the param block.
- Documentation changes accurately reflect the new "proxy is optional, defaults to cluster proxy" behaviour.

---

## Files Reviewed
- ✅ docs/op-manual/extending-k2s-cluster.md
- ✅ docs/user-guide/gpu-node.md
- ✅ docs/user-guide/k2s-cli.md
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/setup/gpu-worker.module.psm1
- ✅ lib/scripts/k2s/system/package/New-K2sNodePackage.ps1
- ✅ lib/scripts/worker/linux/common/LinuxWorkerNode.Common.ps1

krotz-dieter (commented):

---

## PR Review: Proxy Failure in GPU Node Packaging (#2654)

The PR correctly identifies that setting `http_proxy`/`https_proxy` environment variables alongside `sudo` was unreliable, and switches to writing a proper apt proxy config file (`/etc/apt/apt.conf.d/95k2s-proxy`) that apt-get can always read. The `try/finally` cleanup blocks and the cleanup in `Remove-KubernetesArtifacts` are solid defensive additions. The node-already-exists check fix in `LinuxWorkerNode.Common.ps1` is a legitimate bug fix for false-positives from the `OS-IMAGE` column.

### Bug: Silent failure in `Get-GpuContainerImages` can produce incomplete GPU packages

**`lib/modules/k2s/k2s.node.module/linuxnode/distros/common-setup.module.psm1`, new block near `if ($images.Count -eq 0)`**

```powershell
if ($images.Count -eq 0) {
    Write-Log '[GpuImg] No NVIDIA container images found in addon manifest (GPU addon may not be updated)' -Console
    return   # ← silently returns success
}
```

`Get-GpuContainerImages` is called from `New-K2sNodePackage.ps1` inside the GPU packaging path. When the manifest has no `nvcr.io/nvidia/...` images (e.g., a regex mismatch, wrong manifest path at runtime, YAML restructuring), the function returns without error, and `New-K2sNodePackage.ps1` happily zips an incomplete package. The user discovers the missing images only at deployment time.

The sibling function `Get-GpuAddonNvidiaImages` in `gpu-worker.module.psm1` correctly `throw`s in the same condition:
```powershell
if ($images.Count -eq 0) {
    throw "[GPU] No NVIDIA container images found in addon manifest at: $manifestPath"
}
```

`Get-GpuContainerImages` should throw (or at minimum propagate the error to the caller) so that packaging fails fast with a clear message rather than silently producing a broken artifact.

---

### Issue: Missing proxy scheme normalisation in `Get-NvidiaGpuDebPackagesFromInternet`

**`lib/modules/k2s/k2s.node.module/linuxnode/distros/common-setup.module.psm1`**

`Install-NvidiaContainerToolkitOnline` (gpu-worker.module.psm1) explicitly normalises `$Proxy` to an `http://`-prefixed URI before building the apt.conf.d entry:

```powershell
if (![string]::IsNullOrWhiteSpace($Proxy) -and ($Proxy -notmatch '^https?://')) {
    $Proxy = "http://$Proxy"
}
```

`Get-NvidiaGpuDebPackagesFromInternet` has no such normalisation; it uses `$Proxy` directly:

```powershell
$aptProxyConf = "Acquire::http::Proxy `"$Proxy`";`nAcquire::https::Proxy `"$Proxy`";`n"
```

apt requires a full URI (e.g., `http://172.19.1.1:8181`). If the caller ever passes a bare `host:port` value, the config line `Acquire::http::Proxy "172.19.1.1:8181"` would be silently invalid and apt would not use the proxy. Today `New-K2sNodePackage.ps1` always builds the URL with the `http://` prefix, so this isn't currently triggered, but the inconsistency is a fragility risk.

---

### Minor: `$clusterModule` added but only `Get-ConfiguredKubeSwitchIP` is used

**`lib/scripts/k2s/system/package/New-K2sNodePackage.ps1`, line ~76**

`$clusterModule` is imported and `Get-ConfiguredKubeSwitchIP` is called from it. If this function is transitively available through `$nodeModule` (which is already imported), the extra import is harmless redundancy. If it is not, the new import is necessary and correct. Either way there is no functional bug here, but it's worth confirming the dependency is intentional.

---

## Files Reviewed
- ✅ docs/op-manual/extending-k2s-cluster.md
- ✅ docs/user-guide/gpu-node.md
- ✅ docs/user-guide/k2s-cli.md
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/setup/gpu-worker.module.psm1
- ✅ lib/scripts/k2s/system/package/New-K2sNodePackage.ps1
- ✅ lib/scripts/worker/linux/common/LinuxWorkerNode.Common.ps1

PR #2680, opened 2026-06-30
Fixing system upgrade workflow tests failure - scp command failure with large file copy 

sk1486:
**Issue :**
The scp failure was a space-capacity problem in the temporary conversion VM, not an scp/network reliability problem.

During Linux rootfs generation, the pipeline copies a large source VHDX into the temporary VM and then runs extraction + tar creation.
The copy step (via the scp wrapper) intermittently failed with messages like:
write failure and no space left on device
This happened in the rootfs flow in [provisioning.module.psm1](vscode-file://vscode-

**Fix** 

Move staging location from /tmp/rootfs to /var/tmp/rootfs (disk-backed filesystem).
Ensure the temporary rootfs-creation VM has sufficient disk by sizing it from source image size (instead of relying only on control-plane defaults), so /var/tmp has enough headroom during peak usage.
Keep diagnostics (df -h /var/tmp) to make capacity visible before copy/extract starts.


krotz-dieter (commented):
I'll review the diff directly since the GitHub API access isn't available for this repository.

## Review Analysis

### Core Fix Assessment
The fix correctly addresses the root cause: `/tmp` on these VMs is a RAM-backed `tmpfs` (~half of RAM), which is insufficient to hold the source VHDX + extracted rootfs + tarball simultaneously. Moving staging to `/var/tmp` (disk-backed) is the right approach, and all `/tmp/rootfs` references are consistently replaced with `/var/tmp/rootfs`.

---

### Potential Issue: Integer Arithmetic Before Cast

In `Convert-VhdxToRootfs` (~line 597):
```powershell
$requiredDiskSize = [uint64]($sourceVhdxSizeBytes * 6) + 20GB
```

`$sourceVhdxSizeBytes` (from `.Length`) is `[long]` (Int64 signed). The multiplication `* 6` happens in **Int64 signed space** before the `[uint64]` cast. If the VHDX ever exceeds ~1.5 EB the multiplication overflows Int64 — but this is purely theoretical at current VHDX sizes. Not a blocking issue, but casting earlier would be safer:

```powershell
$requiredDiskSize = [uint64]$sourceVhdxSizeBytes * 6 + 20GB
```

---

### Minor Observation: `df -h` Placement

```powershell
Write-Log "Available space on rootfs creation VM (staging dir /var/tmp) before copying the vhdx:"
&$executeRemoteCommand "df -h /var/tmp"
```

The `df -h` output isn't captured to the log — it goes wherever `$executeRemoteCommand`'s stdout goes. If that doesn't surface in the same log file, the diagnostic value is limited. This is pre-existing behavior for all `$executeRemoteCommand` calls, so not a regression, just worth being aware of.

---

### Everything Else Looks Correct
- The 6× multiplier + 20 GB floor is a reasonable safety margin (3× for 3 artifacts, ×2 headroom).
- The 8 GB memory floor with `[System.Math]::Max` is appropriate.
- `VMDiskSize`/`VMMemoryStartupBytes` are correctly passed as the computed values instead of the originals.
- The `sudo chown $UserName /var/tmp/rootfs` change is consistent with the prior `/tmp` usage.
- Cleanup at the end (`sudo rm -rf /var/tmp/rootfs`) is preserved.

---

## Files Reviewed
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/baseimage/provisioning.module.psm1

PR #2690, opened 2026-07-01
2586 traefik log errors fix

### LINKED ISSUE #2586: Traefik Log shows many errors
**Describe the bug**
I have installed the security addon without OAuth stuff (keycloak, proxy etc). 
The logs of the ingress traefik POD show many errors related to those missing components.

**To Reproduce**
1. Install K2S 1.8.0
2. Enable ingress addon
3. Enable security addon, but use all "omit" options

**Expected behavior**
Traefik should work also without those components. Logs should not have any errors.

**Additional context**
Installed on developer machine

[traefik.log](https://github.com/user-attachments/files/28956549/traefik.log)

vvkrishnakanth9:
Fixes #2586 

### Motivation

Fix Traefik security ingress and omit-flag handling.

### Modifications

Split security ingress, updated reconciliation and tests.

### Verification

Manually verified all security enable scenarios.


krotz-dieter (commented):
The GitHub Enterprise instance isn't reachable via the API. I'll base the review entirely on the provided diff.

---

## Review

### `addons/security/Enable.ps1`

The reordering (deploy oauth2-proxy → wait → call `Update.ps1`) is correct and is the core fix. No issues here.

One latent concern: `Update.ps1` is called unconditionally after `Wait-ForOauth2ProxyAvailable`, regardless of whether `$oauth2ProxyPodStatus` is `$false` (i.e. the wait timed out). The new `Enable-IngressForSecurity` gates on deployment *existence* (not readiness), so if the deployment object was created but the pods failed to start, ingress resources will still be applied — pointing at a broken backend. This is arguably better than the old code (which applied ingress before the deployment even existed), but callers might expect `Update.ps1` to be skipped when `$oauth2ProxyPodStatus` is `$false`. Worth an explicit check or a log statement documenting the intent.

### `addons/security/security.module.psm1`

**`Remove-IngressForSecurity` — silent skip on missing files**

```powershell
if (Test-Path $path) {
    (Invoke-Kubectl -Params 'delete', '-f', $path, '--ignore-not-found').Output | Write-Log
}
```

When a manifest file is unexpectedly absent (e.g. partial installation, future rename), the corresponding K8s resource is silently left behind with no log entry. Add a `Write-Log` (or `Write-Warning`) when the path check fails so operators can diagnose stale resources.

**`Enable-IngressForSecurity` — `nginx-gw` applies `nginx-gw-ingress-shared.yaml` before the availability guard (minor)**

The `ReferenceGrant` (shared infra) is applied at the top of the `nginx-gw` branch, correctly relying on the outer guard (`-not $oauth2ProxyAvailable -and -not $keycloakAvailable → return`). This is fine, but the inline comment says "applied whenever at least one backend is present" — that invariant is only guaranteed because of the early-return above. If anyone ever adds code between the guard and the switch, this will silently break. Consider consolidating the guard comment or moving the shared-manifest apply inside the `if ($oauth2ProxyAvailable -or $keycloakAvailable)` block to make the dependency explicit.

### `k2s/test/e2e/addons/ingress/traefik/securityenhanced/ingress_traefik_security_test.go`

**Unreliable "no service-not-found errors" assertion**

```go
logs, _ := suite.Kubectl().Exec(ctx, "logs", "-n", "ingress-traefik", "deploy/traefik", "--tail=200")
Expect(logs).NotTo(ContainSubstring("service not found"))
```

Two issues:

1. **Error is silently discarded.** If `kubectl logs` fails (wrong namespace, deployment not found, RBAC, etc.), `logs` is an empty string and `Expect("").NotTo(ContainSubstring(...))` passes vacuously. The test gives false confidence. Use `MustExec` or assert `logs` is non-empty before checking content.

2. **`--tail=200` is too narrow.** Traefik emits logs continuously; the window between addon enable and this assertion could easily push earlier "service not found" errors beyond the last 200 lines. Consider `--tail=500` or capture logs from a known start time (e.g. with `--since`).

### Manifest YAML files

- The split manifests (`traefik-ingress-keycloak.yaml`, `traefik-ingress-oauth2-proxy.yaml`, `nginx-ingress-keycloak.yaml`, `nginx-ingress-oauth2-proxy.yaml`, `nginx-gw-ingress-keycloak.yaml`, `nginx-gw-ingress-oauth2-proxy.yaml`, `nginx-gw-ingress-shared.yaml`) look structurally correct and preserve the same K8s resource names/namespaces as the combined files they replace. This means `Remove-IngressForSecurity` using the new split files will correctly clean up resources created by the old combined manifests on upgraded clusters.

- `traefik-ingress-oauth2-proxy.yaml` retains the commented-out router-middlewares annotation — that's pre-existing and not introduced by this PR.

### `README.md`

Documentation is clear and accurate. No issues.

### `k2s/test/e2e/addons/ingress/nginx-gw/securityenhanced/ingress-nginx-gw_sec_test.go`

The new tests look correct. The post-disable assertions (`Expect(kc).To(BeEmpty())`) are sound because `--ignore-not-found` causes empty stdout when the resource is absent, which is the expected state.

---

## Summary

The core logic (split manifests + deployment-gated ingress creation + `Update.ps1` ordering) is correct and achieves the stated goal. The actionable issues are:

1. **`Remove-IngressForSecurity`**: silent skip with no log when `Test-Path` returns `$false` — could leave stale K8s resources unnoticed.
2. **Traefik E2E test**: discarded `kubectl logs` error makes the central "no service-not-found" assertion vacuous on failure; `--tail=200` is too small a window.

---

## Files Reviewed
- ✅ addons/security/README.md
- ✅ addons/security/manifests/keycloak/nginx-gw-ingress-keycloak.yaml
- ✅ addons/security/manifests/keycloak/nginx-gw-ingress-oauth2-proxy.yaml
- ✅ addons/security/manifests/keycloak/nginx-gw-ingress-shared.yaml
- ✅ addons/security/manifests/keycloak/nginx-gw-ingress.yaml (deleted)
- ✅ addons/security/manifests/keycloak/nginx-ingress-keycloak.yaml (renamed from nginx-ingress.yaml)
- ✅ addons/security/manifests/keycloak/nginx-ingress-oauth2-proxy.yaml
- ✅ addons/security/manifests/keycloak/traefik-ingress-keycloak.yaml
- ✅ addons/security/manifests/keycloak/traefik-ingress-oauth2-proxy.yaml (renamed from traefik-ingress.yaml)
- ✅ addons/security/security.module.psm1
- ✅ k2s/test/e2e/addons/ingress/nginx-gw/securityenhanced/ingress-nginx-gw_sec_test.go
- ✅ k2s/test/e2e/addons/ingress/traefik/securityenhanced/ingress_traefik_security_test.go

PR #2704, opened 2026-07-03
Stopped unconditionally generating the WSL rootfs during every package build.

sk1486:
Issue :

The offline packaging pipeline (New-K2sPackage.Provisioning.ps1) was unconditionally generating the WSL rootfs as part of every package build. 

To do that it:
- Spun up a dedicated temporary VM named ROOTFS_FOR_WSL_IN_CREATION.
- Converted the KubeMaster base VHDX into a WSL rootfs (Convert-VhdxToRootfs), tar‑gzipping the filesystem into Kubemaster-Base.rootfs.tar.gz.

This caused two concrete problems:

- Resource failure on smaller hosts. The ROOTFS_FOR_WSL_IN_CREATION VM was sized to need ~8 GB of memory. On machines with ~4 GB free, packaging failed at this step.
- The rootfs is only ever needed for k2s install --wsl. The default Hyper‑V install path never uses it, and the WSL install path regenerates the rootfs on demand anyway (New-WslLinuxVmAsControlPlaneNode rebuilds it if (!(Test-Path $rootfsPath))). So packaging was building an expensive artifact that most users never consume and that is recreated when actually required.
- The logs and provisioning constants still referenced the old Debian 12 filenames even though the default distro had moved to Debian 13, producing misleading output like Debian-12-Base-In-Provisioning-For-Kubemaster.vhdx.

Fix:

- Removed WSL rootfs generation from packaging. 
- Corrected the stale Debian‑12 → Debian‑13 filenames
- also reverted the method content Convert-VhdxToRootfs from the previous version to use the hardware configurations from the command line while using k2s install with --wsl flag.

### DIFF
--- a/lib/modules/k2s/k2s.node.module/linuxnode/baseimage/provisioning.module.psm1
+++ b/lib/modules/k2s/k2s.node.module/linuxnode/baseimage/provisioning.module.psm1
@@ -16,17 +16,17 @@ $kubenodeBaseFileName = 'Kubenode-Base.vhdx'
 $kubeNodeBaseImagePath = "$kubeBinPath\$kubenodeBaseFileName"
 
 $KubemasterVmProvisioningVmName = 'KUBEMASTER_IN_PROVISIONING'
-$RawBaseImageInProvisioningForKubemasterImageName = 'Debian-12-Base-In-Provisioning-For-Kubemaster.vhdx'
+$RawBaseImageInProvisioningForKubemasterImageName = 'Debian-13-Base-In-Provisioning-For-Kubemaster.vhdx'
 $VmProvisioningNatName = 'KubemasterVmProvisioningNat'
 $VmProvisioningSwitchName = 'KubemasterVmProvisioningSwitch'
 
 $KubeworkerVmProvisioningVmName = 'KUBEWORKER_IN_PROVISIONING'
-$RawBaseImageInProvisioningForKubeworkerImageName = 'Debian-12-Base-In-Provisioning-For-Kubeworker.vhdx'
+$RawBaseImageInProvisioningForKubeworkerImageName = 'Debian-13-Base-In-Provisioning-For-Kubeworker.vhdx'
 $KubeworkerVmProvisioningNatName = 'KubeworkerVmProvisioningNat'
 $KubeworkerVmProvisioningSwitchName = 'KubeworkerVmProvisioningSwitch'
 
 $KubenodeVmProvisioningVmName = 'KUBENODE_IN_PROVISIONING'
-$RawBaseImageInProvisioningForKubenodeImageName = 'Debian-12-Base-In-Provisioning-For-Kubenode.vhdx'
+$RawBaseImageInProvisioningForKubenodeImageName = 'Debian-13-Base-In-Provisioning-For-Kubenode.vhdx'
 $KubenodeVmProvisioningNatName = 'KubenodeVmProvisioningNat'
 $KubenodeVmProvisioningSwitchName = 'KubenodeVmProvisioningSwitch'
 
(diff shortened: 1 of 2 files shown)


krotz-dieter (commented):
Now I have a complete picture. Let me write the review.

---

## Code Review: PR #2704

This PR makes two independent changes: (1) renames image filenames from Debian-12 to Debian-13, and (2) removes WSL rootfs pre-building from the package provisioning step and strips the disk/memory scaling logic from `Convert-VhdxToRootfs`.

---

### 🔴 Bug: Removal of disk/memory scaling in `Convert-VhdxToRootfs` is a regression

**File:** `lib/modules/k2s/k2s.node.module/linuxnode/baseimage/provisioning.module.psm1`

The PR removes the block that computed minimum safe disk/memory sizes for the rootfs creation VM:

```powershell
# Removed:
$sourceVhdxSizeBytes = (Get-Item -Path $SourceVhdxPath).Length
$requiredDiskSize = [uint64]$sourceVhdxSizeBytes * 6 + 20GB
$rootfsVmDiskSize = [System.Math]::Max([uint64]$VMDiskSize, $requiredDiskSize)
$rootfsVmMemory  = [System.Math]::Max([long]$VMMemoryStartupBytes, [long]8GB)
```

And now passes `$VMDiskSize` / `$VMMemoryStartupBytes` directly. The removed comment explicitly warned:

> _"The rootfs creation VM must temporarily hold… the copied source vhdx, the fully extracted (uncompressed) root filesystem and the resulting tarball all at once… Provide more disk than the control plane defaults to avoid scp 'write failure' / 'no space left on device' errors as the produced image grows in size."_

The `Convert-VhdxToRootfs` function is **still actively called** during WSL installation via `new-wsl-vm-as-linuxnode.module.psm1` → `New-WslRootfsForControlPlaneNode` → `Convert-VhdxToRootfs`. The values passed through (`$VMDiskSize`, `$VMMemoryStartupBytes`) are the user's configured VM parameters, which may well be below the required threshold. With a growing Debian image this is increasingly likely to cause disk exhaustion.

If this removal is intentional because callers are now expected to pass large enough values, the callers in `new-wsl-vm-as-linuxnode.module.psm1` do **not** enforce any minimum — they pass through whatever the user configured:

```powershell
$wslRootfsForControlPlaneNodeCreationParams = @{
    VMDiskSize           = $VMDiskSize        # user's value, no floor
    VMMemoryStartupBytes = $VMMemoryStartupBytes  # user's value, no floor
    ...
}
```

The scaling logic should be retained in `Convert-VhdxToRootfs`, or at minimum moved into `New-WslRootfsForControlPlaneNode` so all callers are protected.

---

### ⚠️ Potential offline-install regression: WSL rootfs no longer pre-bundled in package

**File:** `lib/scripts/k2s/system/package/New-K2sPackage.Provisioning.ps1`

The PR removes the call to `New-WslRootfsForControlPlaneNode` inside `New-ProvisionedKubemasterBaseImage`. Previously, the rootfs was pre-built and included in the offline package. After this change, the rootfs is created on-demand during WSL installation — which requires the kubemaster VHDX to be available and sufficient disk/memory at install time.

For **offline installations**, `new-wsl-vm-as-linuxnode.module.psm1` handles this (`if (!(Test-Path -Path $rootfsPath))` it builds the rootfs from the VHDX). This may work if the VHDX is present in the package. However, the rootfs file deletion code is also removed:

```powershell
# Removed: pre-creation cleanup of stale rootfs
if (Test-Path -Path $rootfsPath) {
    Remove-Item -Path $rootfsPath -Force
}
```

If a stale rootfs already exists in the output directory during package build, it will no longer be cleaned up. This is benign since rootfs creation is now skipped entirely in the package step — but worth confirming the downstream package contents are still correct for WSL-based deployment scenarios.

---

### ℹ️ Note: Debian-12 → Debian-13 renames appear consistent within changed files

The four filename renames (in-provisioning images and provisioned kubenode image) are consistent with each other. `cfg/config.json` and `New-K2sPackage.ps1` already have both `Debian-12` and `Debian-13` entries (supporting both flavors) and are not affected by this PR.

---

## Files Reviewed
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/baseimage/provisioning.module.psm1
- ✅ lib/scripts/k2s/system/package/New-K2sPackage.Provisioning.ps1

PR #2734, opened 2026-07-08
Headlamp improvements

vvkrishnakanth9:
Fixes #TODO
Adding keda and kyverno plugins.

### Motivation

Restore offline Headlamp plugin support.

### Modifications

Implemented TGZ-based plugin architecture.

### Verification

Unit tests, E2E, manual validation.


krotz-dieter (commented):

Here is my review:

---

## Review: PR #2734 — Headlamp improvements

This PR adds KEDA and Kyverno Headlamp plugins with conditional capability detection and auto-sync. The overall design is solid and consistent with existing patterns. I found one functional gap worth addressing.

---

### Bug: KEDA CRD fallback defeats `Test-NamespaceActive` protection on disable

**File**: `addons/dashboard/dashboard.module.psm1` — `Test-KedaCapabilityAvailable`  
**File**: `addons/autoscaling/Disable.ps1`

`Test-KedaCapabilityAvailable` checks the `autoscaling` namespace via `Test-NamespaceActive` first, then falls back to checking the `scaledobjects.keda.sh` CRD. `Sync-HeadlampPlugins` is called immediately after `kubectl delete -f $kedaManifest` in `Disable.ps1`.

At that moment:
- `autoscaling` namespace → `Terminating` → `Test-NamespaceActive` returns `$false` ✅ (Terminating guard works)
- `scaledobjects.keda.sh` CRD → **still present** (async garbage-collection) → function returns `$true` ❌

The CRD fallback overrides the Terminating-namespace protection, so the KEDA plugin init-container is **not removed** during the first sync after disabling the autoscaling addon.

The PR itself demonstrates awareness of this class of bug — the new regression test `'cert-manager namespace is Terminating and CRD already gone'` guards against exactly this pattern, but only for cert-manager's namespace path. The equivalent scenario for KEDA ("namespace Terminating, CRD still present") is not tested and the code path is not protected.

A consistent fix would add a `Test-NamespaceActive -Name 'keda'` or a CRD-Terminating guard (or at minimum a regression test that asserts `$false` when the namespace is Terminating and the CRD is still present, to document the known gap):

```powershell
# In Test-KedaCapabilityAvailable: note the CRD check still races deletion;
# the namespace check (Terminating guard) is the primary deactivation signal.
```

---

### Observation: Kyverno CRD-only detection has no Terminating guard

**File**: `addons/dashboard/dashboard.module.psm1` — `Test-KyvernoCapabilityAvailable`

Kyverno detection is CRD-only (no namespace check), so there is no `Test-NamespaceActive` guard at all. CRD deletion can be slow when Kyverno has many policies (finalizer-driven). The same race condition applies when the security addon is disabled. This is less severe because Kyverno's CRD deletion tends to be faster in practice, but it's worth noting for consistency.

---

### Everything else looks correct

- `autoscaling` namespace name confirmed in `manifests/keda.yaml` (line 12: `name: autoscaling`) — `Test-KedaCapabilityAvailable`'s namespace check is accurate.
- `security/Disable.ps1` already calls `Sync-HeadlampPlugins`, so Kyverno plugin lifecycle is covered when the security addon is toggled. ✅
- `Test-NamespaceActive` is correctly not exported (private helper). ✅
- The `$script:` counter pattern in `BeforeEach`/`BeforeAll` mocks mirrors established patterns in this test file and is correct. ✅
- Lock file count update (3→5), test count updates (3→5), and `Export-ModuleMember` additions are all consistent. ✅
- The `Wait-ForHeadlampAvailable` timeout increase (200s→300s) with its justification is reasonable. ✅

---

## Files Reviewed
- ✅ addons/autoscaling/Disable.ps1
- ✅ addons/autoscaling/Enable.ps1
- ✅ addons/dashboard/README.md
- ✅ addons/dashboard/addon.manifest.yaml
- ✅ addons/dashboard/build/headlamp-plugin-images.unit.tests.ps1
- ✅ addons/dashboard/build/headlamp-plugins.lock.json
- ✅ addons/dashboard/build/plugins/keda-0.1.2.tar.gz
- ✅ addons/dashboard/build/plugins/keda-0.1.2.tar.gz.license
- ✅ addons/dashboard/build/plugins/kyverno-0.1.0.tar.gz
- ✅ addons/dashboard/build/plugins/kyverno-0.1.0.tar.gz.license
- ✅ addons/dashboard/dashboard.module.psm1
- ✅ addons/dashboard/dashboard.module.unit.tests.ps1

krotz-dieter (commented):
The base `k8s.go` already has `"strings"` in its imports, so the new `strings.Contains` usage is fine. Now I have enough to write the review.

---

## Code Review: PR #2734 — Headlamp improvements

### 🐛 Bug: Stale prometheus plugin version in e2e test

**File:** `k2s/test/e2e/addons/dashboard/05-plugins/dashboard_plugins_test.go` (~line 78)

```go
{
    description:   "Prometheus plugin via the monitoring addon",
    enableArgs:    []string{"addons", "enable", "monitoring", "-o"},
    disableArgs:   []string{"addons", "disable", "monitoring", "-o"},
    initContainer: "prometheus-plugin",
    image:         "headlamp-plugin-prometheus:0.8.2",   // ← stale, should be 0.9.0
},
```

This PR upgrades the Prometheus Headlamp plugin from `0.8.2` → `0.9.0` everywhere (lock file, module registry, manifest), but the new e2e test suite still references the old version. `ExpectDeploymentInitContainer` performs a `strings.Contains(ic.Image, expectedImage)` check, so `"0.8.2"` won't be found in the deployed `0.9.0`-tagged image and this test will always fail in CI.

**Fix:** Change `"headlamp-plugin-prometheus:0.8.2"` → `"headlamp-plugin-prometheus:0.9.0"` to match the lock file entry and the module registry.

---

### Minor observation (not a blocker)

In `Test-NamespaceActive`, the kubectl jsonpath output is compared with `-eq 'Active'`. If `Invoke-Kubectl` doesn't trim its `.Output` field and kubectl emits a trailing newline, `"Active\n" -eq "Active"` is `$false`. This would cause silent false-negatives in the capability detectors (a namespace that's truly Active would be treated as absent). The unit tests mock `Output` with a trimmed `'Active'` string so the tests pass regardless. It's worth auditing `Invoke-Kubectl`'s output handling if this hasn't already been verified.

---

## Files Reviewed
- ✅ addons/autoscaling/Disable.ps1
- ✅ addons/autoscaling/Enable.ps1
- ✅ addons/dashboard/README.md
- ✅ addons/dashboard/addon.manifest.yaml
- ✅ addons/dashboard/build/headlamp-plugin-images.unit.tests.ps1
- ✅ addons/dashboard/build/headlamp-plugins.lock.json
- ✅ addons/dashboard/build/plugins/keda-0.1.2.tar.gz (binary, not reviewed)
- ✅ addons/dashboard/build/plugins/keda-0.1.2.tar.gz.license
- ✅ addons/dashboard/build/plugins/kyverno-0.1.0.tar.gz (binary, not reviewed)
- ✅ addons/dashboard/build/plugins/kyverno-0.1.0.tar.gz.license
- ✅ addons/dashboard/build/plugins/prometheus-0.9.0.tar.gz (binary, not reviewed)
- ✅ addons/dashboard/build/plugins/prometheus-0.9.0.tar.gz.license
- ✅ addons/dashboard/dashboard.module.psm1
- ✅ addons/dashboard/dashboard.module.unit.tests.ps1
- ✅ k2s/test/e2e/addons/dashboard/01-enabledisable/dashboard_test.go
- ✅ k2s/test/e2e/addons/dashboard/02-securityenhanced/dashboard_security_test.go
- ✅ k2s/test/e2e/addons/dashboard/05-plugins/dashboard_plugins_test.go

PR #2771, opened 2026-07-13
2586 traefik log errors fix

vvkrishnakanth9:
Fixes resolving ingress tls errors in traefik logs.

### Motivation

Remove redundant Traefik TLS configuration.

### Modifications

Removed redundant TLS blocks; added E2E regression tests.

### Verification

Verified manually; all E2E tests passed.


krotz-dieter (commented):
I now have enough information to write the review. Let me verify the `VerifyDeploymentReachableFromHostWithStatusCode` signature I found:

```go
func VerifyDeploymentReachableFromHostWithStatusCode(ctx context.Context, expectedStatusCode int, url string, headers ...map[string]string)
```

The new calls omit `headers`, which is valid since it's variadic. Let me check one more detail about the test ordering in the PR.

Looking at the diff for the "traefik first" scenario, the new `It` block is inserted correctly before "Deactivates all the addons", so it runs while both addons are still active.

---

## Review

### YAML changes — `traefik-ingress-keycloak.yaml` & `traefik-ingress-oauth2-proxy.yaml`

The removed `spec.tls` blocks referenced `secretName: k2s-cluster-local-tls` in the `security` namespace. That secret only lives in the `ingress-traefik` namespace, so Traefik was logging `secret security/k2s-cluster-local-tls does not exist` on every reconciliation. Removing these blocks is the correct fix: TLS for `k2s.cluster.local` is already owned by the central `ingress-traefik/traefik-cluster-local` ingress and Traefik will serve its certificate for all routes on that host regardless of which namespace the route-defining ingress lives in. No functional regression.

### Test file — `ingress_traefik_security_test.go`

**`VerifyDeploymentReachableFromHostWithStatusCode` call signature** — verified against `main`. The function has a variadic `headers ...map[string]string` parameter, so the new calls without headers compile correctly.

**New HTTPS connectivity tests** — Both new `It` blocks verify the fix end-to-end by confirming the K2s CA-trusted certificate is still served after the `spec.tls` removal. The placement (after the log-check, before deactivation) is correct — addons are still running at that point in both scenarios.

**One issue worth noting:**

In the second scenario's new `It` block (`"emits no redundant TLS-secret error..."`, line ~226), the log check uses `--since=10m`. If the security addon has been running for longer than 10 minutes before this `It` block executes (possible given the `30 * time.Second` sleep plus the albums connectivity test), the window could miss the original TLS-secret errors logged right after addon activation. The existing test in the "security first" scenario has the same potential blind spot, but it was already there before this PR. For a regression guard this is an accepted trade-off, but it's worth noting that `--since` needs to span at least the reconciliation interval after the ingress objects are created.

No bugs, security vulnerabilities, or breaking changes found. The fix is correct and the added tests are appropriate regression guards.

## Files Reviewed
- ✅ `addons/security/manifests/keycloak/traefik-ingress-keycloak.yaml`
- ✅ `addons/security/manifests/keycloak/traefik-ingress-oauth2-proxy.yaml`
- ✅ `k2s/test/e2e/addons/ingress/traefik/securityenhanced/ingress_traefik_security_test.go`

krotz-dieter (commented):
Now I have enough context for a complete review. Let me verify one more thing:

## Review

### YAML Changes (traefik-ingress-keycloak.yaml & traefik-ingress-oauth2-proxy.yaml)

The removal of `spec.tls` blocks is correct. The secret `k2s-cluster-local-tls` only exists in `ingress-traefik` namespace, not in `security`. Traefik was looking for it as `security/k2s-cluster-local-tls` on every reconcile and logging errors. TLS termination is correctly delegated to the central `ingress-traefik/traefik-cluster-local` ingress, and the `traefik.ingress.kubernetes.io/router.tls: 'true'` annotation is correctly retained.

### Test Changes (ingress_traefik_security_test.go)

**Issue: Missing `"service not found"` assertion in the "traefik first" scenario**

The new `It` block added to the second `Describe` block ("traefik first"):

```go
// k2s/test/e2e/addons/ingress/traefik/securityenhanced/ingress_traefik_security_test.go ~line 233
logs := suite.Kubectl().MustExec(ctx, "logs", "-n", "ingress-traefik", "deploy/traefik", "--since=1h")
Expect(logs).NotTo(ContainSubstring("secret security/k2s-cluster-local-tls does not exist"))
```

This only guards against the TLS secret error but omits the `"service not found"` check that exists in the first `Describe` block (security-first scenario). If a service-not-found regression occurs specifically in the traefik-first activation order, it goes undetected. For consistency and completeness:

```go
Expect(logs).NotTo(ContainSubstring("service not found"))
Expect(logs).NotTo(ContainSubstring("secret security/k2s-cluster-local-tls does not exist"))
```

**No issue** with the `--since=1h` window: the suite disables all addons in `BeforeSuite`, so the traefik pod is always fresh when the test runs, and `1h` is well above the 20m per-step timeout.

**No issue** with calling `VerifyDeploymentReachableFromHostWithStatusCode` without `headers`: the function signature is `func VerifyDeploymentReachableFromHostWithStatusCode(ctx context.Context, expectedStatusCode int, url string, headers ...map[string]string)` — variadic, so omitting `headers` is valid.

---

## Files Reviewed
- ✅ addons/security/manifests/keycloak/traefik-ingress-keycloak.yaml
- ✅ addons/security/manifests/keycloak/traefik-ingress-oauth2-proxy.yaml
- ✅ k2s/test/e2e/addons/ingress/traefik/securityenhanced/ingress_traefik_security_test.go

PR #2823, opened 2026-07-23
Add functions for Kubernetes interaction and Windows node route manag…

GermanGLLopez:
Fix for unstable test addon-ingress-nginx

### DIFF
--- a/lib/scripts/k2s/system/startup/Start-System.ps1
+++ b/lib/scripts/k2s/system/startup/Start-System.ps1
@@ -115,10 +115,18 @@ function Invoke-KubectlWithKubeConfig {
     )
 
     $kubeBinPathTools = Get-KubeToolsPath
-    $ErrorActionPreference = 'Continue'
-    $output = &"$kubeBinPathTools\kubectl.exe" --kubeconfig="$KubeConfigPath" @Params 2>&1
-    $exitCode = $LASTEXITCODE
-    $ErrorActionPreference = 'Stop'
+    $savedErrorActionPreference = $ErrorActionPreference
+    $output = $null
+    $exitCode = 1
+
+    try {
+        $ErrorActionPreference = 'Continue'
+        $output = &"$kubeBinPathTools\kubectl.exe" --kubeconfig="$KubeConfigPath" @Params 2>&1
+        $exitCode = $LASTEXITCODE
+    }
+    finally {
+        $ErrorActionPreference = $savedErrorActionPreference
+    }
 
     return [pscustomobject]@{
         Success  = ($exitCode -eq 0)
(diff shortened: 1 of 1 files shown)


krotz-dieter (commented):
I've now confirmed the return type of `Invoke-CmdOnControlPlaneViaSSHKey` — it returns `[pscustomobject]@{ Success = $success; Output = $output }`, so the `.Success` and `.Output` accesses are correct. Let me now write up the findings.

---

## PR Review: Add functions for Kubernetes interaction and Windows node route management

### Bug 1 — `Invoke-KubectlWithKubeConfig` permanently modifies `$ErrorActionPreference` in callers' scope

**File:** `lib/scripts/k2s/system/startup/Start-System.ps1` (new `Invoke-KubectlWithKubeConfig`)

```powershell
$ErrorActionPreference = 'Continue'
$output = &"$kubeBinPathTools\kubectl.exe" ...
$exitCode = $LASTEXITCODE
$ErrorActionPreference = 'Stop'
```

The function unconditionally sets `$ErrorActionPreference = 'Stop'` on exit without saving and restoring the caller's value. Any caller that had a different preference (e.g. `'Continue'` or `'SilentlyContinue'`) will silently have it overwritten. This can cause unhandled terminating errors in `Restart-FlannelDaemonSetWithWindowsRouteRepair` if an unexpected error occurs after the call.

**Fix:** Save and restore the preference:
```powershell
$savedPreference = $ErrorActionPreference
$ErrorActionPreference = 'Continue'
# ...
$ErrorActionPreference = $savedPreference
```

---

### Bug 2 — `Get-WindowsWorkerNodeRoutes` kubectl failure is indistinguishable from "no Windows nodes"

**File:** `lib/scripts/k2s/system/startup/Start-System.ps1` (`Get-WindowsWorkerNodeRoutes` and `Restart-FlannelDaemonSetWithWindowsRouteRepair`)

`Get-WindowsWorkerNodeRoutes` returns an empty `ArrayList` on **both** a genuine absence of Windows nodes and a kubectl query failure (network issue, API server not ready, etc.). The caller treats both identically:

```powershell
if ($windowsNodeRoutes.Count -eq 0) {
    Write-Log "[$logUseCase] No Windows worker nodes found, restarting flannel daemonset once"
    # ... restart and return immediately
    return
}
```

When `Get-WindowsWorkerNodeRoutes` fails due to a transient kubectl error, the function restarts flannel once and exits on the **first attempt**, without route validation and without retrying. This could leave Windows pod routes broken if the query failure occurred because kubectl wasn't ready yet.

**Fix:** Have `Get-WindowsWorkerNodeRoutes` return a distinguished error state (e.g. a second return value or a `$null` vs empty list convention), and let the caller distinguish query failure from "no Windows nodes found."

---

### Bug 3 — Primary `spec.podCIDR` not validated as IPv4 before SSH command interpolation

**File:** `lib/scripts/k2s/system/startup/Start-System.ps1` (`Get-WindowsWorkerNodeRoutes` / `Get-MissingWindowsPodRoutesOnControlPlane`)

The IPv4-CIDR regex filter is applied **only** when falling back to `spec.podCIDRs`:
```powershell
$podCIDR = $node.spec.podCIDR   # ← No validation here
if ([string]::IsNullOrWhiteSpace($podCIDR) -and $node.spec.podCIDRs) {
    foreach ($candidateCIDR in $node.spec.podCIDRs) {
        if ($candidateCIDR -match '^\d{1,3}(\.\d{1,3}){3}/\d{1,2}$') { ... }
    }
}
```

The raw `spec.podCIDR` is then passed directly into an SSH command string:
```powershell
Invoke-CmdOnControlPlaneViaSSHKey -CmdToExecute "ip -4 route show $($nodeRoute.PodCIDR)"
```

In dual-stack clusters where IPv6 is the primary family, `spec.podCIDR` could be an IPv6 CIDR, causing the `ip -4 route show` command to silently produce no output and incorrectly flag the route as missing on every attempt. Apply the same IPv4 regex to the primary `spec.podCIDR` field.

---

## Files Reviewed

PR #2825, opened 2026-07-23
Fix addon logging test

GermanGLLopez:

### DIFF
--- a/k2s/test/e2e/addons/logging/logging_test.go
+++ b/k2s/test/e2e/addons/logging/logging_test.go
@@ -25,7 +25,11 @@ import (
 	"github.com/onsi/gomega/gstruct"
 )
 
-const testClusterTimeout = time.Minute * 20
+const (
+	testClusterTimeout        = time.Minute * 20
+	statusCheckRetryTimeout   = 3 * time.Minute
+	statusCheckRetryPolling   = 10 * time.Second
+)
 
 var (
 	suite                 *framework.K2sTestSuite
@@ -322,9 +326,9 @@ func expectLoggingResourcesRemoved(ctx context.Context) {
 }
 
 func expectStatusToBePrinted(ctx context.Context) {
-	output := suite.K2sCli().MustExec(ctx, "addons", "status", "logging")
-
-	Expect(output).To(SatisfyAll(
+	Eventually(func() string {
+		return suite.K2sCli().MustExec(ctx, "addons", "status", "logging")
+	}).WithTimeout(statusCheckRetryTimeout).WithPolling(statusCheckRetryPolling).Should(SatisfyAll(
 		MatchRegexp("ADDON STATUS"),
 		MatchRegexp(`Addon .+logging.+ is .+enabled.+`),
 		MatchRegexp("Opensearch dashboards are working"),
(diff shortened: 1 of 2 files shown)


krotz-dieter (commented):
Now I have enough context to write a complete review.

---

## PR Review: Fix addon logging test (#2825)

The PR makes E2E tests more resilient by wrapping addon enable/disable calls and status checks in `Eventually` retry loops. The approach is sound overall, but there is one confirmed bug and a couple of issues worth noting.

---

### 🐛 Bug: `collectCoreWorkloadRolloutDiagnostics` passes a potentially-cancelled context

**File:** `k2s/test/e2e/cluster/core/core_test.go` — new `collectCoreWorkloadRolloutDiagnostics` function (lines 237–253)

```go
func collectCoreWorkloadRolloutDiagnostics(ctx context.Context) {
    GinkgoWriter.Println("Collecting core workload rollout diagnostics..")
    runDiagnosticCommand(ctx, "get", "deployment", ...)
    ...
}
```

This function is called **only when the rollout deadline has been exceeded** (`time.Now().After(deadline)` at line 226). By that point, the `ctx` passed from `BeforeSuite` may already be cancelled or near expiry, causing all the `kubectl` diagnostic commands to fail immediately and silently (they only log non-zero exit codes). The diagnostics will be useless exactly when they're most needed.

The existing `collectWindowsWorkloadNetworkDiagnostics()` correctly handles this by creating a **fresh context**:

```go
func collectWindowsWorkloadNetworkDiagnostics() {
    diagCtx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
    defer cancel()
    runDiagnosticCommand(diagCtx, ...)
}
```

`collectCoreWorkloadRolloutDiagnostics` should do the same — accept no context parameter (or create a fresh one internally) rather than forwarding the potentially-cancelled `ctx`.

---

### ⚠️ Issue: `waitForCoreWorkloadRollout` loop ignores context cancellation

**File:** `k2s/test/e2e/cluster/core/core_test.go` — `waitForCoreWorkloadRollout` (line 174)

The polling loop only checks `time.Now().After(deadline)`. If `ctx` is cancelled before the function's own `deadline`, `Get()` calls start returning context errors, each deployment is added to `pending`, and the loop keeps sleeping and retrying (up to the 14-minute deadline) — burning time without making progress and producing confusing error messages that mask the real cause (context cancellation).

The loop should also select on `ctx.Done()`:

```go
select {
case <-ctx.Done():
    Fail(fmt.Sprintf("context cancelled while waiting for rollout: %v", ctx.Err()))
default:
}
time.Sleep(pollInterval)
```

---

### ℹ️ Minor: Negative assertions in `expectOmitOpensearchStatusToBePrinted` are checked in a separate `Eventually`

**File:** `k2s/test/e2e/addons/logging/logging_test.go` — `expectOmitOpensearchStatusToBePrinted`

The positive assertions (containing certain output) and the negative assertions (`NotTo(ContainSubstring(...))`) are now in two separate `Eventually` blocks. They query `addons status logging` at different times and could theoretically observe different states. This is low-risk in practice for a stable addon, but for completeness both checks could be combined into a single `Eventually(func(g Gomega) { ... })` to guarantee they're evaluated on the same output.

---

## Files Reviewed
- ✅ k2s/test/e2e/addons/logging/logging_test.go
- ✅ k2s/test/e2e/cluster/core/core_test.go

PR #2830, opened 2026-07-24
#2829 improve with retries the internet connectivity check

### LINKED ISSUE #2829: [Bug]: Test failures in K2s v2.0.0
### Description
Validation testing for K2s **v2.0.0** encountered test failures that need to be resolved.

### Environment
- **K2s Version:** `v2.0.0`
- **Host OS:** Windows 11
- **Test Suite:** `[E2E / Integration / Unit]`

### Failing Tests & Error Logs
- **`TestName1`**: `[Brief failure summary or log snippet]`
- **`TestName2`**: `[Brief failure summary or log snippet]`

### Steps to Reproduce
1. Checkout version `v2.0.0`.
2. Run test suite command:
   ```powershell
   # [Insert test execution command]

krotz-dieter:
#2829 improve with retries the internet connectivity check


krotz-dieter (commented):
Here is my review of PR #2830:

## Review: `k2s/test/framework/k8s/k8s.go`

The overall approach is sound — wrapping the single-shot exec in an `Eventually` loop adds real resilience for flaky network conditions. `ctx` is correctly in scope, `executeCommandInPod` exists and returns `error`, and the buffer resets before each poll are correct.

### ⚠️ Timeout Mismatch Between `Eventually` and curl

**File:** `k2s/test/framework/k8s/k8s.go`

`internetReachabilityCheckTimeout` is **3 minutes**, but with `-m 60` and `--retry 5`, a single curl invocation can block for up to **360 seconds** in older curl versions (where `--max-time` applies *per attempt*, not to the total operation).

When `Eventually` times out it does **not** cancel the in-flight `executeCommandInPod` call (which uses `param.Ctx`, not an Eventually-internal context), so the function may still be blocked at the 180-second mark. Gomega will declare the assertion failed, but the underlying stream won't be torn down until `param.Ctx` itself expires.

**Concrete risk:** On a host with an older curl build (pre-7.71.0) and repeated network failures, all 5 retries × 60s each = 360s > 180s `Eventually` timeout, meaning the test will always time out on the very first poll cycle, making the retry logic ineffective.

**Suggested fix:** Either:
1. Add `--retry-max-time 60` to cap the total retry duration to 60 seconds (making behavior version-independent), or
2. Increase `internetReachabilityCheckTimeout` to at least 7–8 minutes to accommodate worst-case curl retry duration, or
3. Remove `--retry` from the curl flags entirely and let `Eventually` handle all retries (which already does so at a higher level with proper timeout control).

Option 3 is simplest — the `Eventually` loop already provides the retry logic, so the built-in curl `--retry` is redundant layering with unpredictable total duration.

---

No other functional bugs found. The constants, buffer resets, HTTP status check logic, and `ctx` passing are all correct.

## Files Reviewed

PR #2832, opened 2026-07-24
#1149 reverted k2s setup: improvements for linux only

### LINKED ISSUE #1149: Reverted default K2s setup: Linux Host and Windows VM
Reverted default K2s setup: Linux Host and Windows VM

Currently the default setup for K2s is depicted in the diagram:
[Default Setup](https://github.com/Siemens-Healthineers/K2s/blob/main/docs/assets/overview.png)

For certain use cases an additional setup would be required.
This additional setup would be using the main host as an Linux System and the VM behind would be an optional worker node using the Windows VM.

krotz-dieter:
#1149 reverted k2s setup: improvements for linux only


krotz-dieter (commented):
Now I have enough context to write the full review.

---

## Code Review: PR #2832 — Linux-only K2s install improvements

Overall the PR is well-structured. There are a few confirmed bugs and one security-related concern worth addressing.

---

### 🐛 Bug: `exeName` mismatch on Linux (`install.go`)

**File:** `k2s/cmd/k2s/cmd/install/install.go` (~line 188)

```go
exeName := "k2s.exe"
if runtime.GOOS == "linux" {
    exeName = "k2s"
}
```

The binary is built as **`k2s.linux`** (`build.sh` and `Makefile` were both updated), but the duplicate-installation guard (`checkForOldK2sExecutables`) scans PATH for `"k2s"`. On a system where only `k2s.linux` exists, this check will never find any duplicates. Conversely, it could trip on an unrelated binary named `k2s` (e.g. a distro-packaged tool). The name should be `"k2s.linux"` to match the actual binary.

---

### 🐛 Bug: `%` not escaped in systemd unit file (`provisioning_linux.go`)

**File:** `k2s/internal/setuporchestration/provisioning_linux.go` — `installHTTPProxy` / `systemdValue`

```go
func systemdValue(value string) string {
    return strings.ReplaceAll(strings.ReplaceAll(value, "\\", "\\\\"), "\"", "\\\"")
}
```

systemd treats `%` as a specifier prefix in unit file directives (including `Environment=` and `ExecStart=`). `systemdValue` escapes `\` and `"` but not `%`. A proxy URL with a percent-encoded credential (e.g. `http://user:p%40ss@proxy:3128`) or a no-proxy entry with a `%` would cause the service to fail to start with a cryptic "unknown specifier" error. Fix: add `strings.ReplaceAll(value, "%", "%%")` as the first replacement.

---

### 🐛 Bug: `ExecStart` is not properly quoted (`provisioning_linux.go`)

**File:** `k2s/internal/setuporchestration/provisioning_linux.go` — `installHTTPProxy`

```go
unit := fmt.Sprintf(`...
ExecStart=%s %s
...`, proxyBinary, strings.Join(args, " "))
```

If `proxyBinary` (derived from `cfg.InstallDir`) or the proxy URL (via `--forwardproxy`) contains a space, this `ExecStart` line will be parsed incorrectly by systemd. systemd does not use shell expansion here; spaces delimit argument boundaries. The binary path should at minimum be quoted: `ExecStart="%s" %s`. Alternatively, build the line by writing individual `ExecStart` arguments on separate lines using systemd's multi-line syntax, or ensure spaces in paths are properly escaped.

---

### ⚠️ Silent dependency download failures (`download-k8s-packages.sh`)

**File:** `cfg/nodeextension/debian13/scripts/download-k8s-packages.sh` (~line 170)

```bash
| xargs -r sudo apt-get download 2>/dev/null || true
```

Removing `--allow-unauthenticated` is correct for security, but the `|| true` means any authentication or network failure during **dependency** downloads is silently swallowed. The primary package download now fails fast (no `|| true`), but its dependencies can silently be missing. This leads to a package set that appears complete but fails only at install time with an opaque dpkg error. At minimum, log a warning on non-zero exit from the dependency download pipeline.

---

### ℹ️ Minor: dead constant `winVMName` (`orchestrator_linux.go`)

**File:** `k2s/internal/setuporchestration/orchestrator_linux.go`

The constant `winVMName = "k2s-win-worker"` is defined but is no longer referenced anywhere after the Windows VM code was removed. Go does not error on unused constants, but it is dead code. Remove it to keep the file clean.

---

### ℹ️ Minor: `lookupUserID` name is misleading (`provisioning_linux.go`)

**File:** `k2s/internal/setuporchestration/provisioning_linux.go`

```go
func lookupUserID(value string) (int, error) {
    id, err := strconv.Atoi(value)
    ...
}
```

The function performs a string-to-int conversion, not a user lookup. The name `parseID` or `mustParseInt` would be clearer.

---

## Files Reviewed

- ✅ Makefile
- ✅ build.sh
- ✅ cfg/nodeextension/debian13/scripts/download-k8s-packages.sh
- ✅ docs/dev-guide/contributing/building-locally.md
- ✅ docs/op-manual/installing-k2s.md
- ✅ k2s/cmd/k2s/cmd/install/install.go
- ✅ k2s/cmd/k2s/cmd/status/cmd.go
- ✅ k2s/cmd/k2s/cmd/uninstall/uninstall.go
- ✅ k2s/internal/provider/cluster_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux.go
- ✅ k2s/internal/setuporchestration/provisioning_linux.go

PR #2840, opened 2026-07-28
2805 Add node-targeted gpu-node enablement with --node support

### LINKED ISSUE #2805: GPU-Node Addon: Allow specifying nodes
**Is your feature request related to a problem? Please describe.**
We want to use GPU on a dedicate linux node attached to the cluster, but NOT on the kubemaster node, as it would conflict with other, bare-metal, GPU consumers.

Currently, the GPU addon installs workload to any Linux node - and thus puts nvidia related docker images into the kubemaster cache, even though they will never be used. As disk space is valuable, we would want to avoid this.

**Describe the solution you'd like**
When enabling the gpu-addon, we would want to specify (optionally) a list of nodes for which it should be enabled. We would want to enable it only for one single node, especially NOT for kubemaster.
If given, no workload and no nvidia related docker images should not be deployed into kubemaster. 

**Describe alternatives you've considered**
We might remove the images from the cache after enabling the gpu-addon as a workaround.

**Additional context**
Using K2s 1.8.1

sk1486:
Description
This change adds node-targeted enablement for the gpu-node addon so GPU support can be applied to one or more specific nodes instead of always configuring the KubeMaster. When a node list is provided, K2s labels those nodes for GPU scheduling, configures GPU runtime support on the selected targets, and skips KubeMaster setup unless it is explicitly included.

The update also documents how multi-node targeting works, including comma-separated node lists and the PowerShell quoting requirement when spaces are used. The addon documentation now makes the lifecycle clearer for external GPU workers: GPU configuration is detected automatically when a node is added, scheduling is driven by node labels, and disable remains a global operation that removes the device plugin cluster-wide without taking a node filter.

What changed

- Added support for enabling gpu-node with a node list via --node.
- Allowed multiple nodes to be targeted in one command.
- Preserved KubeMaster-skipping behavior when it is not included in the node list.
- Clarified external GPU worker behavior and scheduling model in the docs.
- Documented that disable is cluster-wide and does not accept --node.

k2s addons enable gpu-node --node "gpu-worker-a, gpu-worker-b"

 Impact

- GPU workloads can now be enabled on selected worker nodes without affecting every node in the cluster.
- Node-level GPU health matters more clearly: a broken node can fail GPU pods even when another targeted node is healthy.
- Existing single-node enablement remains unchanged.


krotz-dieter (commented):
Now I have enough context to write a thorough review. Let me compile the findings:

Here is my full review:

---

## Code Review: PR #2840 — GPU node-targeted enablement with `--node` support

### 🐛 Bug: Dead code in `validateAddonCommandArgs` — helpful error message is never shown

**`k2s/cmd/k2s/cmd/addons/generic/generic.go`, lines 330–336**

The trailing-comma hint (the message about whitespace splitting) is **unreachable**:

```go
// Line 326: if --node is non-empty, we return nil immediately
if nodeFlag != nil && strings.TrimSpace(nodeFlag.Value.String()) != "" {
    return nil   // ← exits here for "node1," case
}

// Lines 331-335: only reached when nodeFlag == nil OR nodeVal == ""
if nodeFlag != nil {
    nodeVal := strings.TrimSpace(nodeFlag.Value.String())  // always "" here
    if strings.HasSuffix(nodeVal, ",") {                   // always false
        return fmt.Errorf("... It looks like the --node value was split...")
    }
}
```

Whenever `--node` has *any* non-empty value (including `"node1,"` with a trailing comma), execution returns `nil` at line 326. By the time the `strings.HasSuffix(nodeVal, ",")` check is evaluated, `nodeVal` is guaranteed to be `""`, making the condition permanently false. The user-friendly diagnostic message can never be emitted. The test at line ~172 of `generic_test.go` that covers this (`"accepts extra args when --node is set"`) only confirms the `nil` return, not the specific error path.

---

### 🐛 Bug: `EscapeWithSingleQuotes` wraps but doesn't escape internal single quotes

**`k2s/cmd/k2s/cmd/addons/generic/generic.go`, line 312** + **`k2s/cmd/k2s/utils/utilities.go`**

```go
// utilities.go
func EscapeWithSingleQuotes(str string) string {
    return "'" + str + "'"   // no escaping of internal ' characters
}
```

The PR now applies this function to **all** string-typed CLI flags across **all** addons (not just `--node`). If any string flag value contains a single quote, the generated PowerShell invocation will be syntactically invalid. In PowerShell, a literal `'` inside a single-quoted string must be doubled (`''`). Example: a backup path like `C:\User's\data` would produce `-BackupPath 'C:\User's\data'` — a broken command.

For Kubernetes node names this is benign (RFC 1123 forbids `'`), but other addon string parameters (file paths, registry URLs, etc.) are exposed to this risk. This is a pre-existing bug in the helper, but the PR is the first to apply it universally to all string flags, widening the blast radius.

---

### 🐛 Bug: `<DESCRIPTION>` instead of `.DESCRIPTION` in PowerShell comment-based help

**`addons/gpu-node/Enable.ps1`, line 11**

```powershell
<#
.SYNOPSIS
Enables GPU support for KubeMaster or selected Linux worker nodes.

<DESCRIPTION>               ← wrong tag
The "gpu-node" addon enables GPU scheduling...
#>
```

The correct PowerShell keyword for comment-based help is `.DESCRIPTION`. The `<DESCRIPTION>` tag is not recognised by PowerShell's help system; `Get-Help .\Enable.ps1` will show no description text. The original file used `.DESCRIPTION` correctly.

---

### ⚠️ Design concern: positional args silently merged into `--node` without user confirmation

**`k2s/cmd/k2s/cmd/addons/generic/generic.go`, lines 341–373**

`mergeExtraNodeArgs` merges *any* positional argument into the `--node` list as long as `--node` is already non-empty. A user typo such as:

```
k2s addons enable gpu-node --node worker1 --tiime-slices 4
```

…would cause `--tiime-slices` and `4` to be silently appended to the node list, producing `node=worker1,--tiime-slices,4`. The `slog.Warn` is emitted but is not surfaced in the user-facing terminal output, so the user would only notice this when the enable operation fails with a "node not found" error rather than a clear usage error.

---

### 📄 Docs: example in `addon.manifest.yaml` comment is singular but targets three nodes

**`addons/gpu-node/addon.manifest.yaml`, line ~57**

```yaml
- cmd: k2s addons enable gpu-node --node my-gpu-worker, my-other-gpu-worker , kubemaster
  comment: Enable gpu-node only for the specified cluster node.   ← "node" should be "nodes"
```

---

## Files Reviewed
- ✅ addons/gpu-node/Disable.ps1
- ✅ addons/gpu-node/Enable.ps1
- ✅ addons/gpu-node/Get-Status.ps1
- ✅ addons/gpu-node/README.md
- ✅ addons/gpu-node/addon.manifest.yaml
- ✅ k2s/cmd/k2s/cmd/addons/generic/generic.go
- ✅ k2s/cmd/k2s/cmd/addons/generic/generic_test.go
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/setup/gpu-worker.module.psm1

PR #2790, opened 2026-07-16
#2216 feat(build): add native Linux build system (build.sh + Makefile)

Saurabh-Kumar24:
Fixes #2216

### Motivation

K2s could only be built on Windows via `BuildGoExe.ps1`. This adds a native
Linux build path (`build.sh` + `Makefile`) so the Go executables can be built
on Linux with the same version metadata, as part of the "Reverted Setup Linux"
effort (#1149).

### Modifications

- `build.sh` — new POSIX build script (`set -euo pipefail`); builds all four Go
  executables with ldflags parity to `BuildGoExe.ps1`. Routes the `k2s` CLI
  binary to `bin/` alongside the other three, avoiding a name collision with the
  `k2s/` source directory.
- `Makefile` — new; `make build` / `make clean` targets wrapping `build.sh`.
- `k2s/internal/core/config/k2sconfig.go` — replaced the hardcoded Windows path
  separator in the config-file path with `filepath.Join` at all three call sites
  (`ReadK2sConfig`, `ReadSupportedWorkerOS`, `ReadKubeSwitchCIDR`) so the config
  path is correct on Linux.
- `docs/dev-guide/contributing/building-locally.md` — documented the Linux build.

### Verification

- Manual: `./build.sh` on a Linux VM built all four executables
  (`k2s`, `cloudinitisobuilder`, `httpproxy`, `yaml2json`) into `bin/`, with
  version metadata injected (VERSION 2.0.0, BUILD_DATE, GIT_COMMIT, tree-state).

### Acceptance Criteria

- [x] AC1: `k2s` binary builds into `bin/` alongside the 3 helpers (avoids `./k2s`
  vs `k2s/` source-dir collision) — VM build log `k2s -> .../bin/k2s`.
- [x] AC2: helper binaries build into `bin/` — VM build log.
- [x] AC3: ldflags parity with `BuildGoExe.ps1` — metadata injected in VM build.
- [x] AC4: SPDX (c) 2026 headers + `set -euo pipefail` — file inspection.

### Note

Two items are intentionally out of scope for this build-tooling change and are
expected to be picked up with downstream "Reverted Setup Linux" features:

- Running `./k2s version` (and other `k2s` commands) directly from the source
  tree cannot yet resolve `cfg/config.json`, because `InstallDir()` derives from
  the executable's own directory and assumes a deployed install layout. This is
  install-path behavior owned by #2229 (systemd / `k2s install` on Linux) and
  the command issues #2220-2227.
[...]

Issue #2216, opened 2026-04-22
[Reverted Setup Linux] Add native Linux build system (Makefile / build.sh)


krotz-dieter:
## Problem
The build system is PowerShell-only (`bgow.cmd`/`bgol.cmd` -> `BuildGoExe.ps1`). It is impossible to build the k2s CLI natively on a Linux host without PowerShell.

## Proposed Solution
Add a `Makefile` or shell script (`build.sh`) at the repo root that can build `k2s` and all helper binaries on Linux natively using `go build`.

## Relevant Files
- `bin/bgow.cmd`, `bin/bgol.cmd` - existing Windows build entry points
- `lib/scripts/k2s/build/BuildGoExe.ps1` - PowerShell build logic (reference for build targets)

## Acceptance Criteria
- [ ] Running `make build` (or `./build.sh`) on an Ubuntu host produces a working `k2s` Linux binary
- [ ] All Linux-targeted helper binaries (httpproxy, etc.) are also built
- [ ] Build script mirrors the same Go build flags and ldflags used by `BuildGoExe.ps1`

PR #2842, opened 2026-07-28
#2216 fix(build): fall back to unknown git metadata outside a git repo

Saurabh-Kumar24:
Refs #2216

### Motivation

`build.sh` (added in #2216 / PR #2790) called `git rev-parse HEAD`
unconditionally. When run from a directory without a `.git` folder (e.g.
files copied to a build/test VM without a full clone), `set -euo pipefail`
caused the entire build to abort immediately with
`fatal: not a git repository`, before any binaries were produced.

### Modifications

- `build.sh` — wrapped git commit/tag/tree-state detection in a
  `git rev-parse --is-inside-work-tree` guard. When true, behavior is
  unchanged (same commit/tag/tree-state logic). When false, falls back to
  `unknown`/empty version metadata (matching `internal/version`'s own
  defaults) and prints `No git repository found.` instead of aborting.

### Verification

- Manual: ran `./build.sh` on a Linux VM with no `.git` directory present.
  Confirmed it now prints `No git repository found.`, sets
  `GIT_COMMIT=unknown` / `GIT_TREE_STATE=unknown`, and completes the build,
  producing all four executables (`k2s.linux`, `cloudinitisobuilder`,
  `httpproxy`, `yaml2json`) against a vendored module cache.

### Acceptance Criteria

- [x] Script no longer aborts when run outside a git working tree — verified
  by VM run (fallback message printed, build completed).
- [x] Falls back to `unknown`/empty version metadata instead of failing —
  verified by VM run output.
- [x] Behavior unchanged when run inside a git repo — verified by code
  review (original logic preserved, only wrapped in a guard).

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -86,24 +86,29 @@ echo "VERSION: $VERSION"
 BUILD_DATE="$(date -u +'%Y-%m-%dT%H:%M:%SZ')"
 echo "BUILD_DATE: $BUILD_DATE"
 
-# GIT COMMIT
-GIT_COMMIT="$(git rev-parse HEAD)"
-echo "GIT_COMMIT: $GIT_COMMIT"
+# GIT COMMIT, TREE STATE, AND TAG
+# Falls back to "unknown" (instead of aborting the build under `set -e`)
+# when the script runs outside a git working tree, e.g. from files copied
+# without .git.
+if git rev-parse --is-inside-work-tree >/dev/null 2>&1; then
+    GIT_COMMIT="$(git rev-parse HEAD)"
 
-# GIT TREE STATE AND TAG
-GIT_TAG=""
-GIT_TREE_STATE="clean"
-if [[ -n "$(git status --porcelain)" ]]; then
-    GIT_TREE_STATE="dirty"
-else
-    # Clean tree: check for an exact tag to declare an official release.
-    if GIT_TAG="$(git describe --exact-match --tags HEAD 2>/dev/null)"; then
-        echo "GIT_TAG: $GIT_TAG"
+    GIT_TAG=""
+    GIT_TREE_STATE="clean"
+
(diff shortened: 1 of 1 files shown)


krotz-dieter (commented):
Here is my review of PR #2842:

---

## Review: `build.sh`

The fix correctly wraps the git metadata collection in a `git rev-parse --is-inside-work-tree` guard, preventing `set -e` from aborting the build when `.git` is absent. The fallback values (`unknown`) are safe to pass through to the Go `-ldflags`.

### One Issue: Missing `GIT_TAG` log output

**`build.sh`, lines ~114–121 (new code):**

The original script echoed `GIT_TAG: $GIT_TAG` when a tag was found, and printed `"No tag found for the git commit"` when it wasn't. The new code emits **neither** — `GIT_TAG` is silently computed but never logged. Every other metadata variable (`GIT_COMMIT`, `GIT_TREE_STATE`, `VERSION`, `BUILD_DATE`) has a corresponding `echo` after this block, but `GIT_TAG` has none.

This creates a blind spot in build logs: there is no way to confirm from output whether a release tag was detected or not. Add:
```bash
echo "GIT_TAG: $GIT_TAG"
```
after the `echo "GIT_TREE_STATE: $GIT_TREE_STATE"` line (or inside the `if` branch where the tag is set).

### Everything Else Looks Correct

- `git rev-parse --is-inside-work-tree >/dev/null 2>&1` is the right idiom for detecting a git working tree.
- `git describe --exact-match --tags HEAD 2>/dev/null || true` correctly suppresses both stderr and the non-zero exit code when no exact tag exists, assigning `""` to `GIT_TAG`.
- The dirty-tree check is preserved and logically equivalent to the original.
- Fallback values (`"unknown"`) are safe string literals compatible with Go's ldflags `-X` injection.

---

## Files Reviewed
- ✅ build.sh

Issue #2217, opened 2026-04-22
[Reverted Setup Linux] Add Linux CI workflow for build and unit tests


krotz-dieter:
## Problem
All CI workflows run on `windows-latest`. Linux Go code (build-tagged `_linux.go` files) is never compiled or tested in CI.

## Proposed Solution
Add `.github/workflows/build-k2s-cli-linux.yml` that:
- Runs on `ubuntu-latest`
- Builds all Linux binaries
- Runs `go test ./...` with the linux build tag

## Relevant Files
- `.github/workflows/build-k2s-cli.yml` - Windows CI workflow (use as template)

## Acceptance Criteria
- [ ] PR builds trigger Linux CI workflow
- [ ] All Linux-tagged Go files compile successfully
- [ ] Go unit tests pass on Linux runner
- [ ] Workflow mirrors caching patterns from existing Windows workflow

PR #2844, opened 2026-07-28
#2217 Improve status cmd for reverted setup

krotz-dieter:
#2217 Improve status cmd for reverted setup


krotz-dieter (commented):
Based on the diff provided, I can analyze all three files. The GitHub API endpoint appears to be an internal GHE instance, so I'll base the review on the diff.

---

## PR Review: #2844 — Improve status cmd for reverted setup

### `k2s/internal/provider/cluster_linux.go` — **Bug: inverted kubeconfig priority**

The new `linuxKubeconfig()` function unconditionally prefers `~/.kube/config` over `linuxAdminKubeconfig`:

```go
func linuxKubeconfig() string {
    if homeDir, err := os.UserHomeDir(); err == nil {
        userKubeconfig := filepath.Join(homeDir, ".kube", "config")
        if _, err := os.Stat(userKubeconfig); err == nil {
            return userKubeconfig  // always wins if it exists
        }
    }
    return linuxAdminKubeconfig
}
```

If both `~/.kube/config` and `linuxAdminKubeconfig` exist (normal operating condition), kubectl will use `~/.kube/config`. If that file points to a different cluster (e.g., a minikube, EKS, or previous K2s context), `k2s status` will silently report the status of the wrong cluster. The K2s-specific admin kubeconfig should take precedence; `~/.kube/config` should only be a fallback:

```go
func linuxKubeconfig() string {
    if _, err := os.Stat(linuxAdminKubeconfig); err == nil {
        return linuxAdminKubeconfig
    }
    if homeDir, err := os.UserHomeDir(); err == nil {
        userKubeconfig := filepath.Join(homeDir, ".kube", "config")
        if _, err := os.Stat(userKubeconfig); err == nil {
            return userKubeconfig
        }
    }
    return linuxAdminKubeconfig
}
```

### `k2s/internal/setuporchestration/provisioning_linux.go` — OK

The `MkdirAll` + `Chmod` pattern is correct: `MkdirAll` respects the process umask, so the explicit `Chmod` is needed to guarantee `0755` on an already-existing directory. No issues.

### `k2s/internal/setuporchestration/orchestrator_linux.go` — OK

Applying `0644` to `setup.json` immediately after writing it is appropriate to allow regular users to read cluster state. No issues.

---

## Files Reviewed
- ✅ k2s/internal/provider/cluster_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux.go
- ✅ k2s/internal/setuporchestration/provisioning_linux.go

krotz-dieter (commented):

---

## PR Review: #2844 — Improve status cmd for reverted setup

### `test/execute_all_tests.sh` — Bug: IFS trick produces wrong Ginkgo label filter

**Lines 104–107:**
```bash
if [[ ${#label_terms[@]} -gt 0 ]]; then
  label_filter="$(IFS=' || '; echo "${label_terms[*]}")"
```

This does **not** produce ` || `-separated output. In bash, `${array[*]}` uses only the **first character** of `IFS` as the separator — not the full string. Since `IFS=' || '`, the first character is a space, so elements are joined by a single space: `"tag1 tag2 tag3"` instead of `"tag1 || tag2 || tag3"`.

In Ginkgo's label filter language, space-separated labels mean AND (conjunction), which is the opposite of the intended OR. This would silently filter tests incorrectly whenever more than one tag is collected (e.g. multiple `--tags` entries, or any `--tags` combined with `--offline-mode`).

Fix using a loop:
```bash
label_filter="${label_terms[0]}"
for tag in "${label_terms[@]:1}"; do
  label_filter+=" || $tag"
done
```

---

### `k2s/internal/setuporchestration/orchestrator_linux.go` — Minor logic concern

The `os.Chmod` on `setup.json` is placed after `config.WriteRuntimeConfig`. If `WriteRuntimeConfig` fails to create the file (e.g., writes to a temp path and renames), the Chmod succeeds but targets a stale or wrong file. The error message `"set runtime config permissions"` also lacks a colon+space pattern consistent with the sibling `"create package staging directory: %w"` errors — should be `"set runtime config permissions: %w"`.

> Actually, looking again, the error string is already `"set runtime config permissions: %w"` — that's fine. The logic concern about temp-file atomicity is low risk but worth awareness.

---

### Everything else looks correct

- **`cluster_linux.go`** — `linuxKubeconfig()` fallback chain (admin → user → admin default) is correct for the reverted-setup scenario. No issues.
- **`provisioning_linux.go`** — `MkdirAll` + explicit `Chmod` pattern is the right way to override umask for the config directory. Ordering (before staging dir creation) is correct.
- **`kubeproxy.go`** — `|| r.nssmCli == nil` guard is necessary and correct; `r.restart(ctx)` directly uses `r.nssmCli`, confirmed from source.
- **`status_linux.go`** — New `StatusChecker` using `kubectl cluster-info --request-timeout=10s` is appropriate. Build tag is correct (`//go:build linux`). Complements the existing `status.go` (`//go:build windows`) cleanly.
- **`suite.go`** — Platform-specific CLI paths, nil `nssmCli` on Linux, and `statusChecker` always initialized are all correct. The `StatusChecker` type is properly split across `status.go` (Windows) and `status_linux.go`.

---

## Files Reviewed
- ✅ k2s/internal/provider/cluster_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux.go
- ✅ k2s/internal/setuporchestration/provisioning_linux.go
- ✅ k2s/test/framework/k2s/kubeproxy.go
- ✅ k2s/test/framework/k2s/status_linux.go
- ✅ k2s/test/framework/suite.go
- ✅ test/execute_all_tests.sh

krotz-dieter (commented):
The GitHub API isn't accessible for this repo. I'll base my review entirely on the provided diff.

---

## Code Review: PR #2844 — Improve status cmd for reverted setup

### `test/execute_all_tests.sh` — Logic bug in offline-mode label filter

**Lines ~107–124**: The `offline_mode` exclusion is appended to `label_terms` as an OR-joined item, but it should narrow results via AND:

```bash
if [[ "$offline_mode" == true ]]; then
  label_terms+=("!internet-required")   # ← becomes part of the OR chain
fi
# ...
label_filter="${label_terms[0]}"
for tag in "${label_terms[@]:1}"; do
  label_filter+=" || $tag"              # ← joined with ||
done
```

When `--tags acceptance --offline-mode` is passed, this produces:

```
label_filter="acceptance || !internet-required"
```

That matches *any* acceptance test **or** any non-internet test — so acceptance tests that require internet are **not** excluded as intended. The `exclude_tags` branch (lines ~128–133) correctly uses `&& !$tag`, but offline mode bypasses that path.

The fix is to apply the offline exclusion after the main filter is assembled, consistent with `exclude_tags`:

```bash
if [[ "$offline_mode" == true ]]; then
  if [[ -n "$label_filter" ]]; then
    label_filter="($label_filter) && !internet-required"
  else
    label_filter="!internet-required"
  fi
fi
```

**Also in the same script** (~line 100): `SYSTEM_TEST_PROXY` is unconditionally exported with the (possibly empty) value of `proxy`. If a caller already has `SYSTEM_TEST_PROXY` set in its environment and omits `--proxy`, this silently overrides it with an empty string. Consider only exporting when `proxy` is non-empty:

```bash
[[ -n "$proxy" ]] && export SYSTEM_TEST_PROXY="$proxy"
```

---

### `k2s/test/framework/k2s/status_linux.go` — Unused `ReloadRuntimeConfig()` call

**Line 34**: `sc.setupInfo.ReloadRuntimeConfig()` is called but its result is not used by the subsequent `kubectl cluster-info` check (which relies on the ambient `KUBECONFIG` env var, not on the runtime config). If this call exists to refresh shared state for other callers it should be documented; if it is not needed here it is dead code that could mislead maintainers.

---

### Remaining files — No issues found

- **`cluster_linux.go`**: `linuxKubeconfig()` fallback chain is correct and safe.
- **`orchestrator_linux.go`**: `os.Chmod` after `WriteRuntimeConfig` is idiomatic and correct.
- **`provisioning_linux.go`**: `MkdirAll` + `Chmod` pattern correctly bypasses umask to ensure 0755.
- **`kubeproxy.go`**: `|| r.nssmCli == nil` guard is necessary and correct given the suite.go change setting `nssmCli` to nil on Linux.
- **`suite.go`**: `runtime.GOOS` checks are logically consistent with the rest of the PR.

---

## Files Reviewed
- ✅ k2s/internal/provider/cluster_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux.go
- ✅ k2s/internal/setuporchestration/provisioning_linux.go
- ✅ k2s/test/framework/k2s/kubeproxy.go
- ✅ k2s/test/framework/k2s/status_linux.go
- ✅ k2s/test/framework/suite.go
- ✅ test/execute_all_tests.sh

krotz-dieter (commented):
Based on my thorough analysis of all changed files, here is my review:

---

## PR Review: #2844 — Improve status cmd for reverted setup

### `k2s/internal/setuporchestration/provisioning_linux.go`

**Security concern — proxy now binds to all interfaces**

The proxy `--addr` changed from `127.0.0.1:8181` to `:8181`. Access is now solely enforced via `--allowed-cidr` at the application level. If the proxy application's CIDR enforcement has any edge cases (e.g., IPv4-mapped IPv6 addresses), traffic from unintended sources could pass through. This change should be explicitly called out in release notes as a security-relevant behavior change.

Additionally, `localProxyURL = "http://127.0.0.1:8181"` (line 27) is now misleading — the proxy no longer binds to only `127.0.0.1`, yet this constant is still advertised as the canonical URL. It's only used for installation scripts that run locally, so there's no functional bug, but it will cause confusion for future maintainers.

**`linuxProxyNetworkExists` silently swallows all errors**

```go
output, err := runCommandOutput("ip", "-o", "-4", "addr", "show", "dev", proxyNetworkDev)
if err != nil {
    return false, nil // Interface not found is the normal first-install state.
}
```

Any failure — permission denied, `ip` binary path wrong, unexpected error — is treated as "interface not found". While subsequent calls to `ensureProxyNetworkIsAvailable` will surface `ip`-related failures, a permission error on the first `ip` call could be misdiagnosed, leading to a confusing second-failure message about address conflicts rather than the root cause.

### `k2s/internal/setuporchestration/orchestrator_linux.go`

**No rollback in `Start()` on partial failure**

```go
if cfg.LinuxOnly {
    if err := runCommand("systemctl", "start", proxyNetworkSvc); err != nil {
        return fmt.Errorf(...)
    }
}
if err := runCommand("systemctl", "start", proxyService); err != nil {
    return fmt.Errorf("failed to start local HTTP proxy: %w", err)
}
```

If `proxyNetworkSvc` starts successfully but `proxyService` (or any later `runCommand`) fails, `proxyNetworkSvc` is left running and `Start()` returns an error. The cluster is in an inconsistent state. Re-running `Start()` would succeed (systemd no-ops on an already-running service), so this self-heals, but a `Stop()` call in the failure path would be cleaner.

### `k2s/internal/setuporchestration/orchestrator_linux.go` — Uninstall `LinuxOnly` propagation

`removeHTTPProxy(cfg.LinuxOnly)` now depends on the caller passing the correct `LinuxOnly` flag during uninstall. If a user runs `k2s uninstall` without `--linux-only` on a system installed with `--linux-only`, the `k2s-proxy-network` service and its unit file are not cleaned up, leaving a dangling systemd service. The flag should ideally be read from the stored `setup.json` rather than from the CLI argument during uninstall. (This is a design issue not unique to this PR, but the new code makes it more impactful.)

### `k2s/internal/core/config/k2sconfig.go`

`prefixLength, _ := network.Mask.Size()` — silently ignores the "not in canonical form" case (returns `0, 0`). After `net.ParseCIDR` succeeds this is practically impossible, but if it ever happened, the resulting `AddressCIDR` would be `"<ip>/0"`, which would install a default-route-like address on the dummy interface — a serious misconfiguration. Using an explicit check (even just `if prefixLength == 0 { return error }`) would be more defensive.

### `test/execute_all_tests.sh`

The script uses `go run github.com/onsi/ginkgo/v2/ginkgo` which will re-download the module on each run in a cold cache environment (common on CI). Should use `go tool ginkgo` or a pre-built binary consistent with the Windows runner, to avoid version drift and cold-start latency.

### Everything else looks good

- The `linuxKubeconfig()` fallback chain is correct for the reverted-setup case.
- The `AfterSuite` nil check for `suite` is a correct defensive fix.
- `ReadKubeSwitchConfig` validation logic (address-in-network check) is correct.
- The `chmod 0644` on `setup.json` is the right fix for allowing non-root `k2s status`.
- The kubelet log drop-in approach is correct.

---

## Files Reviewed
- ✅ docs/op-manual/installing-k2s.md
- ✅ k2s/internal/logging/logging.go
- ✅ k2s/internal/provider/cluster_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux.go
- ✅ k2s/internal/setuporchestration/provisioning_linux.go
- ✅ k2s/test/e2e/cluster/core/core_test.go
- ✅ k2s/test/framework/k2s/kubeproxy.go
- ✅ k2s/test/framework/k2s/status_linux.go
- ✅ k2s/test/framework/suite.go
- ✅ test/execute_all_tests.sh

krotz-dieter (commented):
Here is my review of PR #2844:

---

## Review

### 🔴 Bug: `configureHostDNS` overwrites original resolver state if called while K2s is already running

**`k2s/internal/setuporchestration/provisioning_linux.go` — `configureHostDNS` / `readResolverUpstreams` / `saveResolverState`**

When `k2s start` is called while K2s is already running (e.g., after reboot without prior stop, or if `install` didn't invoke stop), `/etc/resolv.conf` already contains the K2s-managed content (`nameserver 127.0.0.1` with `resolverHeader` prefix).

1. `readResolverUpstreams()` reads `127.0.0.1` as the only upstream.
2. `saveResolverState()` overwrites the saved state file with the K2s-managed content (no guard against this).
3. The DNS proxy is configured with external upstream `127.0.0.1` — itself — causing an infinite forwarding loop for all non-Kubernetes queries.
4. `verifyHostDNS` only checks Kubernetes DNS (`kubernetes.default.svc.cluster.local`), **not** external DNS, so the loop is not caught and install/start succeeds.
5. On `k2s stop`, `restoreResolverState` restores `/etc/resolv.conf` to the now-saved K2s-managed state (`nameserver 127.0.0.1`) while the proxy is stopped — breaking all DNS.

**Fix**: At the start of `configureHostDNS`, check whether `/etc/resolv.conf` already starts with `resolverHeader` and either refuse with an error or skip resolver state save:
```go
current, _ := os.ReadFile(resolverPath)
if strings.HasPrefix(string(current), resolverHeader) {
    return fmt.Errorf("K2s already owns /etc/resolv.conf; stop K2s before starting again")
}
```

---

### 🟡 Bug: `stopHostDNS` leaves the DNS proxy running if resolver restoration fails

**`k2s/internal/setuporchestration/provisioning_linux.go` — `stopHostDNS`**

```go
func stopHostDNS(configDir string) error {
    if err := restoreResolverState(configDir); err != nil {
        return err  // DNS proxy keeps running
    }
    if err := runCommand("systemctl", "stop", dnsProxyService); err != nil { ...
```

If `restoreResolverState` fails (e.g., permissions issue, missing state file read error), the function returns immediately without stopping the DNS proxy. The proxy continues to run, the caller (`Stop()`) returns the error and stops, and subsequent `k2s start` or `k2s stop` calls will be confused about system state. The DNS proxy should be stopped regardless (in a `defer` or unconditionally), or at minimum a warning should be logged and the stop should continue.

---

### 🟡 Bug: `renderClusterIPWebhookDeployment` silently produces wrong manifest if defaults change

**`k2s/internal/setuporchestration/orchestrator_linux.go`**

```go
func renderClusterIPWebhookDeployment(manifest string, serviceCIDRs config.ServiceCIDRConfig) string {
    manifest = strings.ReplaceAll(manifest, "--linux-subnet=172.21.0.0/24", "--linux-subnet="+serviceCIDRs.LinuxCIDR)
    return strings.ReplaceAll(manifest, "--windows-subnet=172.21.1.0/24", "--windows-subnet="+serviceCIDRs.WindowsCIDR)
}
```

If the default CIDRs in `lib/manifests/clusterip-webhook/deployment.yaml` are ever changed, `ReplaceAll` silently finds nothing to replace and the manifest is deployed with the original hardcoded values — applying wrong Service CIDR partitions. There is no post-render validation that the substitution actually occurred. Consider verifying that the returned string contains the configured CIDRs, or using a proper template engine.

---

### 🟡 Security: `build.sh` downloads `dnsproxy` binary without integrity verification

**`build.sh`**

```bash
curl --fail --location --retry 3 --output "$BIN_DIR/$DNSPROXY_ARCHIVE" "$DNSPROXY_URL"
tar --extract ... "./linux-amd64/dnsproxy"
```

The `dnsproxy` binary is fetched over HTTPS and pinned to a version, but no SHA-256 checksum or GPG signature is verified against a value committed to the repository. The AdGuard release page publishes checksums. A compromised CDN or MITM could substitute a malicious binary that would be installed on every native Linux host. Add at minimum a `sha256sum --check` step using a pinned digest committed alongside the version string.

---

### 🟡 Minor: `verifyHostDNS` does not verify external (non-Kubernetes) DNS

**`k2s/internal/setuporchestration/provisioning_linux.go` — `verifyHostDNS`**

The function only resolves `kubernetes.default.svc.<zone>`. It does not validate that external DNS still resolves through the captured upstreams. Given the resolver-state overwrite bug above, a silent external-DNS loop would pass this check. Adding a basic external DNS lookup (e.g., `net.LookupHost("kubernetes.io")`) would catch misconfigured external upstreams.

---

## Files Reviewed

- ✅ build.sh
- ✅ docs/dev-guide/dns-host-topologies.md
- ✅ docs/op-manual/external-dns.md
- ✅ docs/op-manual/installing-k2s.md
- ✅ docs/op-manual/native-linux-host-dns.md
- ✅ k2s/internal/logging/logging.go
- ✅ k2s/internal/provider/cluster_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux_test.go
- ✅ k2s/internal/setuporchestration/provisioning_linux.go
- ✅ k2s/test/e2e/cluster/core/core_test.go
- ✅ k2s/test/framework/k2s/kubeproxy.go
- ✅ k2s/test/framework/k2s/status_linux.go
- ✅ k2s/test/framework/suite.go
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/distros/scripts/install-helm-yq.sh
- ✅ mkdocs.yml
- ✅ test/execute_all_tests.sh

krotz-dieter (commented):
I can't access the GitHub Enterprise API, so I'll work exclusively from the diff. Let me analyze all changed files carefully.

Here is my review of PR #2844:

---

## Code Review: PR #2844 — Improve status cmd for reverted setup

Despite the title, this is a large feature PR adding native Linux host DNS management, ClusterIP webhook deployment, proxy network configuration, and acceptance test infrastructure.

---

### 🔴 Security — Predictable temp file path

**`k2s/internal/setuporchestration/orchestrator_linux.go` — `deployClusterIPWebhook`**

```go
renderedPath := filepath.Join("/tmp", "k2s-clusterip-webhook-deployment.yaml")
if err := os.WriteFile(renderedPath, []byte(renderedDeployment), 0600); err != nil {
    ...
}
defer os.Remove(renderedPath)
```

This uses a fixed, predictable path under `/tmp`. A local attacker can pre-create a symlink at this path to overwrite an arbitrary file when the install runs as root. Use `os.CreateTemp("", "k2s-clusterip-webhook-*.yaml")` instead:

```go
tmpFile, err := os.CreateTemp("", "k2s-clusterip-webhook-*.yaml")
if err != nil { ... }
renderedPath := tmpFile.Name()
if _, err := tmpFile.WriteString(renderedDeployment); err != nil { ... }
_ = tmpFile.Close()
defer os.Remove(renderedPath)
```

---

### 🔴 Functionality — External DNS check breaks offline/air-gapped installs

**`k2s/internal/setuporchestration/provisioning_linux.go` — `verifyHostDNS`**

```go
if _, err := net.LookupHost("kubernetes.io"); err != nil {
    return fmt.Errorf("verify external host DNS through K2s DNS proxy: %w", err)
}
```

This lookup of a public internet name will always fail in air-gapped environments, causing `k2s install` and `k2s start` to fail even when internal cluster DNS is working correctly. This should either be skipped in offline mode or use a host-local name that is guaranteed to resolve (e.g., the host's own FQDN, or the configured gateway).

---

### 🟡 Logic — Post-replacement verification is dead code

**`k2s/internal/setuporchestration/orchestrator_linux.go` — `renderClusterIPWebhookDeployment`**

```go
manifest = strings.Replace(manifest, defaultLinuxArgument, "--linux-subnet="+serviceCIDRs.LinuxCIDR, 1)
manifest = strings.Replace(manifest, defaultWindowsArgument, "--windows-subnet="+serviceCIDRs.WindowsCIDR, 1)
if !strings.Contains(manifest, "--linux-subnet="+serviceCIDRs.LinuxCIDR) || !strings.Contains(manifest, "--windows-subnet="+serviceCIDRs.WindowsCIDR) {
    return "", fmt.Errorf("rendered ClusterIP webhook deployment does not contain configured service CIDRs")
}
```

After `strings.Replace` succeeds (guaranteed by the `strings.Count == 1` check just above), `strings.Contains` on the replaced value will *always* return true. The post-replacement check never fires and provides no protection. The `strings.Count` guard before the replacement is the right validation point; the second check should be removed to avoid misleading future readers.

---

### 🟡 Behavior — DNS restored before kubelet stops, but before API server stops

**`k2s/internal/setuporchestration/orchestrator_linux.go` — `Stop`**

```go
func (o *LinuxOrchestrator) Stop(cfg StopConfig) error {
    if err := stopHostDNS(cfg.ConfigDir); err != nil { ... }  // DNS restored here
    if err := runCommand("systemctl", "stop", "kubelet"); err != nil { ... }
    // CRI-O stops after kubelet
    if err := runCommand("systemctl", "stop", crioServiceName); err != nil { ... }
```

Restoring `/etc/resolv.conf` before stopping kubelet means Kubernetes internal service names (e.g., `kubernetes.default.svc.cluster.local`) immediately stop resolving on the host while the API server is still active. This is documented as intentional in the PR, but it can cause kubelet shutdown races if kubelet makes DNS queries to the API server during its shutdown sequence. Worth a `slog.Info` note or a brief comment at minimum.

---

### 🟡 Missing trailing newlines in documentation files

Both new documentation files end without a trailing newline:

- `docs/dev-guide/dns-host-topologies.md` — `\ No newline at end of file`
- `docs/op-manual/native-linux-host-dns.md` — `\ No newline at end of file`

Most Markdown linters and Git tools warn about this.

---

### ℹ️ Minor — `readResolverUpstreams` may fail on default systemd-resolved configurations

**`k2s/internal/setuporchestration/provisioning_linux.go` — `readResolverUpstreams`**

On systems where systemd-resolved is active and all DNS comes from per-interface DHCP (a common Ubuntu cloud VM setup), `resolvectl dns` may return only loopback addresses (127.0.0.53). The function then falls back to reading `/etc/resolv.conf`, which on such systems is a symlink to `/run/systemd/resolve/stub-resolv.conf` containing only `nameserver 127.0.0.53`. Since loopback addresses are filtered, this results in the error `no non-loopback nameserver entries found`, blocking installation. Consider using `resolvectl dns --json short` or `systemd-resolve --status` to get the actual configured upstream servers from networkd/NetworkManager.

---

### ℹ️ Minor — `build.sh` — SHA256 pin needs independent verification

The pinned hash `e3b82e72...` for `dnsproxy-linux-amd64-v0.83.1.tar.gz` cannot be verified from the diff alone. The format is correct (64 hex chars), but the value should be cross-checked against the GitHub release checksums file for AdguardTeam/dnsproxy v0.83.1 before merging.

---

## Files Reviewed

- ✅ build.sh
- ✅ docs/dev-guide/dns-host-topologies.md
- ✅ docs/op-manual/external-dns.md
- ✅ docs/op-manual/installing-k2s.md
- ✅ docs/op-manual/native-linux-host-dns.md
- ✅ k2s/internal/logging/logging.go
- ✅ k2s/internal/provider/cluster_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux_test.go
- ✅ k2s/internal/setuporchestration/provisioning_linux.go
- ✅ k2s/test/e2e/cluster/core/core_test.go
- ✅ k2s/test/framework/k2s/kubeproxy.go
- ✅ k2s/test/framework/k2s/status_linux.go
- ✅ k2s/test/framework/suite.go
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/distros/scripts/install-helm-yq.sh
- ✅ mkdocs.yml
- ✅ smallsetup/helpers/SwitchRegistry.ps1
- ✅ test/execute_all_tests.sh

krotz-dieter (commented):
Here is my review of PR #2844:

---

## Code Review

### Bug 1 — `runCommandOutput` returns combined stdout+stderr on success (Critical)

**`k2s/internal/setuporchestration/orchestrator_linux.go`**

The function was changed from `cmd.Output()` (stdout only) to `cmd.CombinedOutput()` (stdout+stderr), and this mixed output is returned to callers **even when the command succeeds**:

```go
output, err := cmd.CombinedOutput()
if err != nil {
    return "", fmt.Errorf("%s failed: %w; output: %s", name, err, ...)
}
return string(output), nil  // ← stderr mixed in here
```

This breaks every caller that parses the return value as structured data:

- **`discoverCoreDNS`** (`provisioning_linux.go`): parses the return value of `kubectl get service kube-dns -o jsonpath={.spec.clusterIP}` with `net.ParseIP`. If kubectl emits any deprecation/warning text to stderr, `serviceIP` will contain that text and `net.ParseIP` returns nil, causing `"kube-dns service has invalid ClusterIP"` failures.

- **`resolverIsImmutableFromOutput`** (`provisioning_linux.go`): expects `fields[0]` to be an lsattr attribute string like `----i---------`. Any stderr output from `lsattr` would occupy `fields[0]` instead.

- **`primaryHostIPv4FromRoute`** (`provisioning_linux.go`): parses `ip route get 1.1.1.1` output looking for `src <IP>`. stderr text would corrupt the field parsing.

- **`coreDNSZones`**: the entire Corefile content is parsed by the caller; stderr mixed in would produce garbage zones or cause `"no Kubernetes zones found"`.

The fix is to capture stdout and stderr separately — include stderr only in error messages, not in the success return value:

```go
var stdout, stderr bytes.Buffer
cmd.Stdout = &stdout
cmd.Stderr = &stderr
if err := cmd.Run(); err != nil {
    return "", fmt.Errorf("%s failed: %w; stderr: %s", name, err, stderr.String())
}
return stdout.String(), nil
```

---

### Bug 2 — `verifyHostDNS` requires public internet access (High)

**`k2s/internal/setuporchestration/provisioning_linux.go`**

```go
func verifyHostDNS(coreDNSIP string, zones []string) error {
    ...
    if _, err := net.LookupHost("kubernetes.io"); err != nil {
        return fmt.Errorf("verify external host DNS through K2s DNS proxy: %w", err)
    }
    return nil
}
```

`kubernetes.io` is a public domain. In air-gapped, firewall-restricted, or purely proxy-based corporate environments this lookup will fail, causing `k2s install` and `k2s start` to return an error and roll back the DNS configuration. The same DNS proxy configuration that worked up to this point would be torn down. The intent — verifying external DNS forwarding works — should use a domain that is guaranteed to resolve in the deployment environment, or this check should be skipped when an internet-accessible domain cannot be used.

---

### Bug 3 — `ReadKubeSwitchCIDR` behavioral regression if `kubeSwitch` not in `cfg/config.json`

**`k2s/internal/core/config/k2sconfig.go`**

`ReadKubeSwitchCIDR` previously read only `MasterNetworkCIDR`. It now delegates to `ReadKubeSwitchConfig`, which **also** validates the new `kubeSwitch` JSON field. If existing `config.json` files on installed systems don't have the `kubeSwitch` key, the field defaults to `""`, `net.ParseIP("")` returns nil, and `ReadKubeSwitchCIDR` fails with `"invalid KubeSwitch address"` — a regression for any caller that previously succeeded. The upgrade path for existing installations needs to be addressed (e.g., tolerate an empty/missing `kubeSwitch` field in the validation, or document a required config migration).

---

### Minor — `resolverUpstreamsFromOutput` picks up link-scoped DNS servers

**`k2s/internal/setuporchestration/provisioning_linux.go`**

The function tokenises all non-loopback IPs from `resolvectl dns` output including link-scoped servers (`Link 2 (eth0): ...`). On multi-homed hosts, VPN-interface DNS servers that should only resolve corporate names would be included as global upstreams for the K2s proxy, potentially causing incorrect DNS routing. The function should filter for `Global:` scope only when `resolvectl` output is present.

---

## Files Reviewed

- ✅ build.sh
- ✅ docs/dev-guide/dns-host-topologies.md
- ✅ docs/op-manual/external-dns.md
- ✅ docs/op-manual/installing-k2s.md
- ✅ docs/op-manual/native-linux-host-dns.md
- ✅ k2s/internal/logging/logging.go
- ✅ k2s/internal/provider/cluster_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux_test.go
- ✅ k2s/internal/setuporchestration/provisioning_linux.go
- ✅ k2s/test/e2e/cluster/core/core_test.go
- ✅ k2s/test/framework/k2s/kubeproxy.go
- ✅ k2s/test/framework/k2s/status_linux.go
- ✅ k2s/test/framework/suite.go
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/distros/scripts/install-helm-yq.sh
- ✅ mkdocs.yml
- ✅ smallsetup/helpers/SwitchRegistry.ps1
- ✅ test/execute_all_tests.sh

krotz-dieter (commented):
The SHA256 hash is correctly 64 hex digits. Now let me analyze the full diff carefully.

Here is my review:

---

## Code Review: PR #2844 — Improve status cmd for reverted setup

This is a large PR introducing native Linux host DNS management, proxy networking, ClusterIP webhook deployment, control-plane taint removal, and test framework improvements. Overall the structure and safety mechanics are well-thought-out. I found two confirmed bugs and one additional concern.

---

### 🐛 Bug 1: `runCommandOutput` mixes stderr into stdout on success (confirmed)

**File:** `k2s/internal/setuporchestration/orchestrator_linux.go`

```go
func runCommandOutput(name string, args ...string) (string, error) {
    cmd := exec.Command(name, args...)
    output, err := cmd.CombinedOutput()    // ← BUG
    if err != nil {
        return "", fmt.Errorf(...)
    }
    return string(output), nil             // ← returns stderr + stdout on SUCCESS
}
```

The comment says *"Keeping stderr in the error is essential"* — but `CombinedOutput()` includes stderr in the returned string on **success** too. On a successful exit, stderr warnings (e.g., kubectl API deprecation notices) are silently merged into the result string, corrupting any downstream parser.

**Directly broken callers:**
- `removeControlPlaneTaints()` uses `runCommandOutput` to list control-plane node names via JSONPath, then passes the result to `nodeNamesFromKubectlOutput()` which calls `strings.Fields()`. If kubectl emits any warning like `"Warning: Use tokens from TokenRequest API..."`, `strings.Fields` will split it into individual tokens (`"Warning:"`, `"Use"`, `"tokens"`, ...) that are then treated as node names. The subsequent `kubectl get node "Warning:"` call will fail with a misleading error.
- Same issue applies to the JSONPath taint-key query in the same function.

**Fix:** Capture stdout and stderr separately:
```go
var stdout, stderr bytes.Buffer
cmd.Stdout = &stdout
cmd.Stderr = &stderr
if err := cmd.Run(); err != nil {
    return "", fmt.Errorf("%s failed: %w; stderr: %s", name, err, strings.TrimSpace(stderr.String()))
}
return stdout.String(), nil
```

---

### 🐛 Bug 2: `verifyHostDNS` always queries the public internet (confirmed)

**File:** `k2s/internal/setuporchestration/provisioning_linux.go`

```go
func verifyHostDNS(coreDNSIP string, zones []string) error {
    ...
    if _, err := net.LookupHost("kubernetes.io"); err != nil {   // ← requires internet
        return fmt.Errorf("verify external host DNS through K2s DNS proxy: %w", err)
    }
    return nil
}
```

K2s supports offline/air-gapped installations (the test runner has `--offline-mode` and `SYSTEM_OFFLINE_MODE`, and the install script accepts `--no-proxy`). An air-gapped install would fail here because `kubernetes.io` cannot be resolved. This verification step would then call `removeDNSProxy` and abort installation even though Kubernetes DNS itself is working perfectly.

Either honor an offline flag passed through `InstallConfig`, or replace `kubernetes.io` with a well-known external IP like `8.8.8.8` for a reachability-only check.

---

### ⚠️ Concern: `nssmCli` nil-safety in the test framework

**File:** `k2s/test/framework/suite.go` + `k2s/test/framework/k2s/kubeproxy.go`

`nssmCli` is now `nil` on Linux. The `kubeproxy.go` was updated to guard against this. However, if `nssmCli` is passed to any other test-framework component that calls methods on it directly (without nil checking), those tests will panic on Linux. Verify there are no other call sites of `nssmCli` in the test framework.

---

### Minor: Missing newline at end of two new documentation files

- `docs/dev-guide/dns-host-topologies.md`
- `docs/op-manual/native-linux-host-dns.md`

Both end without a trailing newline (`\ No newline at end of file`). POSIX requires text files to end with a newline, and tools like `diff` and some Markdown processors warn about this.

---

## Files Reviewed

- ✅ build.sh
- ✅ docs/dev-guide/dns-host-topologies.md
- ✅ docs/op-manual/external-dns.md
- ✅ docs/op-manual/installing-k2s.md
- ✅ docs/op-manual/native-linux-host-dns.md
- ✅ k2s/internal/logging/logging.go
- ✅ k2s/internal/provider/cluster_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux_test.go
- ✅ k2s/internal/setuporchestration/provisioning_linux.go
- ✅ k2s/test/e2e/cluster/core/core_test.go
- ✅ k2s/test/framework/k2s/kubeproxy.go
- ✅ k2s/test/framework/k2s/status_linux.go
- ✅ k2s/test/framework/suite.go
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/distros/scripts/install-helm-yq.sh
- ✅ mkdocs.yml
- ✅ smallsetup/helpers/SwitchRegistry.ps1
- ✅ test/execute_all_tests.sh

krotz-dieter (commented):
Here is my review:

---

## Review of PR #2844 — Improve status cmd for reverted setup

This is a substantial native-Linux K2s feature PR. I found four genuine issues; the first is safety-critical.

---

### 🔴 Critical: `restoreResolverState` fails with ENOENT, leaving system without `/etc/resolv.conf`

**File:** `k2s/internal/setuporchestration/provisioning_linux.go`

`activateK2sResolver` calls `os.Remove(resolverPath)` and then `os.WriteFile(resolverPath, ...)`. If the write fails (e.g., out of disk space, permissions), `/etc/resolv.conf` is deleted and `activateK2sResolver` returns an error. The caller `configureHostDNS` then calls `removeDNSProxy → restoreResolverState`. However:

```go
func restoreResolverState(configDir string) error {
    // ...state file read succeeds...
    current, err := os.ReadFile(resolverPath)
    if err != nil {
        return fmt.Errorf("read current resolver configuration: %w", err) // ← returns error on ENOENT
    }
    if !isK2sManagedResolver(current) {
        return nil  // ← never reached
    }
    // ← restoration never happens
```

When `resolverPath` was just deleted, `os.ReadFile` returns ENOENT, so `restoreResolverState` returns an error **without** performing the restoration. The saved state exists on disk but is never applied, permanently removing DNS resolution from the host.

**Fix:** Treat ENOENT as a missing-but-restorable state:
```go
current, err := os.ReadFile(resolverPath)
if err != nil && !os.IsNotExist(err) {
    return fmt.Errorf("read current resolver configuration: %w", err)
}
if !os.IsNotExist(err) && !isK2sManagedResolver(current) {
    return nil
}
// proceed with restoration
```

---

### 🟠 Security: Fixed `/tmp` path for webhook deployment is vulnerable to symlink attacks

**File:** `k2s/internal/setuporchestration/orchestrator_linux.go`

```go
renderedPath := filepath.Join("/tmp", "k2s-clusterip-webhook-deployment.yaml")
if err := os.WriteFile(renderedPath, []byte(renderedDeployment), 0600); err != nil {
```

The installer runs as root. A low-privilege local process can pre-create `/tmp/k2s-clusterip-webhook-deployment.yaml` as a symlink to any root-writable file (e.g., `/etc/sudoers`, a systemd unit), and the root installer will write to that target. Use `os.CreateTemp("/tmp", "k2s-clusterip-webhook-*")` instead.

---

### 🟡 `execute_all_tests.sh` default label filter "acceptance" likely doesn't match any tests

**File:** `test/execute_all_tests.sh`

```sh
# Start with the portable core suite for the native Linux-only test. The
# suite is labelled sanity and skips Windows workloads when LinuxOnly is set.
suites=(./test/e2e/cluster/core)
...
else
  label_filter="acceptance"  # ← default when no --tags given
fi
```

The in-file comment says the suite is labelled **sanity**, but the default `label_filter` is **acceptance**. Unless the core tests carry an `acceptance` label (which the comment contradicts), running this script with no `--tags` argument will match no tests and silently pass. Either the default filter or the comment is wrong.

---

### 🟡 `SwitchRegistry.ps1`: Registry is recorded as logged-in even when Docker login is skipped

**File:** `smallsetup/helpers/SwitchRegistry.ps1`

```powershell
if ($nssmActionSucceeded) {
    Login-Docker -registry $RegistryName
} else {
    Write-Log "Skipping docker login..."
}

# Always runs regardless of $nssmActionSucceeded:
Set-ConfigValue -Path $global:SetupJsonFile -Key $global:ConfigKey_LoggedInRegistry -Value $RegistryName
```

When the Docker service action times out (ContinueOnTimeout path), the registry name is still written to `setup.json` as `LoggedInRegistry` even though Docker was not actually logged in. Downstream code that relies on this config key to determine Docker registry availability will be wrong.

---

### Minor observations (no action required)

- `orchestrator_linux.go` `waitForAPIServer` timeout change from warning to error is intentional and correct — `configureHostDNS` needs a live API server.
- `linuxKubeconfig()` fallback chain in `cluster_linux.go` is reasonable.
- `coreDNSZones` parsing correctly handles `{` on the same line and trailing dots.
- `renderClusterIPWebhookDeployment` post-render assertion is logically unreachable but harmless.
- `ReadServiceCIDRConfig` containment check validates only the network IP; a misconfigured subnet that starts inside the parent CIDR but extends beyond it would not be caught, but this requires perverse config values.

---

## Files Reviewed

- ✅ build.sh
- ✅ docs/dev-guide/dns-host-topologies.md
- ✅ docs/op-manual/external-dns.md
- ✅ docs/op-manual/installing-k2s.md
- ✅ docs/op-manual/native-linux-host-dns.md
- ✅ k2s/internal/logging/logging.go
- ✅ k2s/internal/provider/cluster_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux.go
- ✅ k2s/internal/setuporchestration/orchestrator_linux_test.go
- ✅ k2s/internal/setuporchestration/provisioning_linux.go
- ✅ k2s/test/e2e/cluster/core/core_test.go
- ✅ k2s/test/framework/k2s/kubeproxy.go
- ✅ k2s/test/framework/k2s/status_linux.go
- ✅ k2s/test/framework/suite.go
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/distros/scripts/install-helm-yq.sh
- ✅ mkdocs.yml
- ✅ smallsetup/helpers/SwitchRegistry.ps1
- ✅ test/execute_all_tests.sh

krotz-dieter:

```json
{
  "baseBranch": "main",
  "changedFilesCount": 21,
  "changedPaths": [
    {
      "path": "build.sh",
      "status": "modified"
    },
    {
      "path": "cfg/nodeextension/debian13/scripts/install-k8s-packages.sh",
      "status": "modified"
    },
    {
      "path": "docs/dev-guide/dns-host-topologies.md",
      "status": "added"
    },
    {
      "path": "docs/op-manual/external-dns.md",
      "status": "modified"
    },
    {
      "path": "docs/op-manual/installing-k2s.md",
      "status": "modified"
    },
    {
      "path": "docs/op-manual/native-linux-host-dns.md",
      "status": "added"
    },
    {
      "path": "k2s/internal/core/config/k2sconfig.go",
      "status": "modified"
    },
    {
      "path": "k2s/internal/logging/logging.go",
      "status": "modified"
    },
    {
      "path": "k2s/internal/provider/cluster_linux.go",
      "status": "modified"
    },
    {
      "path": "k2s/internal/setuporchestration/orchestrator.go",
      "status": "modified"
    },
    {
      "path": "k2s/internal/setuporchestration/orchestrator_linux.go",
      "status": "modified"
    },
    {
      "path": "k2s/internal/setuporchestration/orchestrator_linux_test.go",
      "status": "added"
    },
    {
      "path": "k2s/internal/setuporchestration/provisioning_linux.go",
      "status": "modified"
    },
    {
      "path": "k2s/test/e2e/cluster/core/core_test.go",
      "status": "modified"
    },
    {
      "path": "k2s/test/framework/k2s/kubeproxy.go",
      "status": "modified"
    },
    {
      "path": "k2s/test/framework/k2s/status_linux.go",
      "status": "added"
    },
    {
      "path": "k2s/test/framework/suite.go",
      "status": "modified"
    },
    {
      "path": "lib/modules/k2s/k2s.node.module/linuxnode/distros/scripts/install-helm-yq.sh",
      "status": "modified"
    },
    {
      "path": "mkdocs.yml",
      "status": "modified"
    },
    {
      "path": "smallsetup/helpers/SwitchRegistry.ps1",
      "status": "modified"
    },
    {
      "path": "test/execute_all_tests.sh",
      "status": "added"
    }
  ],
  "prNumber": 2844,
  "prTitle": "#2217 Improve status cmd for reverted setup",
  "prUrl": "https://github.com/Siemens-Healthineers/K2s/pull/2844",
  "rankedCandidates": [
    {
      "matchedFileCount": 14,
      "maxSpecificity": 3,
      "reasons": [
        {
          "path": "k2s/internal/core/config/k2sconfig.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/internal/logging/logging.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/internal/provider/cluster_linux.go",
          "rule": "prefix:provider",
          "specificity": 3,
          "weight": 110
        },
        {
          "path": "k2s/internal/provider/cluster_linux.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 65
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 65
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux_test.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 65
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux_test.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/internal/setuporchestration/provisioning_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 65
        },
        {
          "path": "k2s/internal/setuporchestration/provisioning_linux.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/test/e2e/cluster/core/core_test.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/test/framework/k2s/kubeproxy.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/test/framework/k2s/status_linux.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/test/framework/suite.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "smallsetup/helpers/SwitchRegistry.ps1",
          "rule": "prefix:smallsetup",
          "specificity": 2,
          "weight": 100
        },
        {
          "path": "test/execute_all_tests.sh",
          "rule": "prefix:test-folder",
          "specificity": 2,
          "weight": 85
        },
        {
          "path": "<mandatory>",
          "rule": "mandatory:cluster",
          "specificity": 0,
          "weight": 0
        }
      ],
      "score": 1435,
      "usecase": "cluster"
    },
    {
      "matchedFileCount": 6,
      "maxSpecificity": 3,
      "reasons": [
        {
          "path": "cfg/nodeextension/debian13/scripts/install-k8s-packages.sh",
          "rule": "keyword:node",
          "specificity": 1,
          "weight": 65
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux_test.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "k2s/internal/setuporchestration/provisioning_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "lib/modules/k2s/k2s.node.module/linuxnode/distros/scripts/install-helm-yq.sh",
          "rule": "keyword:node",
          "specificity": 1,
          "weight": 65
        }
      ],
      "score": 490,
    },
    {
      "matchedFileCount": 5,
      "maxSpecificity": 3,
      "reasons": [
        {
          "path": "k2s/internal/provider/cluster_linux.go",
          "rule": "prefix:provider",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 100
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 100
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux_test.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 100
        },
        {
          "path": "k2s/internal/setuporchestration/provisioning_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 100
        }
      ],
      "score": 490,
      "usecase": "linuxonly"
    },
    {
      "matchedFileCount": 4,
      "maxSpecificity": 3,
      "reasons": [
        {
          "path": "k2s/internal/setuporchestration/orchestrator.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux_test.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "k2s/internal/setuporchestration/provisioning_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        }
      ],
      "score": 360,
      "usecase": "add-linux-node-offline"
    },
    {
      "matchedFileCount": 4,
      "maxSpecificity": 3,
      "reasons": [
        {
          "path": "k2s/internal/setuporchestration/orchestrator.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 85
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 85
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux_test.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 85
        },
        {
          "path": "k2s/internal/setuporchestration/provisioning_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 85
        }
      ],
      "score": 340,
      "usecase": "add-linux-node-hyperv"
    },
    {
      "matchedFileCount": 2,
      "maxSpecificity": 3,
      "reasons": [
        {
          "path": "k2s/internal/provider/cluster_linux.go",
          "rule": "prefix:provider",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "smallsetup/helpers/SwitchRegistry.ps1",
          "rule": "prefix:smallsetup",
          "specificity": 2,
          "weight": 90
        }
      ],
      "score": 180,
      "usecase": "wsl"
    },
    {
      "matchedFileCount": 1,
      "maxSpecificity": 3,
      "reasons": [
        {
          "path": "k2s/internal/provider/cluster_linux.go",
          "rule": "prefix:provider",
          "specificity": 3,
          "weight": 75
        }
      ],
      "score": 75,
      "usecase": "startstop"
    },
    {
      "matchedFileCount": 1,
      "maxSpecificity": 3,
      "reasons": [
        {
          "path": "k2s/internal/provider/cluster_linux.go",
          "rule": "prefix:provider",
          "specificity": 3,
          "weight": 70
        }
      ],
      "score": 70,
      "usecase": "upgrade"
    },
    {
      "matchedFileCount": 1,
      "maxSpecificity": 2,
      "reasons": [
        {
          "path": "smallsetup/helpers/SwitchRegistry.ps1",
          "rule": "prefix:smallsetup",
          "specificity": 2,
          "weight": 65
        }
      ],
      "score": 65,
      "usecase": "ddrive"
    },
    {
      "matchedFileCount": 1,
      "maxSpecificity": 2,
      "reasons": [
        {
          "path": "smallsetup/helpers/SwitchRegistry.ps1",
          "rule": "prefix:smallsetup",
          "specificity": 2,
          "weight": 60
        }
      ],
      "score": 60,
      "usecase": "cluster-new-installdir"
    }
  ],
  "selectedRanked": [
    {
      "matchedFileCount": 14,
      "maxSpecificity": 3,
      "reasons": [
        {
          "path": "k2s/internal/core/config/k2sconfig.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/internal/logging/logging.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/internal/provider/cluster_linux.go",
          "rule": "prefix:provider",
          "specificity": 3,
          "weight": 110
        },
        {
          "path": "k2s/internal/provider/cluster_linux.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 65
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 65
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux_test.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 65
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux_test.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/internal/setuporchestration/provisioning_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 65
        },
        {
          "path": "k2s/internal/setuporchestration/provisioning_linux.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/test/e2e/cluster/core/core_test.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/test/framework/k2s/kubeproxy.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/test/framework/k2s/status_linux.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "k2s/test/framework/suite.go",
          "rule": "prefix:k2s",
          "specificity": 1,
          "weight": 80
        },
        {
          "path": "smallsetup/helpers/SwitchRegistry.ps1",
          "rule": "prefix:smallsetup",
          "specificity": 2,
          "weight": 100
        },
        {
          "path": "test/execute_all_tests.sh",
          "rule": "prefix:test-folder",
          "specificity": 2,
          "weight": 85
        },
        {
          "path": "<mandatory>",
          "rule": "mandatory:cluster",
          "specificity": 0,
          "weight": 0
        }
      ],
      "score": 1435,
      "usecase": "cluster"
    },
    {
      "matchedFileCount": 6,
      "maxSpecificity": 3,
      "reasons": [
        {
          "path": "cfg/nodeextension/debian13/scripts/install-k8s-packages.sh",
          "rule": "keyword:node",
          "specificity": 1,
          "weight": 65
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux_test.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "k2s/internal/setuporchestration/provisioning_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "lib/modules/k2s/k2s.node.module/linuxnode/distros/scripts/install-helm-yq.sh",
          "rule": "keyword:node",
          "specificity": 1,
          "weight": 65
        }
      ],
      "score": 490,
    },
    {
      "matchedFileCount": 5,
      "maxSpecificity": 3,
      "reasons": [
        {
          "path": "k2s/internal/provider/cluster_linux.go",
          "rule": "prefix:provider",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 100
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 100
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux_test.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 100
        },
        {
          "path": "k2s/internal/setuporchestration/provisioning_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 100
        }
      ],
      "score": 490,
      "usecase": "linuxonly"
    },
    {
      "matchedFileCount": 4,
      "maxSpecificity": 3,
      "reasons": [
        {
          "path": "k2s/internal/setuporchestration/orchestrator.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux_test.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        },
        {
          "path": "k2s/internal/setuporchestration/provisioning_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 90
        }
      ],
      "score": 360,
      "usecase": "add-linux-node-offline"
    },
    {
      "matchedFileCount": 4,
      "maxSpecificity": 3,
      "reasons": [
        {
          "path": "k2s/internal/setuporchestration/orchestrator.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 85
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 85
        },
        {
          "path": "k2s/internal/setuporchestration/orchestrator_linux_test.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 85
        },
        {
          "path": "k2s/internal/setuporchestration/provisioning_linux.go",
          "rule": "prefix:setuporchestration",
          "specificity": 3,
          "weight": 85
        }
      ],
      "score": 340,
      "usecase": "add-linux-node-hyperv"
    }
  ],
  "selectedUsecases": [
    "cluster",
    "add-linux-node-online",
    "linuxonly",
    "add-linux-node-offline",
    "add-linux-node-hyperv"
  ],
  "selectorVersion": "v1",
  "status": "success",
  "timestampUtc": "2026-08-18T15:38:54Z",
  "unknownPaths": [
    "build.sh",
    "docs/dev-guide/dns-host-topologies.md",
    "docs/op-manual/external-dns.md",
    "docs/op-manual/installing-k2s.md",
    "docs/op-manual/native-linux-host-dns.md",
    "mkdocs.yml"
  ],
  "warnings": [
    "Unmapped paths were found; review unknownPaths for rule improvements."
  ]
}
```

krotz-dieter:

```json
{
  "baseBranch": "main",
  "confidence": 0.9211,
  "headSha": "bdbf387c4353e286f3594a4ef523a892b0ff0ced",
  "prNumber": 2844,
  "selectedUsecases": [
    "cluster",
    "add-linux-node-online",
    "linuxonly",
    "add-linux-node-offline",
    "add-linux-node-hyperv"
  ],
  "selectorVersion": "v1",
  "shortReasons": [
    {
      "reason": "prefix:k2s",
      "usecase": "cluster"
    },
    {
      "reason": "keyword:node",
    },
    {
      "reason": "prefix:provider",
      "usecase": "linuxonly"
    },
    {
      "reason": "prefix:setuporchestration",
      "usecase": "add-linux-node-offline"
    },
    {
      "reason": "prefix:setuporchestration",
      "usecase": "add-linux-node-hyperv"
    }
  ],
  "status": "success",
  "timestampUtc": "2026-08-19T08:51:59Z"
}
```

krotz-dieter:

```json
{
  "baseBranch": "main",
  "confidence": 0.9211,
  "headSha": "3baf7377c4f1a400a35cb02d85c224761f887692",
  "prNumber": 2844,
  "selectedUsecases": [
    "cluster",
    "add-linux-node-online",
    "linuxonly",
    "add-linux-node-offline",
    "add-linux-node-hyperv"
  ],
  "selectorVersion": "v1",
  "shortReasons": [
    {
      "reason": "prefix:k2s",
      "usecase": "cluster"
    },
    {
      "reason": "keyword:node",
    },
    {
      "reason": "prefix:provider",
      "usecase": "linuxonly"
    },
    {
      "reason": "prefix:setuporchestration",
      "usecase": "add-linux-node-offline"
    },
    {
      "reason": "prefix:setuporchestration",
      "usecase": "add-linux-node-hyperv"
    }
  ],
  "status": "success",
  "timestampUtc": "2026-08-21T08:12:55Z"
}
```

krotz-dieter:

```json
{
  "baseBranch": "main",
  "confidence": 0.9211,
  "headSha": "efd19fcda915ef70e504b716956aff8abb0746ef",
  "prNumber": 2844,
  "selectedUsecases": [
    "cluster",
    "add-linux-node-online",
    "linuxonly",
    "add-linux-node-offline",
    "add-linux-node-hyperv"
  ],
  "selectorVersion": "v1",
  "shortReasons": [
    {
      "reason": "prefix:k2s",
      "usecase": "cluster"
    },
    {
      "reason": "keyword:node",
    },
    {
      "reason": "prefix:provider",
      "usecase": "linuxonly"
    },
    {
      "reason": "prefix:setuporchestration",
      "usecase": "add-linux-node-offline"
    },
    {
      "reason": "prefix:setuporchestration",
      "usecase": "add-linux-node-hyperv"
    }
  ],
  "status": "success",
  "timestampUtc": "2026-08-24T11:41:25Z"
}
```

PR #2851, opened 2026-07-29
Proxy handling refactor for image pull/push and registry add on Linux targets

### LINKED ISSUE #2839: Allow additional node to use proxy when accessing registry
**Is your feature request related to a problem? Please describe.**
In our case, the Windows host running KubeMaster can reach the target container registry. But the extended Linux node resides in an isolated internal network and cannot access the registry directly. Thus `k2s image pull ... --node [MyWorker]` fails (so does `registry add`)

**Describe the solution you'd like**
Is it possible to use proxy for extended node when accessing registry, like how it currently works for the KubeMaster?

sk1486:
Issue
Image operations on Linux nodes had inconsistent proxy handling.
As a result, image pull and registry add login could fail in isolated/corporate-proxy environments because proxy logic was duplicated and not uniformly applied across flows.

Fixes made :

Applied the similar proxy configuration as kubemaster image pull.

- Centralizes transparent proxy env setup (HTTP_PROXY, HTTPS_PROXY, NO_PROXY)
- Skips proxy for internal registry references (k2s.registry.local)
- Updated pull flow to use shared helper in 
- Updated registry add/login flow to use shared helper in [Add-Registry.ps1]
- Updated push flow to use shared helper and removed push-specific proxy helper in 
- Pull and registry add now use the same proxy behavior as push.
- External registry operations consistently route via K2s transparent proxy when configured.
- Internal k2s.registry.local traffic is not incorrectly proxied.
- Reduced duplicate code and proxy behavior drift.


krotz-dieter (commented):
Now I have enough context to write a thorough review.

Here is my review:

---

## PR Review: Proxy handling refactor for image pull/push and registry add on Linux targets

### Bug 1 — LinuxWorker push fallback routes internal traffic through corporate proxy
**File: `lib/scripts/k2s/image/Push-Image.ps1`, line 219**

```powershell
$fallbackCmdWithProxy = $fallbackCmd -replace '^sudo ', "sudo ${proxyCmd}"
$result = Invoke-CmdOnVmViaSSHKey -CmdToExecute $fallbackCmdWithProxy ...
```

`$proxyCmd` is computed based on `$Image` (the external source reference). When it contains proxy env vars, they will be applied to the fallback command, whose destination is `docker://$registryHostIp:30500/...` — a direct internal cluster NodePort IP. That IP is **not** in the NO_PROXY list (`k2s.registry.local,localhost,127.0.0.1`), so `buildah` will attempt to route that connection through the corporate proxy. The fallback is specifically designed to recover from DNS/resolution failures, so applying an external proxy to it defeats the purpose and will break fallback pushes in proxy environments.

The **ControlPlane path** (lines 200-201) correctly does NOT apply proxy to its fallback:
```powershell
$fallbackCmd = "sudo buildah push --tls-verify=false $Image $fallbackDestination 2>&1"
$result = Invoke-CmdOnControlPlaneViaSSHKey $fallbackCmd ...   # no proxy here
```

The LinuxWorker fallback should do the same — use `$fallbackCmd` directly (without `$proxyCmd`) since the fallback destination is always an internal cluster address.

---

### Bug 2 — ControlPlane `buildah login` in `Add-Registry.ps1` skips proxy
**File: `lib/scripts/k2s/image/registry/Add-Registry.ps1`, ~line 102**

The PR adds transparent proxy support for the LinuxWorker `buildah login` call, but the ControlPlane login path is unchanged:
```powershell
# ControlPlane path – no proxy applied:
$loginSuccess = (Invoke-CmdOnControlPlaneViaSSHKey "sudo buildah login ... '$escapedRegistry' > /dev/null 2>&1" ...).Success
```

Push operations on the ControlPlane **do** receive the proxy via `$pushCmdWithProxy` (new in this PR), so this gap means the ControlPlane can push to an external registry with proxy but cannot authenticate to it in an isolated network. Given the PR's stated goal covers "registry add on Linux targets", and the ControlPlane is a Linux target, this looks like an incomplete fix. The `Get-LinuxTransparentProxyPrefix` function should be called and prepended to the ControlPlane login command the same way it is for push.

---

### Minor Observation — `Get-RegistryHostIp` log prefix loses context
**File: `lib/scripts/k2s/image/Image-Common.module.psm1`**

The moved function now logs `[Image] Unable to detect registry pod node...` instead of the original `[Pull]`/`[Push]` context-specific prefix. This reduces traceability when troubleshooting registry detection failures during specific operations. Not a functional bug but worth noting.

---

## Files Reviewed
- ✅ addons/addons.module.psm1
- ✅ lib/scripts/k2s/image/Image-Common.module.psm1
- ✅ lib/scripts/k2s/image/Pull-Image.ps1
- ✅ lib/scripts/k2s/image/Push-Image.ps1
- ✅ lib/scripts/k2s/image/registry/Add-Registry.ps1

PR #2882, opened 2026-08-04
2816 reset containerd storage path

### LINKED ISSUE #2816: K2S uninstall leaves containerd files on the disk
**Describe the bug**

When uninstalling k2s, the containerd folder and files are left on the disk. Those are hard to remove as they have special ownership settings

**To Reproduce**
1. Install K2s freshly using custom path for containerd files (set in K2s config file to e.g. D:\Somaris\appdata)
2. Call "k2s uninstall"
==> containerd folder is still on the disk.

**Expected behavior**
As containerd folder gets created during "k2s install", it should also get removed when uninstalling k2s. It is unclear how this behaves if we later install newer versions of k2s when the folder is still there.

**Additional context**
Used K2s 1.8.1

vvkrishnakanth9:
Fixes #2816 

### Motivation

Support configured containerd storage during reset cleanup.

### Modifications

Resolve configured storage and clean HCS snapshot layers.

### Verification

Verified install, reset, cleanup, unit and Pester tests.


krotz-dieter (commented):

---

## Review: PR #2882 — Reset containerd storage path

### Bug: Lexicographic sort order in `Remove-WindowsSnapshotterLayers`

**`lib/scripts/k2s/image/ResetWinContainerStorage.ps1`, near line ~173:**

```powershell
Get-ChildItem -Path $snapshotsPath -Directory -ErrorAction SilentlyContinue |
    Sort-Object -Property Name -Descending |
    ForEach-Object { ... Invoke-ZapFolder ... }
```

`Sort-Object -Property Name -Descending` sorts snapshot directory names **as strings**, not integers. The comment explicitly states the intent is "Destroy most-derived layers first (child layers reference their parents)", which requires **numeric** descending order. With real-world layer IDs like `4` and `138`, string-descending sort yields `["4", "138"]` — destroying layer 4 before layer 138 — which is the **wrong** order (parent before child). This can cause HCS `DestroyLayer` to fail because a child layer still references the parent.

Fix:
```powershell
Sort-Object -Property { [int]$_.Name } -Descending |
```

The accompanying test (`ResetWinContainerStorage.Tests.ps1`, "attempts to destroy every snapshot layer individually") only asserts that `Write-Log` is called twice with `*Destroying container layer*`, but does **not** verify ordering. The sort bug is therefore invisible to the test suite.

---

### Minor: `Invoke-ZapFolder` silently no-ops when `zap.exe` is missing in cleanup path

**`lib/scripts/k2s/image/ResetWinContainerStorage.ps1`, `Invoke-CleanupOfContainerStorage`:**

The old code (`&"$(Get-KubeBinPath)\zap.exe" ...`) would emit a native shell error if `zap.exe` was absent. The replacement `Invoke-ZapFolder` logs at `-Error` level but then **returns silently**, leaving `$Directory` potentially intact without surfacing any failure to the caller. The subsequent `if (Test-Path $Directory)` check will catch this, but the overall cleanup is still reported as a failure without any clear signal that zap wasn't available. This is a behavioral regression that could make the ForceZap path harder to diagnose in the field.

---

### Observation: `$script:snapshotsPath` scoping in test `ParameterFilter`

**`lib/scripts/k2s/image/ResetWinContainerStorage.Tests.ps1`, ~lines 155–165:**

`$script:snapshotsPath` is assigned in `BeforeEach` but referenced as `$snapshotsPath` (no qualifier) inside `Should -Invoke ... -ParameterFilter { $Path -eq $snapshotsPath }`. Pester executes `ParameterFilter` scriptblocks in a constrained scope; whether the `$script:` variable is visible from there depends on the Pester version and execution context. If it resolves to `$null`, the filter silently matches nothing and the assertion passes vacuously. Using `$script:snapshotsPath` explicitly inside the filter would make this reliable.

---

### Everything else looks good

- The Go-side change (empty `defaultContainerdDir` → delegate path resolution to PowerShell) is correct and well-explained.
- The `Stop-ContainerdForStorageCleanup` function handles the `$null`-from-`Get-Process` case safely.
- The `takeown` output filtering regex (`'ERROR:\s*File or Directory not found\.'`) correctly uses `\.` for a literal period and is appropriately narrow.
- The `$Containerd` path resolution block (lines ~264–280 of the PS1) correctly falls back to `C:\containerd` on error, and the `Join-Path` construction handles the standard `\` separator from `Get-StorageLocalFolderName`.
- The new Go tests are well-structured; the flag mutation in "with explicit containerd directory" is safe assuming the existing `BeforeEach`/`resetFlags()` pattern resets it before subsequent tests.

---

## Files Reviewed
- ✅ k2s/cmd/k2s/cmd/image/resetwinstorage.go
- ✅ k2s/cmd/k2s/cmd/image/resetwinstorage_test.go
- ✅ lib/scripts/k2s/image/ResetWinContainerStorage.ps1
- ✅ lib/scripts/k2s/image/ResetWinContainerStorage.Tests.ps1

krotz-dieter (commented):
Now I have everything needed for a thorough review. Let me write it up.

---

## PR Review: #2882 – Reset containerd storage path

**Overall:** Solid improvement that replaces a hardcoded default path with dynamic resolution using the configured storage drive/folder, adds a safety validation guard, and properly handles Windows container image layers (HCS `DestroyLayer`). A few genuine issues are noted below.

---

### `lib/scripts/k2s/image/ResetWinContainerStorage.ps1`

**Bug: `$Containerd` is never normalized after path construction**

`Get-StorageLocalFolderName` returns `"\\" + $storageLocalDriveFolder` (lines 143-147 in `system.module.psm1`). When no custom folder is configured, `$storageLocalDriveFolder` is `""`, so the function returns `"\\"` (two backslashes). The resulting `$Containerd` value becomes `"C:\\containerd"` (double backslash):

```powershell
$Containerd = Join-Path "$storageDrive$storageFolder" 'containerd'   # → "C:\\containerd"
```

The PR author acknowledges this and normalizes `$displayPath` for logging — but the `$Containerd` variable itself (used in every subsequent `Test-Path`, `Join-Path`, `Test-IsContainerdStorage`, `Invoke-CleanupOfContainerStorage`, etc.) retains the double backslash. Windows file-system calls tolerate this because the kernel normalizes multiple separators, but it creates a subtle inconsistency: `Test-IsContainerdStorage` builds sub-paths like `"C:\\containerd\root"`, and any log messages from those sub-paths will be ugly. The safest fix is to normalize `$Containerd` in-place right after construction, not just the display copy:

```powershell
$Containerd = ($Containerd -replace '\\{2,}', '\')
```

---

### `lib/scripts/k2s/image/ResetWinContainerStorage.Tests.ps1`

**Bug: `Get-StorageLocalFolderName` mock value doesn't match real function output**

The tests mock `Get-StorageLocalFolderName` to return `'\'` (single backslash) and `'\Somaris\appdata'`, but the real function always prepends `"\\"`, returning `'\\'` and `'\\Somaris\appdata'` respectively. As a result, the tests verify path construction against `'C:\containerd'` / `'D:\Somaris\appdata\containerd'`, while production will construct `'C:\\containerd'` / `'D:\\Somaris\appdata\containerd'`. The tests pass only because the mock value is "cleaner" than the real value — they don't exercise the actual double-backslash path.

The mocks should match the real function's output:
```powershell
Mock -CommandName Get-StorageLocalFolderName { return '\\' }          # default
Mock -CommandName Get-StorageLocalFolderName { return '\\Somaris\appdata' }  # custom
```

And correspondingly the `Should -Invoke Test-Path` assertions should check the normalized path (after the proposed fix above) or check the double-backslash version.

---

### `k2s/cmd/k2s/cmd/image/resetwinstorage_test.go`

**Minor: overlapping test coverage**

The new `"with no containerd directory provided"` test (line ~110) and the pre-existing `"with docker directory only"` test (line ~94) both assert that when `containerdDirFlag` is unset the output contains `" -Containerd ''"`. They're not wrong but they overlap. Not a blocking issue.

---

### `k2s/cmd/k2s/cmd/image/resetwinstorage.go`

No issues. The change is clean: `defaultContainerdDir = ""` correctly delegates resolution to PowerShell, and the symmetry with docker (which still falls back to `determineDefaultDir("docker")`) is intentional.

---

## Files Reviewed

- ✅ `k2s/cmd/k2s/cmd/image/resetwinstorage.go`
- ✅ `k2s/cmd/k2s/cmd/image/resetwinstorage_test.go`
- ✅ `lib/scripts/k2s/image/ResetWinContainerStorage.ps1`
- ✅ `lib/scripts/k2s/image/ResetWinContainerStorage.Tests.ps1`

krotz-dieter (commented):
Now let me analyze the diff thoroughly.

---

## Review: PR #2882 — Reset containerd storage path

### Bug: Persistent flag mutation in Go tests causes potential test pollution

**File:** `k2s/cmd/k2s/cmd/image/resetwinstorage_test.go` (new "with explicit containerd directory" context)

```go
Context("with explicit containerd directory", func() {
    It("forwards the explicit path and never falls back to a default", func() {
        resetWinStorageCmd.Flags().Set(containerdDirFlag, `E:\Temp\Containerd`)
        _, params, err := buildResetPsCmd(resetWinStorageCmd)
        ...
    })
})
```

`resetWinStorageCmd` is a **package-level variable** (defined in `resetwinstorage.go`). Setting `containerdDirFlag` on it inside an `It` block with no corresponding reset leaves the flag permanently set to `E:\Temp\Containerd` for all subsequent tests within the enclosing `Ordered` `Describe`. The `Describe("resetWinStorage")` block that follows in the same `Ordered` container will inherit this mutated state, potentially causing false passes or unexpected failures in any test there that calls through `buildResetPsCmd`.

Fix: add an `AfterEach` to restore the flag:
```go
AfterEach(func() {
    resetWinStorageCmd.Flags().Set(containerdDirFlag, "")
})
```

---

### Minor: Redundant test

**File:** `k2s/cmd/k2s/cmd/image/resetwinstorage_test.go`

The new "with no containerd directory provided" test (`ContainElement(" -Containerd ''")`) is essentially duplicated by the **pre-existing** test at line ~108 which already exercises `buildResetPsCmd(resetWinStorageCmd)` with no flags set and asserts `ConsistOf(..., " -Containerd ''", ...)` (since `defaultContainerdDir` is now `""`). Not a bug, but worth noting.

---

### PS1 changes look correct

- `Test-IsContainerdStorage` — prefix-based check (`io.containerd.`) is correctly version-independent and the guard-before-destroy pattern is sound.
- `Remove-WindowsSnapshotterLayers` — numeric descending sort (child-before-parent) for `zap.exe` calls is correct; non-numeric names sorting last is a reasonable fallback.
- `Stop-ContainerdForStorageCleanup` — service + shim cleanup before deletion is a safe precaution.
- `Invoke-ZapFolder` — graceful no-op with an error log when `zap.exe` is absent is appropriate.
- Storage-path resolution (`Get-StorageLocalDrive` + `Get-StorageLocalFolderName` → `Join-Path ... 'containerd'`) with fallback to `C:\containerd` on exception is correct.
- `takeown` false-positive filter (`ERROR: File or Directory not found.`) on reparse points is well-motivated and isolated.

---

## Files Reviewed
- ✅ k2s/cmd/k2s/cmd/image/resetwinstorage.go
- ✅ k2s/cmd/k2s/cmd/image/resetwinstorage_test.go
- ✅ lib/scripts/k2s/image/ResetWinContainerStorage.ps1
- ✅ lib/scripts/k2s/image/ResetWinContainerStorage.Tests.ps1

PR #2914, opened 2026-08-10
fix: add retries to SSH commands for creating registry directories

GermanGLLopez:
Add retries to SSH commands for creating registry directories

### DIFF
--- a/addons/registry/Enable.ps1
+++ b/addons/registry/Enable.ps1
@@ -85,12 +85,12 @@ if ([string]::IsNullOrEmpty($StorageNode)) {
 Write-Log 'Creating authentication files and secrets' -Console
 $controlPlaneHostname = Get-ConfigControlPlaneNodeHostname
 if ($StorageNode -eq $controlPlaneHostname) {
-    (Invoke-CmdOnControlPlaneViaSSHKey -Timeout 2 -CmdToExecute 'sudo mkdir -p /registry /registry/auth /registry/repository').Output | Write-Log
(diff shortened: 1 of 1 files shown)


krotz-dieter (commented):
I've confirmed the function signatures. Both `Invoke-CmdOnControlPlaneViaSSHKey` and `Invoke-CmdOnVmViaSSHKey` are defined in `lib/modules/k2s/k2s.node.module/linuxnode/vm/vm.module.psm1` and both accept a `-Retries [uint16]` parameter. The retry logic is fully implemented.

---

## Review: fix: add retries to SSH commands for creating registry directories

This is a straightforward and correct fix. The `-Retries` parameter is confirmed to exist on both `Invoke-CmdOnControlPlaneViaSSHKey` and `Invoke-CmdOnVmViaSSHKey` (defined in `lib/modules/k2s/k2s.node.module/linuxnode/vm/vm.module.psm1`). The retry logic is already implemented in those functions — with `Retries=3`, the command will be attempted up to 4 times total (1 initial + 3 retries) with a 2-second delay between attempts (governed by the existing `-Timeout 2` parameter).

One minor observation worth mentioning: the existing `-Timeout 2` parameter controls the **inter-retry sleep delay** (in seconds), not an SSH execution timeout. The SSH commands here have **no execution timeout** (`ExecutionTimeoutSeconds` defaults to 0, meaning they can run indefinitely). If the SSH connection hangs rather than failing fast, the retries won't help. This is pre-existing behavior and not introduced by this PR, but it may be worth a follow-up to also pass an `-ExecutionTimeoutSeconds` value for more robust handling.

No bugs or correctness issues with the changes in this PR.

## Files Reviewed
- ✅ addons/registry/Enable.ps1

PR #2919, opened 2026-08-11
Fix test addviewer

GermanGLLopez:
Fix test addviewer

pre-pull cert-manager images on control-plane node

### DIFF
--- a/addons/addons.module.psm1
+++ b/addons/addons.module.psm1
@@ -1746,7 +1746,7 @@ function Install-CertManagerControllers {
     
     foreach ($image in $certManagerImages) {
         Write-Log "[CertManager] Pre-pulling image: $image" -Console
-        $pullResult = Invoke-CmdOnControlPlaneViaSSHKey -Timeout 10 -CmdToExecute "sudo crictl pull '$image' 2>&1"
+        $pullResult = Invoke-CmdOnControlPlaneViaSSHKey -CmdToExecute "sudo timeout 300 crictl pull '$image' 2>&1"
         if ($pullResult.Success -eq $true) {
             Write-Log "[CertManager] Pre-pull succeeded: $image"
         }
--- a/addons/addons.module.psm1
+++ b/addons/addons.module.psm1
@@ -1737,6 +1737,27 @@ function Install-CertManagerControllers {
     [CmdletBinding()]
     param()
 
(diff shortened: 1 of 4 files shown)


krotz-dieter (commented):
Based on my review of the source code, I can now provide a thorough analysis.

---

## Code Review: PR #2919 — Fix test addviewer

### `addons/addons.module.psm1`

#### 🐛 Bug: `-Timeout 10` does not limit image pull duration

**Line:** `$pullResult = Invoke-CmdOnControlPlaneViaSSHKey -Timeout 10 -CmdToExecute "sudo crictl pull '$image' 2>&1"`

After examining `Invoke-CmdOnControlPlaneViaSSHKey` → `Invoke-CmdOnVmViaSSHKey` in `vm.module.psm1`, the `-Timeout` parameter is the **retry delay in seconds** (used in `Start-Sleep -Seconds $Timeout` between retry attempts). With the default `$Retries = 0`, this value has **no practical effect** — there are no retries to delay.

The actual SSH execution timeout is controlled by `-ExecutionTimeoutSeconds`, which defaults to `0` (no timeout). However, `Invoke-CmdOnControlPlaneViaSSHKey` does not expose `ExecutionTimeoutSeconds` as a parameter at all.

**Impact:** A stalled or very slow `crictl pull` (e.g., quay.io down, network degraded) will block the entire `Install-CertManagerControllers` function indefinitely, potentially hanging the addon installation with no recovery path. This is ironic given the stated goal of the PR is to prevent races against readiness deadlines.

**Suggested fix:** Either add `ExecutionTimeoutSeconds` support to `Invoke-CmdOnControlPlaneViaSSHKey` and use it with a sensible value (e.g., 300 seconds per image), or embed a timeout directly in the `crictl pull` command:
```powershell
"sudo timeout 300 crictl pull '$image' 2>&1"
```

---

#### ✅ Timeout increase: `--wait=3m` → `--wait=10m`

The `Wait-ForCertManagerAvailable` change is straightforward and reasonable — 3 minutes is often insufficient in slow or constrained environments. No issues.

---

#### ✅ Image extraction / error handling

`Get-ImagesFromYaml` is confirmed to exist with the correct signature. The failure path for pre-pull (`$pullResult.Success -ne $true`) is handled gracefully and logs a warning without aborting installation, which is correct behavior.

---

## Files Reviewed
- ✅ addons/addons.module.psm1

PR #2920, opened 2026-08-11
Fix test addon rollout argocd

GermanGLLopez:
Fixes for test addon-rollout-argocd
increase API server readiness attempts and adjust related logging
implement retries for ArgoCD status checks in tests
use server-side apply to handle large CRDs in rollout addon installation
increase timeout for rollout status checks in ArgoCD addon installation

### DIFF
--- a/addons/rollout/argocd/Enable.ps1
+++ b/addons/rollout/argocd/Enable.ps1
@@ -107,7 +107,7 @@ $rolloutConfig = Get-RolloutConfig
 
 Write-Log 'Waiting for pods being ready...' -Console
 
-$kubectlCmd = (Invoke-Kubectl -Params 'rollout', 'status', 'deployments', '-n', $rolloutNamespace, '--timeout=300s')
+$kubectlCmd = (Invoke-Kubectl -Params 'rollout', 'status', 'deployments', '-n', $rolloutNamespace, '--timeout=600s')
 Write-Log $kubectlCmd.Output
 if (!$kubectlCmd.Success) {
     $errMsg = 'rollout addon could not be deployed successfully!'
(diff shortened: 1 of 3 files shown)


krotz-dieter (commented):
The GitHub API isn't accessible (private GHE), so I'll analyze strictly from the provided diff.

---

## PR Review: Fix test addon rollout argocd

### `addons/rollout/argocd/Enable.ps1`

**`--force-conflicts` flag risk** — The new invocation:
```powershell
(Invoke-Kubectl -Params 'apply', '--server-side', '--force-conflicts', '-n', $rolloutNamespace, '-k', $rolloutConfig)
```
`--force-conflicts` silently takes field ownership away from any existing field manager without warning. In a fresh install this is benign, but if a cluster operator or another tool has previously patched any of these resources, this will silently overwrite those fields on every subsequent `Enable` run. This isn't a crash-level bug, but it can cause surprising behavior in day-2 operations. Consider logging a warning or scoping `--force-conflicts` only to a known-problematic subset of CRDs rather than the entire apply batch.

Timeout increase (300s → 600s): reasonable and uncontroversial.

---

### `k2s/test/e2e/addons/rollout/argocd/rollout-argocd_test.go`

No bugs found. The two retry patterns are used correctly:

- `Eventually(func() string { return MustExec(...) })...Should(SatisfyAll(...))` — correct for output-matching retries where the command is expected to succeed.
- `Eventually(func(g Gomega) { g.Expect(...) })...Should(Succeed())` — correct Gomega pattern; `g.Expect` failures are caught and retried by `Eventually` rather than aborting the test immediately.

---

### `lib/scripts/k2s/system/startup/Start-System.ps1`

No bugs found. The changes (15→20 attempts, MaxAttempts 3→6) are consistent timeout increases; the log message on the failure path is updated to match.

---

## Files Reviewed
- ✅ addons/rollout/argocd/Enable.ps1
- ✅ k2s/test/e2e/addons/rollout/argocd/rollout-argocd_test.go

PR #2928, opened 2026-08-13
fix: update internet reachability parameters for improved reliability

GermanGLLopez:
update internet reachability parameters for improved reliability

### DIFF
--- a/k2s/test/framework/k8s/k8s.go
+++ b/k2s/test/framework/k8s/k8s.go
@@ -44,10 +44,10 @@ type Cluster struct {
 const clusterReadinessTimeout = 10 * time.Minute
 
 const (
-	internetReachabilityCurlMaxTime       = "60"
-	internetReachabilityCurlRetryCount    = "5"
-	internetReachabilityCheckTimeout      = 3 * time.Minute
-	internetReachabilityCheckPollInterval = 10 * time.Second
+	internetReachabilityCurlMaxTime       = "20"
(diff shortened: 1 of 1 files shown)


krotz-dieter (commented):
## PR Review: fix: update internet reachability parameters for improved reliability

### Summary

This PR adjusts four constants used in `ExpectInternetToBeReachableFromPodOfDeployment` to change how internet reachability is checked after a node reboot.

### Analysis

**The old configuration had a logical problem:**

- `internetReachabilityCurlMaxTime = "60"` with `internetReachabilityCurlRetryCount = "5"` meant a single `curl` invocation could block for up to **360 seconds** (60s × 6 total attempts), while the outer `Eventually` timeout was only **3 minutes (180s)**. The outer `Eventually` would time out and abandon the check while `curl` was still internally retrying — a misaligned design.

**The new configuration is more coherent:**

- Per-attempt curl time drops to 20s, with **0 curl-level retries** — the outer `Eventually` loop now exclusively owns the retry logic.
- The outer timeout increases to 15 minutes, giving ample room for post-reboot network recovery, with the poll interval staying short enough (15s) to detect recovery promptly.

With 20s per curl call + 15s poll interval, you get roughly ~25 retry attempts over the 15-minute window, which is a reasonable number for a reboot scenario.

### No issues found

The changes are logically consistent and fix the mismatch between curl's internal retry budget and the outer `Eventually` timeout. No bugs, regressions, or correctness issues identified.

## Files Reviewed

PR #2949, opened 2026-08-17
fix: skip disk-check gaurd when online installation is not triggered

abhitoshj:
when the k2s artifacts are not present in the bin, then we implemented a gaurd to throw exception when the kubemaster vhdx is not found. This is a scenario of online installation which is not forced by the user. 

In this PR, we skip the check if the vhdx is not found to be present.

### DIFF
--- a/lib/modules/k2s/k2s.node.module/linuxnode/vm/vm.module.psm1
+++ b/lib/modules/k2s/k2s.node.module/linuxnode/vm/vm.module.psm1
@@ -748,7 +748,8 @@ function Assert-ControlPlaneArtifactDiskSizeCompatibility {
 
     $controlPlaneBaseImagePath = Get-ControlPlaneVMBaseImagePath
     if (!(Test-Path -Path $controlPlaneBaseImagePath)) {
-        throw "[PREREQ-FAILED] Precheck failed: expected packaged control-plane artifact is missing at '$controlPlaneBaseImagePath'. Installation cannot validate disk-size compatibility safely. Restore a complete package (including base VHDX) or run an explicit online installation mode."
+        Write-Log "[Precheck] Skip package disk-size guard because packaged control-plane artifact is missing at '$controlPlaneBaseImagePath'. This is allowed for online installation when the base VHDX is not present in the local bin folder."
+        return
     }
 
     try {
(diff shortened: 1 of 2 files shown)


krotz-dieter (commented):
Here's my review:

---

## Review

### `lib/modules/k2s/k2s.node.module/linuxnode/vm/vm.module.psm1`

**Silent bypass weakens a safety-critical guard (line 750–752)**

The function already has an explicit, intentional mechanism to skip the disk-size check for online installations — the `$ForceOnlineInstallation` switch (lines 744–747). This PR adds a **second, implicit bypass**: if the base VHDX file is absent, the guard silently returns regardless of whether online installation was actually intended.

This is problematic because:

1. **Masks real errors**: In an offline/package installation scenario where the VHDX is accidentally missing (corrupted package, failed extraction, wrong path), the original code would fail fast with a clear `[PREREQ-FAILED]` message. The new code silently skips the compatibility check, allowing installation to proceed and likely fail much later with a harder-to-diagnose error.

2. **Duplicates intent via implicit signal**: The correct fix for the scenario described ("online installation where VHDX is not present") should be to ensure those callers pass `-ForceOnlineInstallation`, not to make the guard infer installation mode from file absence. The `$ForceOnlineInstallation` parameter exists precisely for this purpose.

3. **Guard is now effectively**: `if ($ForceOnlineInstallation -or !(Test-Path ...)) { return }` — the second condition can mask the first in a confusing way.

**Recommendation**: If there is a code path that triggers an online install without setting `$ForceOnlineInstallation`, fix that call site to pass `-ForceOnlineInstallation:$true`. Keep the guard throwing on absent VHDX for all other cases.

---

### `lib/modules/k2s/k2s.node.module/linuxnode/vm/vm.module.unit.tests.ps1`

The test update is consistent with the new behavior, but it only verifies that no exception is thrown and that the two mocked functions were invoked. It does **not** verify that `Write-Log` was called with the expected message (the log message is the only observable side-effect of the early return now). This is a minor gap in test coverage but worth noting given the log message is the only signal a user has that the check was skipped.

---

## Files Reviewed
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/vm/vm.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/vm/vm.module.unit.tests.ps1

PR #3008, opened 2026-08-24
feat: add function to restore control plane transit route to Windows …

GermanGLLopez:
feat: add function to restore control plane transit route to Windows loopback adapter in case it is lost

### DIFF
--- a/lib/scripts/k2s/system/startup/Start-System.ps1
+++ b/lib/scripts/k2s/system/startup/Start-System.ps1
@@ -271,6 +271,46 @@ function Get-MissingWindowsPodRoutesOnControlPlane {
     return $missingRoutes
 }
 
+function Restore-ControlPlaneTransitRoute {
+    <#
+    .SYNOPSIS
+        Ensures the control plane has a route to the Windows loopback adapter subnet.
+    .DESCRIPTION
+        After a reboot or k2s stop/start, the Linux control-plane node may lose its
+        route to the Windows loopback adapter subnet (e.g. 172.22.1.0/24). Without this
+        route, flannel's host-gw mode cannot install pod subnet routes because the
+        gateway (the Windows node's InternalIP on the loopback adapter) is unreachable.
+        This function checks for and restores the transit route via the KubeSwitch gateway.
+    #>
+    $loopbackCIDR = Get-LoopbackAdapterCIDR
+    $kubeSwitchIP = Get-ConfiguredKubeSwitchIP
+
+    if ([string]::IsNullOrWhiteSpace($loopbackCIDR) -or [string]::IsNullOrWhiteSpace($kubeSwitchIP)) {
+        Write-Log "[$logUseCase] WARNING: Cannot restore transit route - loopbackCIDR='$loopbackCIDR', kubeSwitchIP='$kubeSwitchIP'"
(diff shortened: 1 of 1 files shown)


krotz-dieter (commented):
Let me look for `Get-LoopbackAdapterCIDR` and `Get-ConfiguredKubeSwitchIP` in the module structure.

Now I have enough context. `Get-ConfiguredKubeSwitchIP` is confirmed in `config.module.psm1` (returns `$ipNextHop` from config). `Get-LoopbackAdapterCIDR` is in `loopbackadapter.module.psm1`. Both functions exist.

---

## Code Review: `lib/scripts/k2s/system/startup/Start-System.ps1`

### Bug: Incomplete Regex Boundary in Route Existence Check

**Location:** `Restore-ControlPlaneTransitRoute`, route check condition (~line 298 in the PR branch)

```powershell
if (-not [string]::IsNullOrWhiteSpace($existingRoute) -and $existingRoute -match "via\s+$([regex]::Escape($kubeSwitchIP))") {
```

The regex is missing a trailing word boundary. The analogous check in `Get-MissingWindowsPodRoutesOnControlPlane` (introduced in the same file) correctly uses:

```powershell
$expectedGatewayPattern = "(^|\s)via\s+$([regex]::Escape($nodeRoute.InternalIP))(\s|$)"
```

Without the trailing `(\s|$)`, if `$kubeSwitchIP` is e.g. `10.0.0.1`, this pattern matches a route line containing `via 10.0.0.10` or `via 10.0.0.100`. This produces a false positive — the function concludes the transit route is correctly in place and returns early without restoring it, when in fact the route points to a different gateway. The fix is to use the same boundary pattern as in the existing function:

```powershell
if (-not [string]::IsNullOrWhiteSpace($existingRoute) -and $existingRoute -match "(^|\s)via\s+$([regex]::Escape($kubeSwitchIP))(\s|$)") {
```

---

### Minor: Missing `$routeResult.Success` Check Before Acting on Output

If the SSH route-check command fails (e.g., control plane unreachable), `$routeResult.Success` is `$false` but the code unconditionally reads `.Output` and falls through to attempt route restoration. The restore will also fail, and both outcomes log a WARNING and continue — so this is not a crash. However, it does mean a transient SSH failure silently triggers an unnecessary route-replace attempt. The existing `Get-MissingWindowsPodRoutesOnControlPlane` follows the same pattern, so this is consistent, but it's worth being aware of in debugging scenarios.

---

## Files Reviewed

krotz-dieter:

```json
{
  "baseBranch": "main",
  "confidence": 0.8083,
  "headSha": "240a35fb04eb2a3c41ba2549d1e1015b6acaa607",
  "prNumber": 3008,
  "selectedUsecases": [
    "cluster",
    "startstop",
    "reboot"
  ],
  "selectorVersion": "v1",
  "shortReasons": [
    {
      "reason": "prefix:system-scripts",
      "usecase": "cluster"
    },
    {
      "reason": "prefix:system-scripts",
      "usecase": "startstop"
    },
    {
      "reason": "prefix:system-scripts",
      "usecase": "reboot"
    }
  ],
  "status": "success",
  "timestampUtc": "2026-08-25T06:49:40Z"
}
```

PR #3011, opened 2026-08-25
2886 upgrade custom config dir

### LINKED ISSUE #2886: Error when using k2s system upgrade
**Describe the bug**
We have installed k2s 1.8.0 with some customizations (see screenshot).
When we want to upgrade to 1.8.1, this fails with the error:

c:\Somaris\bin\k2s-1.8.1-product-offline>k2s system upgrade
🤖 Analyze current cluster and check prerequisites ...
WARNING You have not installed K2s setup yet, please start the installation with command 'k2s.exe install' first

The log didn't directly say why K2s is not installed, I looked into the source code a bit and think it's caused by missing "C:\ProgramData\K2s\setup.json", which is actually located at "C:\ProgramData\Siemens\syngo\k2s\setup.json"

**To Reproduce**
1. Install K2s 1.8.0 but change the configDir.k2s value in config.json to some custom path
2. Exctract the K2s 1.8.1 package
3. Call k2s system upgrade on the 1.8.1. package

**Expected behavior**
The upgrade should also work when configDir.k2s is different from the default values.

**Screenshots**

Image

vvkrishnakanth9:
Fixes #2886 

### Motivation

Support upgrades with customized configDir.k2s.

### Modifications

Discover existing installation and preserve custom configuration during upgrade.

### Verification

Verified custom configDir.k2s upgrade with manual and e2e tests.


krotz-dieter (commented):
Now I have all the context needed. Let me write the review.

---

## PR Review: #3011 — 2886 upgrade custom config dir

### Summary
This PR enables `k2s system upgrade` to work correctly when the installed K2s uses a non-default `configDir.k2s`. The approach — discovering the installed K2s via PATH, then propagating the config dir to child processes via `K2S_SETUP_CONFIG_DIR` while protecting the package's own `cfg/config.json` from modification — is sound and the overall design is well thought-out.

---

### Bug: `resolveInstallDirFromPath` silently drops a corrupted-but-valid install dir

**File:** `k2s/cmd/k2s/cmd/cmd.go`

```go
func resolveInstallDirFromPath(exeDir string, originalErr error) (string, error) {
    installed, resolveErr := config.ResolveInstalledK2s()
    if resolveErr != nil {
        slog.Warn("Could not discover installed K2s via PATH", "error", resolveErr)
        return exeDir, originalErr
    }
    ...
}
```

`config.ResolveInstalledK2s()` can return `(non-nil installed, ErrSystemInCorruptedState)` — that is the documented contract of `resolveInstalledK2s` (case 1 in the switch + `corruptedErr`). Because `resolveErr != nil` is checked without exempting `ErrSystemInCorruptedState`, the function:
1. Logs "Could not discover installed K2s via PATH" — factually wrong, the installation *was* found.
2. Returns `exeDir` instead of `installed.InstallDir` — the delta-package upgrade resolves to the wrong directory.

The upgrade path in `upgrade.go` already handles this correctly with `!errors.Is(err, cconfig.ErrSystemInCorruptedState)`. The same exemption is needed here:

```go
func resolveInstallDirFromPath(exeDir string, originalErr error) (string, error) {
    installed, resolveErr := config.ResolveInstalledK2s()
    if resolveErr != nil && !errors.Is(resolveErr, config.ErrSystemInCorruptedState) {
        slog.Warn("Could not discover installed K2s via PATH", "error", resolveErr)
        return exeDir, originalErr
    }
    if installed == nil {
        return exeDir, originalErr
    }
    slog.Info("Resolved actual install directory via PATH", "install-dir", installed.InstallDir)
    return installed.InstallDir, nil
}
```

---

### Observation: missing test for the corrupted-state branch in `resolveInstallDirFromPath`

**File:** `k2s/cmd/k2s/cmd/cmd.go` (no corresponding test file in the diff)

The `resolveInstallDirFromPath` helper has no unit tests. Given that the handling of the corrupted-state case is wrong (see above), a test exercising `ResolveInstalledK2s` returning `(installed, ErrSystemInCorruptedState)` would have caught it.

---

### Minor: `FindOtherExecutablesInPath` always case-insensitive

**File:** `k2s/internal/os/path.go`

```go
if !strings.EqualFold(absPath, currentExeAbs) {
```

The comment says "case-insensitive to match Windows path semantics", but `strings.EqualFold` makes it case-insensitive on **all** platforms. On Linux, two executables that differ only in case would incorrectly be treated as the same file. This is unlikely in practice and K2s is Windows-first, but worth noting since `path.go` now lives in a shared `os` package.

---

## Files Reviewed
- ✅ docs/op-manual/configuration-reference.md
- ✅ k2s/cmd/k2s/cmd/cmd.go
- ✅ k2s/cmd/k2s/cmd/install/install.go
- ✅ k2s/cmd/k2s/cmd/system/upgrade/upgrade.go
- ✅ k2s/cmd/k2s/cmd/system/upgrade/upgrade_test.go
- ✅ k2s/internal/core/config/installed.go
- ✅ k2s/internal/core/config/installed_test.go
- ✅ k2s/internal/core/config/k2sconfig_setupconfigdir_test.go
- ✅ k2s/internal/definitions/definitions.go
- ✅ k2s/internal/os/path.go
- ✅ lib/modules/k2s/k2s.cluster.module/upgrade/upgrade.module.psm1
- ✅ lib/modules/k2s/k2s.infra.module/config/config.module.psm1
- ✅ lib/modules/k2s/k2s.infra.module/config/config.setupconfigdir.unit.tests.ps1

krotz-dieter (commented):
Now I have enough context to write a thorough review.

---

## Code Review: PR #3011 — Upgrade Custom Config Dir

Overall the PR is well-structured and handles the core scenario correctly. A few issues worth addressing:

---

### 🔴 Bug: Case-sensitive path comparison in `applySetupConfigDirOverride` (Windows)

**File:** `k2s/cmd/k2s/cmd/system/upgrade/upgrade.go`

```go
func applySetupConfigDirOverride(resolvedConfigDir string, packageConfigDir string) error {
    if resolvedConfigDir == "" || resolvedConfigDir == packageConfigDir {  // ← case-sensitive
```

On Windows, file paths are case-insensitive. If `resolvedConfigDir` is `C:\ProgramData\K2s` and `packageConfigDir` is `c:\programdata\k2s` (or any other casing variant), this check returns `false` and an unnecessary `K2S_SETUP_CONFIG_DIR` override is set. This would cause the PowerShell `Invoke-ClusterInstall` to redundantly call `Update-SetupConfigDir`.

The existing test only compares identical strings and doesn't cover the same-path-different-case scenario. This should use `strings.EqualFold` on Windows (or use `dirKey` from `installed.go` for consistent normalization):

```go
if resolvedConfigDir == "" || strings.EqualFold(resolvedConfigDir, packageConfigDir) {
```

---

### 🟡 Inconsistency: `readInstalledConfigFromPath` bypasses injection

**File:** `k2s/cmd/k2s/cmd/system/upgrade/upgrade.go`

`cmd.go` correctly uses an injectable variable for testing:
```go
var resolveInstalledK2s = config.ResolveInstalledK2s
```

But `upgrade.go`'s `readInstalledConfigFromPath` hardcodes the call:
```go
installed, err := config.ResolveInstalledK2s()
```

This makes `readInstalledConfigFromPath` untestable in isolation. The upgrade PATH-discovery codepath has no unit tests, while the analogous `cmd.go` codepath does. For consistency and testability, `upgrade.go` should expose a similar injectable variable.

---

### 🟡 Incomplete nil-guard in `readInstalledConfigFromPath`

**File:** `k2s/cmd/k2s/cmd/system/upgrade/upgrade.go`

```go
installed, err := config.ResolveInstalledK2s()
if err != nil && !errors.Is(err, cconfig.ErrSystemInCorruptedState) {
    return nil, "", err
}
if installed == nil {
    return nil, "", cconfig.ErrSystemNotInstalled  // ← drops ErrSystemInCorruptedState
}
```

If `installed == nil && err == ErrSystemInCorruptedState` were ever reached (currently impossible per `resolveInstalledK2s`'s contract, but defensively worth handling), the corrupted-state error would be silently converted to `ErrSystemNotInstalled`, changing how `upgradeCluster` responds to it (corrupted state would not trigger the user-facing corrupted-state message). Should be:

```go
if installed == nil {
    if err != nil {
        return nil, "", err
    }
    return nil, "", cconfig.ErrSystemNotInstalled
}
```

---

### ✅ Confirmed correct

- `config.module.psm1`: `$script:k2sConfigDir` scope is correctly readable from functions without the prefix (PowerShell's scope lookup), and `Update-SetupConfigDir` updates all three derived variables. ✓
- `upgrade.module.psm1`: `Update-SetupConfigDir` is accessible because the infra module recursively imports all sub-modules including `config.module.psm1`. ✓
- `installed.go`: deduplication via `seenDirs` + `dirKey` correctly handles repeated PATH entries including case-insensitive Windows paths. ✓
- `installed.go`: A single corrupted candidate propagates `ErrSystemInCorruptedState` alongside the result; multiple candidates return an explicit ambiguity error. ✓
- `ReadInstallFolder` wraps the error with `%w`, so the test's `MatchError(os.ErrNotExist)` will pass via `errors.Is`. ✓
- Tests in `installed_test.go` and `cmd_test.go` cover the important edge cases (corrupted state, stale/missing PATH entries, duplicate entries). ✓

---

## Files Reviewed
- ✅ docs/op-manual/configuration-reference.md
- ✅ k2s/cmd/k2s/cmd/cmd.go
- ✅ k2s/cmd/k2s/cmd/cmd_test.go
- ✅ k2s/cmd/k2s/cmd/install/install.go
- ✅ k2s/cmd/k2s/cmd/system/upgrade/upgrade.go
- ✅ k2s/cmd/k2s/cmd/system/upgrade/upgrade_test.go
- ✅ k2s/internal/core/config/installed.go
- ✅ k2s/internal/core/config/installed_test.go
- ✅ k2s/internal/core/config/k2sconfig_setupconfigdir_test.go
- ✅ k2s/internal/definitions/definitions.go
- ✅ k2s/internal/os/path.go
- ✅ k2s/internal/os/path_test.go
- ✅ lib/modules/k2s/k2s.cluster.module/upgrade/upgrade.module.psm1
- ✅ lib/modules/k2s/k2s.infra.module/config/config.module.psm1
- ✅ lib/modules/k2s/k2s.infra.module/config/config.setupconfigdir.unit.tests.ps1

krotz-dieter (commented):
Based on my thorough review of all changed files, this is a well-structured PR implementing PATH-based fallback discovery for K2s installations that use a customized `configDir.k2s`. Here are my findings:

---

## Review

### `k2s/internal/core/config/installed.go`

**`corruptedErr` is never reset across candidates.** If a first candidate is corrupted (`corruptedErr` is set) and a second candidate is also valid (no error), both are added to `found`, which hits the `len(found) > 1` path and returns "multiple installed K2s" — `corruptedErr` is silently discarded. This is the correct behavior here (multiple installations is still an error), but the comment at `case 1:` ("return found[0], corruptedErr") could mislead a future reader into thinking corruptedErr is always consistent with a single result. Worth a clarifying comment.

**`ReadInstallFolder` double-reads `setup.json`.** In `resolveInstalledK2s`, `ReadRuntimeConfig(setupConfigDir)` already parsed `setup.json`. Immediately after, `ReadInstallFolder(setupConfigDir)` reads and parses the same file again just to check `InstallFolder` for a warning. This is a minor inefficiency, not a correctness issue, but it could be made cleaner by extracting `InstallFolder` from the already-parsed `ReadRuntimeConfig` result. (The runtime config struct doesn't expose `InstallFolder`, so this would need contract changes — acceptable to defer.)

### `k2s/cmd/k2s/cmd/system/upgrade/upgrade.go`

**`applySetupConfigDirOverride` has no cleanup on failure after `switchToUpgradeLogFile`.** `applySetupConfigDirOverride` calls `os.Setenv` which essentially can't fail in practice, but if it did return an error, the env var wouldn't be set, `switchToDefaultLogFile` is called, and the function returns an error. This is correct.

**No issue:** when `readInstalledConfigFromPath` returns `ErrSystemInCorruptedState` with a non-nil `installed`, the error is propagated to `upgradeCluster` which returns `CreateSystemInCorruptedStateCmdFailure()` — consistent with pre-existing behavior.

### `lib/modules/k2s/k2s.infra.module/config/config.module.psm1`

**`Get-K2sConfigDir` and `Get-SetupConfigFilePath` use unqualified variable names.** `Get-K2sConfigDir` returns `$k2sConfigDir` and `Get-SetupConfigFilePath` returns `$SetupJsonFile` (both without `$script:`). After `Update-SetupConfigDir` updates `$script:k2sConfigDir` / `$script:SetupJsonFile`, these functions will correctly see the new values because in PowerShell modules, unqualified variable reads in module functions resolve to the module's script scope — which is what `$script:` refers to. No bug.

**All other `$SetupJsonFile` references throughout the module** (lines 343, 356, 359, 367, 470, etc.) similarly use the unqualified name and will correctly pick up updates from `Update-SetupConfigDir`. ✓

### `k2s/internal/core/config/installed_test.go`

This file is in `package config` (internal tests) while the suite runner (`TestConfig`) is in `package config_test`. The Ginkgo specs in `installed_test.go` will run under `TestConfig`'s suite via Ginkgo's `init()`-based registration — correct and intentional, but these tests inherit the `"integration"` label from `config_test.go`. Since `"ci"` is also in the labels, they'll run in CI, so no functional impact.

### Overall

The logic is sound, error propagation is consistent (especially the corrupted-state passthrough pattern used in both `cmd.go` and `upgrade.go`), and test coverage is comprehensive. No security issues identified.

---

## Files Reviewed

- ✅ docs/op-manual/configuration-reference.md
- ✅ k2s/cmd/k2s/cmd/cmd.go
- ✅ k2s/cmd/k2s/cmd/cmd_test.go
- ✅ k2s/cmd/k2s/cmd/install/install.go
- ✅ k2s/cmd/k2s/cmd/system/upgrade/upgrade.go
- ✅ k2s/cmd/k2s/cmd/system/upgrade/upgrade_test.go
- ✅ k2s/internal/core/config/installed.go
- ✅ k2s/internal/core/config/installed_test.go
- ✅ k2s/internal/core/config/k2sconfig_setupconfigdir_test.go
- ✅ k2s/internal/definitions/definitions.go
- ✅ k2s/internal/os/path.go
- ✅ k2s/internal/os/path_test.go
- ✅ lib/modules/k2s/k2s.cluster.module/upgrade/upgrade.module.psm1
- ✅ lib/modules/k2s/k2s.infra.module/config/config.module.psm1
- ✅ lib/modules/k2s/k2s.infra.module/config/config.setupconfigdir.unit.tests.ps1

krotz-dieter:

```json
{
  "baseBranch": "main",
  "confidence": 0.8929,
  "headSha": "a0c6d3029cdf98489657d851b5d5317f11cb2a42",
  "prNumber": 3011,
  "selectedUsecases": [
    "cluster",
    "startstop",
    "linuxonly",
    "wsl",
    "upgrade"
  ],
  "selectorVersion": "v1",
  "shortReasons": [
    {
      "reason": "prefix:k2s-cmd",
      "usecase": "cluster"
    },
    {
      "reason": "prefix:k2s-cmd",
      "usecase": "startstop"
    },
    {
      "reason": "prefix:k2s-cmd",
      "usecase": "linuxonly"
    },
    {
      "reason": "prefix:k2s-cmd",
      "usecase": "wsl"
    },
    {
      "reason": "keyword:upgrade",
      "usecase": "upgrade"
    }
  ],
  "status": "success",
  "timestampUtc": "2026-08-26T10:00:02Z"
}
```

PR #3012, opened 2026-08-25
fix: enhance viewer status check with retry logic for readiness

GermanGLLopez:
fix: enhance viewer status check with retry logic for readiness

### DIFF
--- a/k2s/test/e2e/addons/viewer/viewer_test.go
+++ b/k2s/test/e2e/addons/viewer/viewer_test.go
@@ -418,7 +418,8 @@ func expectStatusToBePrinted(ctx context.Context) {
 	// rollout). The JSON status check performs an HTTP health probe, so we retry
 	// until the application is fully ready.
 	Eventually(func(g Gomega) {
-		output = suite.K2sCli().MustExec(ctx, "addons", "status", "viewer", "-o", "json")
+		output, exitCode := suite.K2sCli().Exec(ctx, "addons", "status", "viewer", "-o", "json")
+		g.Expect(exitCode).To(Equal(0), "k2s addons status viewer -o json exited with code %d", exitCode)
 
 		var addonStatus status.AddonPrintStatus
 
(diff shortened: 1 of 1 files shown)


krotz-dieter (commented):
Based on the diff provided, here is my analysis:

---

## Review

### `k2s/test/e2e/addons/viewer/viewer_test.go`

**Potential issue: `MustExec` may bypass the `Eventually` retry mechanism**

```go
Eventually(func(g Gomega) {
    output = suite.K2sCli().MustExec(ctx, "addons", "status", "viewer", "-o", "json")
    // ...
    g.Expect(...).To(...)
}).WithTimeout(90 * time.Second).WithPolling(10 * time.Second).Should(Succeed())
```

The name `MustExec` strongly implies it fails immediately (e.g., via a global `Expect`/`Ω` or `panic`) when the CLI command exits with a non-zero status, rather than returning an error. In Gomega's `Eventually(func(g Gomega))` pattern, **only failures raised through the supplied `g Gomega` argument are caught and retried**. Failures raised through the global `Expect` / `GinkgoFail` / `panic` abort the entire `Eventually` block immediately.

If `MustExec` uses global Gomega assertions internally, any transient CLI-level failure (e.g. the viewer CLI itself returning an error during the startup window) will terminate the retry loop rather than allowing it to poll again. The retry would only work for the `g.Expect(...)` assertions that follow — not for the `MustExec` call that precedes them.

**Recommendation:** Replace `MustExec` with a variant that returns `(string, error)` and use `g.Expect(err).To(Succeed())`, e.g.:

```go
Eventually(func(g Gomega) {
    output, err := suite.K2sCli().Exec(ctx, "addons", "status", "viewer", "-o", "json")
    g.Expect(err).To(Succeed())
    // ...
}).WithTimeout(90 * time.Second).WithPolling(10 * time.Second).Should(Succeed())
```

This ensures every failure path is retried consistently. Whether there is an `Exec` counterpart to `MustExec` should be verified against the `K2sCli` implementation.

---

No security issues or other logic errors found. The `addonStatus` rename (to avoid shadowing the `status` package) and moving the variable declaration inside the closure are both correct.

## Files Reviewed
- ✅ k2s/test/e2e/addons/viewer/viewer_test.go

PR #3020, opened 2026-08-27
Several fixes for security to stabilize test

GermanGLLopez:
Several fixes for security to stabilize test

### DIFF
--- a/addons/addons.module.psm1
+++ b/addons/addons.module.psm1
@@ -1857,7 +1857,7 @@ Waits for the cert-manager API to be available.
 #>
 function Wait-ForCertManagerAvailable {
     param(
-        [string]$WaitTimeout = '2m'
+        [string]$WaitTimeout = '10m'
     )
     $out = &$cmctlExe check api --wait=$WaitTimeout
     if ($out -match 'The cert-manager API is ready') {
--- a/addons/addons.module.psm1
+++ b/addons/addons.module.psm1
@@ -1856,7 +1856,10 @@ function Initialize-CACertificateIssuer {
 Waits for the cert-manager API to be available.
 #>
 function Wait-ForCertManagerAvailable {
-    $out = &$cmctlExe check api --wait=10m
+    param(
+        [string]$WaitTimeout = '2m'
+    )
+    $out = &$cmctlExe check api --wait=$WaitTimeout
     if ($out -match 'The cert-manager API is ready') {
         return $true
     }
--- a/addons/security/Disable.ps1
+++ b/addons/security/Disable.ps1
@@ -114,6 +114,11 @@ $linkerdYamlCNI = Get-LinkerdConfigCNI
 $linkerdYamlCertManager = Get-LinkerdConfigCertManager
 (Invoke-Kubectl -Params 'delete', '--ignore-not-found', '-f', $linkerdYamlCertManager).Output | Write-Log
 
+# Clean up linkerd PKI secrets to prevent cert-manager key mismatch on re-enable
+(Invoke-Kubectl -Params 'delete', 'secret', 'linkerd-trust-anchor', '-n', 'cert-manager', '--ignore-not-found').Output | Write-Log
+(Invoke-Kubectl -Params 'delete', 'secret', 'linkerd-previous-anchor', '-n', 'cert-manager', '--ignore-not-found').Output | Write-Log
+(Invoke-Kubectl -Params 'delete', 'secret', 'linkerd-identity-issuer', '-n', 'linkerd', '--ignore-not-found').Output | Write-Log
+
 $linkerdYamlTrustManager = Get-LinkerdConfigTrustManager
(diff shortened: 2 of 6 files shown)


krotz-dieter (commented):
Let me check for other callers of `Wait-ForCertManagerAvailable` to assess the timeout change impact:

I found a real issue. `Install-CertManagerControllers` in `addons.module.psm1` calls `Wait-ForCertManagerAvailable` without parameters — confirmed in the main branch source above. The PR reduces the default timeout from `10m` to `2m`, which silently affects this caller.

Here is my full review:

---

## Review of PR #3020 — Several fixes for security to stabilize test

Overall this is a solid set of CI-stabilization fixes with clear intent. I found one breaking regression and two design issues worth noting.

---

### 🔴 Bug: `Install-CertManagerControllers` silently loses 8 minutes of wait budget

**File:** `addons/addons.module.psm1`

`Wait-ForCertManagerAvailable` had its default `WaitTimeout` changed from `10m` to `2m`. The PR correctly updates the callers in `Enable.ps1` and `Get-Status.ps1` to pass `'30s'` explicitly. However, `Install-CertManagerControllers` — also in `addons.module.psm1` — calls `Wait-ForCertManagerAvailable` **without any parameter**:

```powershell
# Install-CertManagerControllers (addons.module.psm1, not touched by this PR)
$certManagerStatus = Wait-ForCertManagerAvailable   # now only waits 2m, not 10m
if ($certManagerStatus -ne $true) {
    throw "cert-manager is not ready..."
}
```

After this PR, that call will wait only 2 minutes before declaring cert-manager unavailable and **throwing during installation**. On a loaded CI node — exactly the environment these fixes target — cert-manager can take longer than 2 minutes to initialise. This could cause fresh installations to fail where they previously succeeded.

**Fix:** Either pass the appropriate timeout explicitly in `Install-CertManagerControllers`, or keep the default at `10m` and only lower it in the callers that intentionally want a short probe.

---

### 🟡 Excessive diagnostic kubectl calls in `Wait-ForCertificateReady`

**File:** `addons/security/security.module.psm1`

The diagnostic block lacks a rate-limit guard:

```powershell
# After 60s of empty status, log cert-manager controller state for diagnostics
if ($elapsed -ge 60 -and [string]::IsNullOrEmpty($readyStatus)) {
    $controllerPods = &"$KubeToolsPath\kubectl.exe" get pods ...   # runs every 5s
    $certDesc       = &"$KubeToolsPath\kubectl.exe" get certificate ... # runs every 5s
}
```

With a 300-second timeout this block executes on every 5-second loop pass from 60 s onward — up to 48 iterations — issuing 96 kubectl calls purely for diagnostics. A one-shot flag (similar to the existing `$renewAttempted` pattern in the same function) would emit the diagnostics once and stop.

---

### 🟡 Stale-secret fallback can mask cert-manager failures

**File:** `addons/security/security.module.psm1` — `Wait-ForCertificateReady`

```powershell
if ($elapsed -ge 30 -and [string]::IsNullOrEmpty($readyStatus)) {
    $secretData = ...kubectl get secret $CertificateName ... '{.data.tls\.crt}'...
    if (-not [string]::IsNullOrEmpty($secretData)) {
        Write-Log "...Accepting as ready."
        return $true   # ← proceeds with potentially expired/mismatched cert
    }
}
```

The fallback accepts any non-empty `tls.crt` in the secret as a signal that the certificate is ready, regardless of expiry or key match. The scenario this is designed for (cert-manager controller restarted mid-issue) is narrow, but the condition (`readyStatus` is empty) is broad enough to also fire when cert-manager is simply slow. A stale cert accepted here will fail later with x509 or `ErrorKeyMatch` errors — exactly the kind of failure this PR is trying to eliminate. Consider at least logging the certificate's `notAfter` field before accepting it, and/or checking that the secret's `ownerReferences` still points to the current Certificate UID.

---

## Files Reviewed
- ✅ addons/addons.module.psm1
- ✅ addons/security/Disable.ps1
- ✅ addons/security/Get-Status.ps1
- ✅ addons/security/security.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/vm/vm.module.psm1

krotz-dieter:

```json
{
  "baseBranch": "main",
  "confidence": 1.0,
  "headSha": "77e859a4b1b742516a2b27982c8c9936125c7fa1",
  "prNumber": 3020,
  "selectedUsecases": [
    "cluster",
    "addon-metrics",
    "addon-ingress-traefik",
    "functional-generic",
    "addon-security"
  ],
  "selectorVersion": "v1",
  "shortReasons": [
    {
      "reason": "mandatory:cluster",
      "usecase": "cluster"
    },
    {
      "reason": "addon-shared-infra",
      "usecase": "addon-metrics"
    },
    {
      "reason": "addon-shared-infra",
      "usecase": "addon-ingress-traefik"
    },
    {
      "reason": "addon-shared-infra",
      "usecase": "functional-generic"
    },
    {
      "reason": "addon-shared-infra",
      "usecase": "addon-security"
    }
  ],
  "status": "success",
  "timestampUtc": "2026-08-28T11:00:45Z"
}
```

PR #3022, opened 2026-08-28
remove conflicting flannel routes on Loopbackk2s to ensure proper pod connectivity

GermanGLLopez:
remove conflicting flannel routes on Loopbackk2s to ensure proper pod connectivity

### DIFF
--- a/lib/modules/k2s/k2s.node.module/linuxnode/setup/control-plane-node.module.psm1
+++ b/lib/modules/k2s/k2s.node.module/linuxnode/setup/control-plane-node.module.psm1
@@ -406,6 +406,25 @@ function Start-ControlPlaneNodeOnNewVM {
 
     Wait-ForSSHConnectionToLinuxVMViaSshKey
 
+    # Ensure the control plane has a route to the Windows loopback adapter subnet
+    # so that flannel's host-gw backend can install pod subnet routes via the
+    # Windows node's public-ip (on the loopback adapter). Without this route,
+    # flannel fails with "network is unreachable" when the loopback subnet is
+    # not directly connected to the control plane's L2 segment.
+    $setupConfigRoot = Get-RootConfigk2s
+    $loopbackCIDR = $setupConfigRoot.psobject.properties['loopbackAdapterCIDR'].value
+    $kubeSwitchIP = Get-ConfiguredKubeSwitchIP
+    if (-not [string]::IsNullOrWhiteSpace($loopbackCIDR) -and -not [string]::IsNullOrWhiteSpace($kubeSwitchIP)) {
+        Write-Log "Adding transit route to $loopbackCIDR via $kubeSwitchIP on control plane" -Console
+        $addRouteResult = Invoke-CmdOnControlPlaneViaSSHKey -CmdToExecute "sudo ip route replace $loopbackCIDR via $kubeSwitchIP" -IgnoreErrors
+        if ($addRouteResult.Success) {
+            Write-Log "Transit route to $loopbackCIDR via $kubeSwitchIP added successfully on control plane"
+        } else {
+            $routeOutput = ($addRouteResult.Output | Out-String).Trim()
+            Write-Log "WARNING: Failed to add transit route to $loopbackCIDR via $kubeSwitchIP on control plane - $routeOutput"
+        }
+    }
+
     # Initialize-Cni0Interface -VmName $controlPlaneVMHostName -WSL:$WSL
 
     $ipindex = Get-NetIPInterface | Where-Object InterfaceAlias -Like "*$switchname*" | Where-Object AddressFamily -Eq IPv4 | Select-Object -expand 'ifIndex'
--- a/lib/modules/k2s/k2s.node.module/linuxnode/setup/control-plane-node.module.psm1
+++ b/lib/modules/k2s/k2s.node.module/linuxnode/setup/control-plane-node.module.psm1
@@ -411,8 +411,7 @@ function Start-ControlPlaneNodeOnNewVM {
     # Windows node's public-ip (on the loopback adapter). Without this route,
     # flannel fails with "network is unreachable" when the loopback subnet is
     # not directly connected to the control plane's L2 segment.
-    $setupConfigRoot = Get-RootConfigk2s
-    $loopbackCIDR = $setupConfigRoot.psobject.properties['loopbackAdapterCIDR'].value
+    $loopbackCIDR = Get-LoopbackAdapterCIDR
     $kubeSwitchIP = Get-ConfiguredKubeSwitchIP
     if (-not [string]::IsNullOrWhiteSpace($loopbackCIDR) -and -not [string]::IsNullOrWhiteSpace($kubeSwitchIP)) {
         Write-Log "Adding transit route to $loopbackCIDR via $kubeSwitchIP on control plane" -Console
--- a/lib/modules/k2s/k2s.node.module/windowsnode/setup/windows-worker-node.module.psm1
+++ b/lib/modules/k2s/k2s.node.module/windowsnode/setup/windows-worker-node.module.psm1
@@ -531,7 +531,7 @@ function Remove-FlannelConflictingRoutesOnLoopback {
         }
         if ($match -and ($podNetworkPrefixLength % 8) -ne 0) {
             $remainingBits = $podNetworkPrefixLength % 8
-            $mask = [byte](0xFF -shl (8 - $remainingBits))
+            $mask = [byte]((0xFF -shl (8 - $remainingBits)) -band 0xFF)
             if (($destBytes[$fullBytes] -band $mask) -ne ($podNetworkBytes[$fullBytes] -band $mask)) { $match = $false }
         }
         return $match
--- a/lib/modules/k2s/k2s.node.module/windowsnode/setup/windows-worker-node.module.psm1
+++ b/lib/modules/k2s/k2s.node.module/windowsnode/setup/windows-worker-node.module.psm1
@@ -211,6 +211,12 @@ function Start-WindowsWorkerNode {
     Start-ServiceAndSetToAutoStart -Name 'containerd'
     Start-ServiceAndSetToAutoStart -Name 'httpproxy'
     Confirm-LoopbackAdapterIP
+
+    # Remove broken flannel routes on Loopbackk2s BEFORE starting flanneld.
+    # Must run first because flanneld's host-gw backend actively maintains routes
+    # and would re-add them if already running when we try to remove.
+    Remove-FlannelConflictingRoutesOnLoopback
+
     # Flanneld must remain SERVICE_DEMAND_START. It creates cbr0 L2Bridge when it
     # doesn't find one, which races with Start-System.ps1 after unclean reboots.
     # Start it explicitly here (after cbr0 is created) without changing the start type.
@@ -220,12 +226,6 @@ function Start-WindowsWorkerNode {
     Start-ServiceAndSetToAutoStart -Name 'kubeproxy'
 
     Wait-NetworkL2BridgeReady -PodSubnetworkNumber $PodSubnetworkNumber
-
-    # Remove broken flannel routes on Loopbackk2s. Flannel's host-gw backend may
-    # add routes for remote subnets via Loopbackk2s, but the gateway IP lives on a
-    # different L2 segment (WSL or KubeSwitch). These lower-metric routes shadow the
-    # correct k2s static routes and break cross-node pod connectivity.
-    Remove-FlannelConflictingRoutesOnLoopback
 
     Add-HostBridgeIpReservation -PodSubnetworkNumber $PodSubnetworkNumber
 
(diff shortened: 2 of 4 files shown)


krotz-dieter (commented):
Now I have enough context to write the review.

Now I have full context. Here's my review:

---

## Code Review: PR #3022

### `lib/scripts/k2s/system/startup/Start-System.ps1`

**Race condition in `Start-K8sNetworkingServices`** (`Start-System.ps1`, around line 388–396):

`flanneld` is started *before* `Remove-FlannelConflictingRoutesOnLoopback` is called:
```powershell
Start-Service -Name 'flanneld' -ErrorAction SilentlyContinue
Start-Service -Name 'kubelet' -ErrorAction SilentlyContinue
Start-Service -Name 'kubeproxy' -ErrorAction SilentlyContinue
Remove-FlannelConflictingRoutesOnLoopback   # ← flanneld already running
```

Flannel's host-gw backend actively maintains routes — it watches the Kubernetes API for node updates and will re-install routes when it detects them missing. Removing routes *after* flanneld is running may create a transient fix that flannel immediately undoes. The same timing issue exists in `Start-WindowsWorkerNode` in `windows-worker-node.module.psm1` — the cleanup runs after `Wait-NetworkL2BridgeReady`, which itself waits for flannel to bring up the bridge. The route removal should ideally happen **before** flanneld is started, or after it is confirmed that flannel has settled and won't re-add the route.

---

### `lib/modules/k2s/k2s.node.module/windowsnode/setup/windows-worker-node.module.psm1`

**`Remove-FlannelConflictingRoutesOnLoopback` only targets `podNetworkMasterCIDR`** (line ~511):

```powershell
$clusterCIDRMaster = $setupConfigRoot.psobject.properties['podNetworkMasterCIDR'].value
...
$conflictingRoutes = Get-NetRoute -DestinationPrefix $clusterCIDRMaster -InterfaceIndex $loopbackIfIndex ...
```

The docstring says this removes "routes for remote subnets" added by flannel, but the implementation only removes a route for the Linux master pod CIDR. If the cluster includes additional Linux worker nodes, flannel installs separate pod subnet routes for each of them on the Loopbackk2s interface. Those routes would also be topologically invalid but are silently left in place. This is a correctness gap given the stated intent of the function.

---

### `test/test.module.psm1`

The `AddRange` → `foreach Add` change is a correct fix. `ArrayList.AddRange` expects `ICollection`; when `ConvertFrom-Json` returns a single string (e.g., a one-element JSON array in PS 5.1), `AddRange` would incorrectly treat it as `IEnumerable<char>`, adding individual characters. The `@($result)` wrapping ensures single items are correctly boxed as a one-element array.

---

### `lib/modules/k2s/k2s.node.module/linuxnode/setup/control-plane-node.module.psm1`

No issues. Config keys (`loopbackAdapterCIDR`, `Get-ConfiguredKubeSwitchIP`) are confirmed present. `ip route replace` is idempotent. The `.Success` property on the return value of `Invoke-CmdOnControlPlaneViaSSHKey` is confirmed by multiple existing usages in the same codebase.

---

## Files Reviewed
- ✅ `lib/modules/k2s/k2s.node.module/linuxnode/setup/control-plane-node.module.psm1`
- ✅ `lib/modules/k2s/k2s.node.module/windowsnode/setup/windows-worker-node.module.psm1`
- ✅ `lib/scripts/k2s/system/startup/Start-System.ps1`
- ✅ `test/test.module.psm1`

krotz-dieter (commented):
Now I have all the information needed for a thorough review.

---

## PR Review: remove conflicting flannel routes on Loopbackk2s

### `lib/modules/k2s/k2s.node.module/windowsnode/setup/windows-worker-node.module.psm1`

**Bug: `[byte]` overflow for non-octet-aligned pod CIDRs**

In `Remove-FlannelConflictingRoutesOnLoopback`, the remainder-bit mask calculation will throw a terminating `InvalidCastException` for any `podNetworkCIDR` with a prefix length not divisible by 8 (e.g. `/10`, `/14`, `/18`, `/22`):

```powershell
$mask = [byte](0xFF -shl (8 - $remainingBits))
```

PowerShell's `-shl` operates on `Int32`, so `0xFF -shl 2` = `1020`, which is out of range for `[byte]` and throws "Cannot convert value '1020' to type 'System.Byte'". This error propagates out of the `Where-Object` scriptblock, leaving `$conflictingRoutes` unassigned and potentially stopping the startup flow entirely.

Fix:
```powershell
$mask = [byte]((0xFF -shl (8 - $remainingBits)) -band 0xFF)
```

This works for all common k2s configs with `/8`, `/16`, or `/24` pod CIDRs (where `$remainingBits` is always 0 and the branch is never entered), but is a latent defect for any non-standard configuration.

---

**Minor: `$null` check vs empty collection on `$conflictingRoutes`**

The `if ($null -ne $conflictingRoutes)` check is consistent with PowerShell's behavior (`Where-Object` returns `$null` when no items match), so this is fine. The logic is correct.

---

### `lib/modules/k2s/k2s.node.module/linuxnode/setup/control-plane-node.module.psm1`

The new transit route block is correct. `Invoke-CmdOnControlPlaneViaSSHKey` does return `[pscustomobject]@{ Success = ...; Output = ... }` (confirmed via source), so `$addRouteResult.Success` is valid. The null-guard on `$loopbackCIDR` and `$kubeSwitchIP` is appropriate.

Note: `ip route replace` without `proto static` means this route is **not persistent** across control-plane reboots. This appears intentional since `Start-ControlPlaneNodeOnNewVM` is called during startup, but if the route needs to survive a Linux reboot independently, it may need to be made persistent (e.g., via a `post-up` rule or netplan config). Worth confirming whether this is called on every startup or only on first setup.

---

### `lib/scripts/k2s/system/startup/Start-System.ps1`

`Remove-FlannelConflictingRoutesOnLoopback` is available here because `k2s.node.module.psm1` recursively imports all child `.psm1` files (including `windows-worker-node.module.psm1`). No issue.

---

### `test/test.module.psm1`

The `AddRange` → `foreach`+`Add` fix is correct. `ConvertFrom-Json` can return a single object (not an array) for a single-element JSON array in some PowerShell versions; wrapping in `@(...)` before iterating ensures `Add` is called per element regardless.

---

## Files Reviewed
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/setup/control-plane-node.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/setup/windows-worker-node.module.psm1
- ✅ test/test.module.psm1

krotz-dieter (commented):
Now let me find `Get-L2BridgeName` and `Invoke-CmdOnControlPlaneViaSSHKey`:

Let me check the `Invoke-CmdOnControlPlaneViaSSHKey` return type with `-IgnoreErrors`:

Let me check the control-plane startup function context:

Now I have enough context to write the full review.

Here is my review of PR #3022:

---

## Review

### `lib/modules/k2s/k2s.node.module/windowsnode/setup/windows-worker-node.module.psm1`

**Bug: Wildcard adapter match catches `vEthernet (Loopbackk2s)` in addition to `Loopbackk2s`**

In `Remove-FlannelConflictingRoutesOnLoopback`:

```powershell
$adapterName = Get-L2BridgeName          # e.g. "Loopbackk2s"
$adapter = Get-NetAdapter -Name "*$adapterName*" -ErrorAction SilentlyContinue
```

The glob pattern `*Loopbackk2s*` also matches **`vEthernet (Loopbackk2s)`** — the vSwitch adapter created by `New-ExternalSwitch` when binding the cbr0 L2Bridge to the loopback adapter. `New-ExternalSwitch` runs before `Remove-FlannelConflictingRoutesOnLoopback` is called in `Start-WindowsWorkerNode`, so `vEthernet (Loopbackk2s)` is guaranteed to exist at call time.

When `$adapter` is a two-element array, `$loopbackIfIndex` becomes an array and `Get-NetRoute` returns routes from **both** interfaces. Routes on `vEthernet (Loopbackk2s)` that happen to fall within `podNetworkCIDR` — in particular the local Windows pod subnet route managed by the HNS L2Bridge network — will be silently removed. This breaks local pod-to-pod connectivity for pods on the Windows node itself.

**Fix:** Use an exact-name match (which is what every other caller of `Get-L2BridgeName` does):

```powershell
$adapter = Get-NetAdapter -Name $adapterName -ErrorAction SilentlyContinue
```

`Get-L2BridgeName` already returns the precise adapter name, so no wildcard is needed.

---

### `lib/modules/k2s/k2s.node.module/linuxnode/setup/control-plane-node.module.psm1`

The new transit route (`sudo ip route replace`) is added each time `Start-ControlPlaneNodeOnNewVM` is called (i.e., every `k2s start`). The approach is idempotent and functionally correct for the normal lifecycle. `Invoke-CmdOnControlPlaneViaSSHKey -IgnoreErrors` correctly returns `[pscustomobject]@{ Success = ...; Output = ... }` — the `.Success` / `.Output` access is valid. No issue here.

---

### `lib/scripts/k2s/system/startup/Start-System.ps1`

`Remove-FlannelConflictingRoutesOnLoopback` is reachable via the imported `k2s.node.module` (which recursively imports `windows-worker-node.module.psm1` and it is exported). No issue here beyond the wildcard bug above.

---

### `test/test.module.psm1`

The fix is correct. `System.Collections.ArrayList.AddRange` requires an `ICollection`, and when `ConvertFrom-Json` returns a single string (scalar, not an array), calling `AddRange` on it fails at runtime because `String` does not implement `ICollection`. Wrapping with `@(...)` and iterating with `Add` correctly handles both single-value and array results.

---

## Files Reviewed
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/setup/control-plane-node.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/setup/windows-worker-node.module.psm1
- ✅ test/test.module.psm1

krotz-dieter (commented):
Based on my analysis, here's my review of PR #3022:

---

## Review

### `lib/modules/k2s/k2s.node.module/linuxnode/setup/control-plane-node.module.psm1`

**Duplicated helper pattern:** The new code accesses `$setupConfigRoot.psobject.properties['loopbackAdapterCIDR'].value` directly, while the already-existing `Restore-ControlPlaneTransitRoute` in `Start-System.ps1` correctly uses `Get-LoopbackAdapterCIDR` (the dedicated helper). Both do the same thing, but if the config key ever changes, this new code won't benefit from the single-place fix. Consider using `Get-LoopbackAdapterCIDR` here.

**Verified:** `Invoke-CmdOnControlPlaneViaSSHKey` with `-IgnoreErrors` does return `[pscustomobject]@{ Success = ...; Output = ... }`, so the `.Success` / `.Output` access is correct. ✓

### `lib/modules/k2s/k2s.node.module/windowsnode/setup/windows-worker-node.module.psm1`

**Actual bug — wrong config accessor for `podNetworkCIDR`:** In `Remove-FlannelConflictingRoutesOnLoopback`:
```powershell
$podNetworkCIDR = $setupConfigRoot.psobject.properties['podNetworkCIDR'].value
```
The existing module-level abstraction for this value is `Get-ConfiguredClusterCIDR` (defined in `config.module.psm1`, which pre-reads `$clusterCIDR` from the same key). Using raw `psobject.properties` access is fragile and inconsistent with the rest of the codebase. This isn't a crash bug (the guard `[string]::IsNullOrWhiteSpace` will catch a null), but it breaks the intended abstraction.

**Missing newline at end of file:** The diff shows `\ No newline at end of file` on the last line. This should be fixed.

**Logic verified correct:**
- `Get-L2BridgeName` (in `loopbackadapter.module.psm1`) finds the Microsoft KM-TEST Loopback Adapter — using it to locate `Loopbackk2s` in `Remove-FlannelConflictingRoutesOnLoopback` is correct. ✓
- The CIDR subnet-matching logic (byte-by-byte with partial byte mask) is correct for IPv4 addresses in network byte order. ✓
- The `$null -ne $conflictingRoutes` guard handles the PowerShell pipeline-null case correctly. ✓
- The function is exported and reachable from `Start-System.ps1` via the `k2s.node.module` recursive import. ✓

### `lib/scripts/k2s/system/startup/Start-System.ps1`

The call to `Remove-FlannelConflictingRoutesOnLoopback` before `Start-Service 'flanneld'` is well-placed and correctly sequenced. ✓

Note: The transit route persistence concern (Linux `ip route replace` doesn't survive reboots) is already handled by the pre-existing `Restore-ControlPlaneTransitRoute` function, called from `Restart-FlannelDaemonSetWithWindowsRouteRepair` on each restart. ✓

### `test/test.module.psm1`

The `AddRange` → `foreach`+`Add` fix is correct. When the JSON result is a scalar (single label), PowerShell deserializes it to a non-array, and `AddRange` requires `IEnumerable<T>` which a scalar doesn't implement. Wrapping with `@()` ensures enumerable iteration. ✓

---

## Files Reviewed

- ✅ lib/modules/k2s/k2s.node.module/linuxnode/setup/control-plane-node.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/setup/windows-worker-node.module.psm1
- ✅ test/test.module.psm1

krotz-dieter (commented):
Now I have enough information to write the review.

Here is my review:

---

## Review: PR #3022 — Remove conflicting flannel routes on Loopbackk2s

### `windows-worker-node.module.psm1` — `Remove-FlannelConflictingRoutesOnLoopback`

**Confirmed correct:**
- `Get-L2BridgeName` does return the loopback adapter name (verified: it looks up adapters by `InterfaceDescription -like 'Microsoft KM-TEST Loopback Adapter*'`), so the adapter lookup is correct.
- The CIDR subnet-containment logic is correct for all prefix lengths, including those that don't fall on a byte boundary.
- The `$null` check on `$conflictingRoutes` is correct: PowerShell 5.1 `Where-Object` returns `$null` when no items match, so the else-branch ("No conflicting routes found") triggers properly.

**Issue — missing newline at end of file:**
The diff shows `\ No newline at end of file` on the last line (`Set-RoutesToWindowsWorkloads`). This can cause issues with some tools and diff/merge operations. A trailing newline should be added.

**Minor concern — route cleanup is not scoped to flannel-added routes:**
The function removes *all* routes on the loopback interface whose destination is within `podNetworkCIDR`, regardless of next-hop. If a valid static route within that CIDR were ever manually added on the loopback, it would be silently removed. The current behavior is documented in the comment and is intentional, but it's worth noting the broad scope.

---

### `control-plane-node.module.psm1`

**Confirmed correct:**
- `Invoke-CmdOnControlPlaneViaSSHKey` returns `[pscustomobject]@{ Success = $success; Output = $output }` (verified in `vm.module.psm1` line 440), so `$addRouteResult.Success` and `$addRouteResult.Output` are valid.
- `Get-ConfiguredKubeSwitchIP` exists and is used elsewhere in the same file.

**Issue — route is not persistent across reboots:**
`sudo ip route replace` creates an in-memory route that is lost on control-plane VM reboot. However, since this code is in `Start-ControlPlaneNodeOnNewVM` (called on every startup), the route is re-added each time the system starts. This is acceptable *only if* this function is always called before anything depends on the route. If the VM is rebooted independently without going through the full startup path, the route will be absent. Consider whether the route should be written to `/etc/network/interfaces` or a `networkd` drop-in for reliability.

---

### `Start-System.ps1`

**Confirmed correct:**
- `Remove-FlannelConflictingRoutesOnLoopback` is available here: `Start-System.ps1` imports `k2s.node.module.psm1`, which recursively imports all sub-modules including `windows-worker-node.module.psm1`, which exports the function.
- Calling cleanup before `Start-Service flanneld` is the correct ordering.

---

### `test/test.module.psm1`

**Fix is correct:**
The original `$foundLabels.AddRange($result)` had a bug: when `$result` is a single `string` (not an array), PowerShell resolves `String` as `IEnumerable<char>`, causing `AddRange` to explode the string into individual characters. The replacement `foreach ($label in @($result))` correctly handles both single-item and multi-item cases.

---

## Files Reviewed
- ✅ lib/modules/k2s/k2s.node.module/linuxnode/setup/control-plane-node.module.psm1
- ✅ lib/modules/k2s/k2s.node.module/windowsnode/setup/windows-worker-node.module.psm1
- ✅ test/test.module.psm1

krotz-dieter:

```json
{
  "baseBranch": "main",
  "confidence": 0.8792,
  "headSha": "cf68ad92ac885606c126c4ab9263323133e364bd",
  "prNumber": 3022,
  "selectedUsecases": [
    "cluster",
    "add-linux-node-online",
    "startstop",
    "reboot"
  ],
  "selectorVersion": "v1",
  "shortReasons": [
    {
      "reason": "prefix:system-scripts",
      "usecase": "cluster"
    },
    {
      "reason": "keyword:node",
    },
    {
      "reason": "prefix:system-scripts",
      "usecase": "startstop"
    },
    {
      "reason": "prefix:system-scripts",
      "usecase": "reboot"
    }
  ],
  "status": "success",
  "timestampUtc": "2026-08-28T11:00:43Z"
}
```

PR #3027, opened 2026-08-28
enhance SSH connection error handling and diagnostics 

GermanGLLopez:
enhance SSH connection error handling and diagnostics in Start-LinuxWorkerNodeServices

### DIFF
--- a/lib/scripts/worker/linux/common/LinuxWorkerNode.Common.ps1
+++ b/lib/scripts/worker/linux/common/LinuxWorkerNode.Common.ps1
@@ -348,9 +348,15 @@ function Invoke-LinuxWorkerNodeStart {
     }
 
     if ($null -ne $nodeConfig -and -not [string]::IsNullOrWhiteSpace($nodeUserName) -and -not [string]::IsNullOrWhiteSpace($nodeConfig.IpAddress)) {
-        Start-LinuxWorkerNodeServices -NodeName $workerNodeName -IpAddress $nodeConfig.IpAddress -UserName $nodeUserName -WaitForReady:$WaitForReady -LogPrefix $LogPrefix
+        try {
+            Start-LinuxWorkerNodeServices -NodeName $workerNodeName -IpAddress $nodeConfig.IpAddress -UserName $nodeUserName -WaitForReady:$WaitForReady -LogPrefix $LogPrefix
+        }
+        catch {
+            Write-Log "$LogPrefix Error starting services on node '$workerNodeName': $($_.Exception.Message)" -Console
+            throw
+        }
     } else {
-        Write-Log "$LogPrefix Cannot start kubelet/runtime on node '$workerNodeName' because Username/IpAddress is missing in cluster config."
+        Write-Log "$LogPrefix Cannot start kubelet/runtime on node '$workerNodeName' because Username/IpAddress is missing in cluster config." -Console
     }
 }
 
--- a/lib/scripts/worker/linux/common/LinuxWorkerNode.Common.ps1
+++ b/lib/scripts/worker/linux/common/LinuxWorkerNode.Common.ps1
@@ -210,7 +210,7 @@ function Start-LinuxWorkerNodeServices {
     }
 
     if (-not $sshProbeSucceeded) {
-        Write-Log "$LogPrefix Failed to establish SSH connection to '$workerNodeName' after retries: $($sshProbeResult.Output)"
+        Write-Log "$LogPrefix Failed to establish SSH connection to '$workerNodeName' after retries: $($sshProbeResult.Output)" -Console
 
         try {
             $tcpCheck = Test-NetConnection -ComputerName $IpAddress -Port 22 -WarningAction SilentlyContinue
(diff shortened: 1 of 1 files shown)


krotz-dieter (commented):
Based on my review of the source file (decoded the full base64 content), here are my findings:

---

## Review: `lib/scripts/worker/linux/common/LinuxWorkerNode.Common.ps1`

### Breaking Change: `return` → `throw` propagation through `Invoke-LinuxWorkerNodeStart`

The wrapper function `Invoke-LinuxWorkerNodeStart` (within the same file, visible in the decoded source) calls `Start-LinuxWorkerNodeServices` **without a try/catch**:

```powershell
function Invoke-LinuxWorkerNodeStart {
    ...
    Start-LinuxWorkerNodeServices -NodeName $workerNodeName -IpAddress ... -WaitForReady:$WaitForReady ...
    # No try/catch around this call
}
```

Changing three `return` paths to `throw` inside `Start-LinuxWorkerNodeServices` will cause unhandled exceptions to bubble through `Invoke-LinuxWorkerNodeStart` to external callers. If those callers (start scripts using `--node` flag) don't have `try/catch`, the exception could terminate the entire script rather than producing a graceful error. This needs to be verified against all callers or `Invoke-LinuxWorkerNodeStart` needs a try/catch wrapper.

### Hardcoded timeout in throw message

The new throw message hardcodes `"60 seconds"`:
```
throw "... did not transition to Ready within 60 seconds after starting services"
```

This value is derived from `$maxRetries = 30` × `Start-Sleep -Seconds 2`. It's correct now, but it's fragile — if either constant changes, the error message silently becomes wrong. Consider computing it: `"within $($maxRetries * 2) seconds"`.

### Everything Else Looks Good

- The `-Console` flag additions improve user visibility for failures ✓
- Diagnostics collection is properly wrapped in try/catch ✓
- `($diagResult.Output | Out-String).Trim()` is null-safe in PowerShell ✓
- The SSH failure diagnostics block (TCP check, VM adapter inspection) is unchanged and still correct ✓
- `$errorOutput` reuse in both the log message and the throw is consistent ✓

---

## Files Reviewed
- ✅ `lib/scripts/worker/linux/common/LinuxWorkerNode.Common.ps1`

krotz-dieter:

```json
{
  "baseBranch": "main",
  "confidence": 0.8042,
  "headSha": "dab832911cb63d7220ced64ea69d795edea2f097",
  "prNumber": 3027,
  "selectedUsecases": [
    "add-linux-node-online",
    "cluster"
  ],
  "selectorVersion": "v1",
  "shortReasons": [
    {
      "reason": "keyword:node",
    },
    {
      "reason": "mandatory:cluster",
      "usecase": "cluster"
    }
  ],
  "status": "success",
  "timestampUtc": "2026-08-29T06:00:02Z"
}
```

krotz-dieter (commented):
Now I have the original source. Let me analyze the full picture.

Based on the full file from `main` and the diff, here's my review:

---

## Review: `lib/scripts/worker/linux/common/LinuxWorkerNode.Common.ps1`

### Bug: `Invoke-LinuxWorkerNodeStart` not updated to handle new exceptions

The `Invoke-LinuxWorkerNodeStart` wrapper (lines ~285–300 in the original file) calls `Start-LinuxWorkerNodeServices` with no `try/catch`:

```powershell
# No exception handling — all new throws propagate unhandled
Start-LinuxWorkerNodeServices -NodeName $workerNodeName -IpAddress ... -WaitForReady:$WaitForReady ...
```

The PR converts three `return` (silent-fail) paths into `throw`:
1. SSH connection failure after retries (line ~213)
2. `WaitForReady` timeout in the success path (line ~264)
3. `WaitForReady` timeout in the kubelet-already-active path (line ~303)

Before this change, all failure paths exited quietly via `return`. Any caller of `Invoke-LinuxWorkerNodeStart` that doesn't wrap it in `try/catch` will now receive unhandled terminating exceptions where it previously got a silent no-op. The wrapper itself must be updated to either (a) propagate with proper documentation, or (b) catch and handle the new exceptions consistently — especially the SSH-failure case, which previously was explicitly a soft failure with diagnostics-only logging.

### Minor: `WaitForReady` timeout in the success path silently changed contract

In the original code, `WaitForReady` timeout just logged and the function returned normally — callers had no way to detect timeout as a failure. The new `throw` is the right direction, but this contract change is implicit. If any callers rely on `Start-LinuxWorkerNodeServices` always completing without exceptions in the success path (i.e., services started but node didn't become Ready in time), those callers will now break. This is tightly linked to the `Invoke-LinuxWorkerNodeStart` issue above.

### Looks correct

- The kubelet-already-active verification logic in the `else` branch is sound — it checks `systemctl is-active kubelet` to avoid false negatives from non-zero `systemctl start` exit codes when services are already running.
- The diagnostic collection on SSH failure and service failure is a good addition.
- The `$totalTimeoutSeconds` variable replacing the hardcoded `~60 seconds` in the log message is correct.
- The `$retryIntervalSeconds` variable is correctly referenced via `$totalTimeoutSeconds = $maxRetries * $retryIntervalSeconds`.

---

## Files Reviewed
- ✅ `lib/scripts/worker/linux/common/LinuxWorkerNode.Common.ps1`

Issue #3044, opened 2026-09-01
[Reverted Setup Linux] Provision an optional Windows worker on a Linux host


krotz-dieter:
## Problem

The documented Linux-host variant describes a native Linux control plane with an optional Windows worker VM managed through libvirt/KVM. The actual install flow currently rejects every Linux installation that does not use `--linux-only`.

The libvirt XML templates and Linux setup-orchestration building blocks exist, but they are not wired into an end-to-end Windows worker provisioning lifecycle.

## Proposed Solution

Extend the native Linux install flow to support a mixed-OS cluster:

1. Create and configure the libvirt network from the shipped template.
2. Create the Windows worker VM and its QCOW2 disk from the domain template.
3. Bootstrap the Windows worker, establish secure remote connectivity, and join it to the Kubernetes cluster.
4. Manage the worker during `k2s start`, `k2s stop`, status, and uninstall.
5. Clean up libvirt resources and temporary artifacts reliably after failures and uninstall.

Keep `--linux-only` as the lightweight native control-plane mode. Do not change the Windows-host setup path.

## Relevant Files

- `cmd/k2s/cmd/install/install.go`
- `internal/setuporchestration/orchestrator_linux.go`
- `internal/setuporchestration/vmmanager_linux.go`
- `internal/setuporchestration/network_linux.go`
- `internal/setuporchestration/winimage_linux.go`
- `internal/setuporchestration/libvirt_template_linux.go`
- `cfg/libvirt/domain.xml.tmpl`
- `cfg/libvirt/network.xml.tmpl`

## Acceptance Criteria

- A Linux host supports both `k2s install --linux-only` and a Linux-host installation with a Windows worker VM.
- The Windows VM is provisioned through the shipped, user-customizable libvirt templates.
- The Windows worker joins the cluster and becomes `Ready`.
- `k2s start`, `stop`, `status`, and uninstall handle the Windows VM correctly.
- Failed installation and uninstall remove libvirt domains, networks, disks, and temporary bootstrap resources owned by K2s.
- Integration tests cover Linux-only and mixed Linux-host/Windows-worker installation paths.
- Documentation accurately describes prerequisites, limitations, and cleanup behavior.

PR #3089, opened 2026-09-08
#3044 finalize the linux host parts for linux only

krotz-dieter:
#3044 finalize the linux host parts for linux only


krotz-dieter (commented):
I'll review this PR systematically, analyzing each changed file for bugs, logic errors, security vulnerabilities, and missing error handling.

Let me start by examining the source code context for key files to verify the changes.

Let me try using the public GitHub API:

I'll proceed with analyzing the provided diff directly. Let me examine each file systematically:

## Analysis of PR #3089 Changes

### 1. **internal/linux/lifecycle/dispatcher.go** - Go file security and validation

**Issues Found:**

- **Line 18-27: Lock release logic change** - The modified `k2s_lock_and_run` function now correctly releases the lock and file descriptor after command execution. However, examining the lifecycle.sh diff more closely:

In `lib/modules/linux/cluster/lifecycle.sh`, lines 30-34 show:
```bash
k2s_lock_and_run() {
  exec 9>"$K2S_CONFIG_DIR/lifecycle.lock"
  flock -w 300 9 || { k2s_log ERROR 'Timed out waiting for another K2s lifecycle operation.'; return 1; }
  "$@"
  local exit_code=$?
  flock -u 9
  exec 9>&-
  return "$exit_code"
}
```

This is **correct** - it now properly captures and returns the exit code while cleaning up resources.

### 2. **lib/modules/linux/cluster/lifecycle.sh** - Shell script for lifecycle management

**Issues Found:**

- **Line 12-13: Operation file permissions verification** - Good defensive check: `[[ $(stat -c '%a' "$2") == 600 ]]`. 

- **Line 14-24: jq validation improvements** - The PR adds comprehensive schema validation before processing JSON. However, there's a potential issue:
  - Line 15: The jq validation on line 15 is strict but the individual jq extractions on lines 16-24 now have `|| return 2` which is good for error handling.
  - **Line 20**: `K2S_CONTROL_PLANE_HOSTNAME=${K2S_CONTROL_PLANE_HOSTNAME:-$(hostname)}` - **Good defensive fallback**, but this modifies the extracted value. This is acceptable behavior change.

- **Line 26-36: Lock handling** - The new lock management properly releases lock and handles exit codes correctly (previously it didn't explicitly release the lock).

- **Line 159-165: Logging stream changes** - The addition of log streaming redirection:
```bash
export K2S_LOG_STREAMED=true
exec > >(tee -a "$K2S_LOG_FILE") 2>&1
```
This redirects stdout/stderr to both terminal and log file. This could cause issues because:
  - The `K2S_LOG_STREAMED` flag is exported, and in `logging.sh` it's checked to **avoid double-logging**. This is good.
  - However, this affects the entire script execution - any output (including from sourced files) will be tee'd.

### 3. **lib/modules/linux/common/command.sh** - Command execution wrapper

**Issues Found:**

- **Line 15-27: Pipeline status handling** - The change adds proper error handling for `tee`. The original code had:
```bash
if "$@" 2>&1 | tee -a "$K2S_LOG_FILE"; then
  return 0
else
  local exit_code=${PIPESTATUS[0]}
  ...
fi
```

The problem with the original: `${PIPESTATUS[0]}` is **unreliable** when used after `else`. The correct fix here is good - it uses `set -o pipefail` and captures `${PIPESTATUS[@]}` immediately after the pipeline.

- However, there's a **subtle issue**: Line 17-19 check the exit code AFTER the if block, but the `PIPESTATUS` array is captured and used correctly on lines 23-24. This is **correct**.

### 4. **lib/modules/linux/common/logging.sh** - Logging function

**Issue Found:**

- **Line 24-26: Conditional logging** - The check `if [[ ${K2S_LOG_STREAMED:-false} != true ]]; then` is meant to avoid double-logging when output is already redirected via `tee`. This is **correct** and necessary to prevent duplicate log entries.

### 5. **lib/modules/linux/infra/proxy.sh** - NEW file for proxy configuration

**Issues Found:**

- **Line 16-21: Error handling in k2s_proxy_network_install** - The condition:
```bash
if ip -o -4 addr show | grep -Fq "$address/" && ! ip -o -4 addr show dev k2s-proxy0 | grep -Fq "$address/"; then
  k2s_log ERROR "Configured KubeSwitch address is already used: $address"
  return 5
fi
```
This correctly checks if the address is already in use and fails early. **Good defensive programming**.

- **Line 51-54: APT proxy backup logic** - The code creates a backup of existing apt config before overwriting:
```bash
if [[ -f "$K2S_APT_PROXY_CONFIG" ]] && ! grep -Fq "$K2S_APT_PROXY_HEADER" "$K2S_APT_PROXY_CONFIG"; then
  local backup="$K2S_CONFIG_DIR/$K2S_APT_PROXY_BACKUP"
  [[ ! -e "$backup" ]] || { k2s_log ERROR "Refusing to overwrite existing apt proxy backup: $backup"; return 5; }
  cp "$K2S_APT_PROXY_CONFIG" "$backup" || return 1
```
This is **good** - it prevents overwriting existing backups. However:
  - Line 53: The condition `[[ ! -e "$backup" ]] || { ... return 5; }` means "if backup DOES exist, return 5". This is backward and **confusing**, though technically correct logic.

- **Line 82-88: Restoration logic** - The cleanup properly restores the APT config from backup:
```bash
if [[ -f "$backup" ]]; then
  cp "$backup" "$K2S_APT_PROXY_CONFIG" || k2s_log WARN "Could not restore apt proxy configuration from $backup"
  rm -f "$backup"
else
  rm -f "$K2S_APT_PROXY_CONFIG"
fi
```
This is **correct** - it handles both cases (backup exists or not).

### 6. **lib/modules/linux/networking/dns.sh** - DNS configuration

**Issues Found:**

- **Line 30-37: Resolver state preservation** - The new `k2s_save_resolver` function saves file permissions and immutable flag:
```bash
mode=$(stat -c '%a' /etc/resolv.conf) || return 1
immutable=false
lsattr -d /etc/resolv.conf 2>/dev/null | grep -q 'i' && immutable=true
```
This is **good defensive programming** to preserve system state.

- **Line 39-51: Restoration logic** - The `k2s_restore_resolver` function properly:
  - Removes immutable flag before modifying
  - Handles both symlink and regular file cases
  - Restores original permissions and immutable flag
  
This is **correct and thorough**.

### 7. **lib/modules/linux/node/packages.sh** - Package management

**Issue Found:**

- **Line 30-31: Version extraction** - The code uses `sed` to extract version from PowerShell module:
```bash
version=$(sed -n "/function Get-DefaultK8sVersion/,/^}/s/.*return ['\"]\(v[0-9][0-9.]*\)['\"].*/\1/p" ...)
```
This regex pattern looks for `return 'vX.Y.Z'` or `return "vX.Y.Z"` in PowerShell. **This is fragile** if the format changes, but acceptable for a specific codebase requirement.

### 8. **lib/scripts/linux/debian/linuxonly/Status.sh** - Status script

**BREAKING CHANGE FOUND:**

```diff
-source "$K2S_INSTALL_DIR/lib/modules/linux/cluster/lifecycle.sh"
-
-k2s_dispatch_lifecycle status status "$@"
+
+k2s_initialize_logging
+k2s_require_install_dir || { k2s_log ERROR 'Native K2s installation directory is invalid.'; exit 1; }
+k2s_log INFO 'Dispatching native Debian 13 status through the K2s CLI.'
+exec "$K2S_INSTALL_DIR/k2s" status "$@"
```

This **delegates to the K2s CLI** instead of using `k2s_dispatch_lifecycle`. This is a significant behavioral change:
- **Before**: Dispatched directly using the lifecycle system with `status status` parameters
- **After**: Delegates to the K2s CLI via `exec`

This is a **breaking change** but may be intentional. It depends on whether the K2s CLI's `status` command properly handles the native Linux setup.

### 9. **lib/modules/linux/cluster/lifecycle.sh** - Major changes

**Issue Found - Line 152:**

```bash
k2s_uninstall_cluster() { ... rm -f /etc/systemd/system/k2s-{httpproxy,dnsproxy,proxy-network}.service /etc/apt/apt.conf.d/proxy.conf /etc/systemd/system/kubelet.service.d/20-k2s-logging.conf; ... }
```

Changed to:

```bash
k2s_uninstall_cluster() { ... rm -f /etc/systemd/system/k2s-dnsproxy.service /etc/k2s/dnsproxy.yaml /etc/systemd/system/kubelet.service.d/20-k2s-logging.conf; k2s_proxy_cleanup; ... }
```

The difference is:
- **Old**: Removed systemd service files directly via globbing
- **New**: Calls `k2s_proxy_cleanup` function which is now in `proxy.sh`

**Potential Issue**: The globbing pattern `k2s-{httpproxy,dnsproxy,proxy-network}` was replaced with a single `k2s-dnsproxy.service` file, and the cleanup is delegated to `k2s_proxy_cleanup`. This is a **refactoring that centralizes proxy cleanup** which is good, but requires that the `proxy.sh` module is sourced in all scripts that call `k2s_uninstall_cluster`.

Looking at the Install.sh changes (lines added):
```bash
+source "$K2S_INSTALL_DIR/lib/modules/linux/infra/proxy.sh"
```

This is sourced in Install.sh. Let me check if it's sourced in Stop.sh and Uninstall.sh...

Looking at the diff:
- **Install.sh**: Sources `proxy.sh`, `packages.sh`, `dns.sh` ✓
- **Start.sh**: Sources only `dns.sh` - **NO proxy.sh!** ⚠️
- **Stop.sh**: Sources `dns.sh` - **NO proxy.sh!** ⚠️  
- **Uninstall.sh**: Sources `dns.sh` - **NO proxy.sh!** ⚠️

**BUG FOUND**: The `k2s_proxy_cleanup` function is called in `lifecycle.sh` line 152 (in the `k2s_uninstall_cluster` function), but `lib/modules/linux/infra/proxy.sh` is NOT sourced in the Uninstall.sh, Stop.sh, or Start.sh scripts. This will cause:
- **Runtime error**: `k2s_proxy_cleanup: command not found` when uninstall is executed
- The function is only defined in `proxy.sh`, which isn't sourced in these scripts

However, looking more carefully, Stop.sh also calls `k2s_stop_cluster` which calls functions like `k2s_dns_stop`. Let me check if it's similar... Actually, Stop.sh sources `dns.sh` so `k2s_dns_stop` will be available. But Stop.sh calls `k2s_stop_cluster`, which contains `systemctl stop ... k2s-httpproxy k2s-proxy-network` directly without calling `k2s_proxy_cleanup`, so it doesn't have a function dependency issue there.

The **real issue**: In `lifecycle.sh`, `k2s_uninstall_cluster` calls `k2s_proxy_cleanup`, but this function won't be available unless `proxy.sh` is sourced. Let me verify this is called from Uninstall.sh:

Looking at Uninstall.sh in the diff:
```bash
source "$K2S_INSTALL_DIR/lib/modules/linux/cluster/lifecycle.sh"
k2s_dispatch_lifecycle uninstall "$@"
```

This doesn't source `proxy.sh`, so when `lifecycle.sh`'s `k2s_uninstall_cluster` runs and calls `k2s_proxy_cleanup`, it will **fail with "command not found"**.

**CRITICAL BUG**: Missing `proxy.sh` sourcing in Uninstall.sh, Stop.sh, and Start.sh.

---

## Review Summary

### Critical Issues:
1. **Missing module sourcing** - `lib/modules/linux/infra/proxy.sh` needs to be sourced in Stop.sh, Uninstall.sh (and likely Start.sh) before `lib/modules/linux/cluster/lifecycle.sh`, but it's only sourced in Install.sh.

### Breaking Changes:
2. **Status.sh behavior change** - Now delegates to K2s CLI instead of using native lifecycle dispatcher. This may be intentional but represents changed behavior.

### Minor Issues:
3. **Backward compatibility with lock handling** - The lock release changes in `k2s_lock_and_run` are good but change the timing of lock release. Should verify this doesn't affect concurrency assumptions elsewhere.

---

## Files Reviewed

- ✅ internal/linux/lifecycle/dispatcher.go
- ✅ lib/manifests/clusterip-webhook/deployment.yaml
- ✅ lib/modules/linux/cluster/lifecycle.sh
- ✅ lib/modules/linux/common/command.sh
- ✅ lib/modules/linux/common/logging.sh
- ✅ lib/modules/linux/infra/proxy.sh
- ✅ lib/modules/linux/networking/dns.sh
- ✅ lib/modules/linux/node/packages.sh
- ✅ lib/scripts/linux/debian/linuxonly/Install.sh
- ✅ lib/scripts/linux/debian/linuxonly/Start.sh
- ✅ lib/scripts/linux/debian/linuxonly/Status.sh
- ✅ lib/scripts/linux/debian/linuxonly/Stop.sh
- ✅ lib/scripts/linux/debian/linuxonly/Uninstall.sh
- ✅ test/native-linux-layout.sh

krotz-dieter (commented):
I'll review this PR systematically, examining each changed file for bugs, logic errors, security issues, and regressions.

## Review Analysis

This PR substantially refactors Linux lifecycle operations, moving orchestration logic from Go to Bash shell modules. I'll examine each category of changes:

### 1. Go Changes - dispatcher.go
**Status:** ✅ Good additional validation

The new validation checks in `Execute()` are sound:
- Validates install/config dirs are absolute paths (not root `/`)
- Checks for non-empty strings
- Verifies lifecycle script is a regular file (not symlink/directory)
- **Validates file permissions (0600)** before writing sensitive data - this is excellent security hardening

The permission invariant documentation is clear and warranted.

### 2. Go Changes - New command_linux.go
**Status:** ✅ Appropriate simplification

The new `command_linux.go` provides minimal helper functions for libvirt support. The separation of concerns is appropriate:
- `runCommand()`: For simple commands with combined output
- `runCommandOutput()`: Separates stdout from stderr for parsing

This is a reasonable reduction from the massive `orchestrator_linux.go`.

### 3. Go Changes - Deleted files
**Status:** ⚠️ **CRITICAL: Missing error handling in lifecycle.sh**

The large Go orchestration logic (`orchestrator_linux.go`, `provisioning_linux.go`) has been moved to Bash. This is a significant architectural change, but I need to verify the Bash implementations have proper error handling.

### 4. Bash Changes - lifecycle.sh
**Status:** ⚠️ **ISSUE FOUND: Premature loop break in `k2s_lock_and_run()`**

**Problem in lock handling (lines 30-36):**
```bash
k2s_lock_and_run() {
  exec 9>"$K2S_CONFIG_DIR/lifecycle.lock"
  flock -w 300 9 || { k2s_log ERROR 'Timed out waiting for another K2s lifecycle operation.'; return 1; }
  "$@"
  local exit_code=$?
  flock -u 9      # <-- ISSUE
  exec 9>&-       # <-- ISSUE
  return "$exit_code"
}
```

The `flock -u 9` and `exec 9>&-` should happen in a trap or finally-like context. If `"$@"` exits with failure, the unlock/close happen, but the original code (before this PR) didn't have explicit cleanup. **However**, if the script exits unexpectedly before reaching the explicit cleanup, the lock remains held. The current change adds cleanup, which is good, but it should be more robust with a trap.

**Better pattern:**
```bash
k2s_lock_and_run() {
  exec 9>"$K2S_CONFIG_DIR/lifecycle.lock" || return 1
  flock -w 300 9 || { k2s_log ERROR 'Timed out...'; return 1; }
  trap 'flock -u 9; exec 9>&-' EXIT
  "$@"
}
```

### 5. Bash Changes - lifecycle.sh continued
**Status:** ✅ **k2s_load_operation() improvements**

The validation enhancements are good:
- Validates jq type checking for required fields
- **Problem spotted:** Line 23 sets `K2S_CONTROL_PLANE_HOSTNAME=${K2S_CONTROL_PLANE_HOSTNAME:-$(hostname)}` 
  - This fallback happens **after** the jq assignment, which is correct
  - However, there's a potential issue: if the jq call for `controlPlaneHostname` succeeds but returns an empty string, the default is applied - **this may be intentional** for the empty-string case, so it's acceptable.

**Logging redirection issue (lines 162-163):**
```bash
export K2S_LOG_STREAMED=true
exec > >(tee -a "$K2S_LOG_FILE") 2>&1
```
This redirects all output to both stderr and the log file. **Potential issue:** If `$K2S_LOG_FILE` path is invalid or unwritable, the redirection will fail and cause the entire script to fail. Need to verify `K2S_LOG_FILE` is validated earlier.

### 6. Bash Changes - command.sh
**Status:** ⚠️ **ISSUE FOUND: PIPESTATUS handling bug**

```bash
set -o pipefail
if "$@" 2>&1 | tee -a "$K2S_LOG_FILE"; then
  return 0
fi
pipeline_status=("${PIPESTATUS[@]}")
command_status=${pipeline_status[0]}
tee_status=${pipeline_status[1]}
```

**Bug:** `set -o pipefail` makes the pipeline fail if ANY command fails. After the pipeline, `PIPESTATUS` is captured. However:
- The `if` statement consumes the pipeline's exit status
- After `if` exits, `PIPESTATUS` is updated to the `if` statement's status, not the pipeline's
- This code tries to access the old `PIPESTATUS` but it's already been overwritten

**Correct pattern would be:**
```bash
"$@" 2>&1 | tee -a "$K2S_LOG_FILE"
command_status=${PIPESTATUS[0]}
tee_status=${PIPESTATUS[1]}
if [[ $command_status -eq 0 && $tee_status -eq 0 ]]; then
  return 0
fi
...
```

### 7. Bash Changes - proxy.sh (NEW FILE)
**Status:** ✅ Looks solid

The proxy module handles:
- Network interface setup with proper validation
- APT proxy configuration with backup handling
- Cleanup logic

One minor observation: In `k2s_proxy_install()`, if the backup path exists, it returns error code 5. This is an opinionated error code choice but acceptable for this use case.

### 8. Bash Changes - dns.sh
**Status:** ✅ Good refactoring

Moved DNS handling into a dedicated module. Includes:
- Resolver state save/restore with immutable attribute preservation
- DNS proxy service management
- systemd-resolved integration

### 9. Bash Changes - logging.sh
**Status:** ✅ Appropriate

The check for `K2S_LOG_STREAMED` prevents double-logging when output is already being teed to the log file.

### 10. Deployment.yaml
**Status:** ✅ Minor documentation improvement

Just adds a comment explaining fsGroup=65534.

### 11. Script Updates (Install.sh, Start.sh, Stop.sh, Uninstall.sh, Status.sh)
**Status:** ⚠️ **ISSUE: Status.sh bypasses lifecycle function**

**Status.sh change:**
```bash
# OLD: k2s_dispatch_lifecycle status status "$@"
# NEW:
k2s_initialize_logging
k2s_require_install_dir || { k2s_log ERROR 'Native K2s installation directory is invalid.'; exit 1; }
k2s_log INFO 'Dispatching native Debian 13 status through the K2s CLI.'
exec "$K2S_INSTALL_DIR/k2s" status "$@"
```

This change **completely bypasses** the native Bash lifecycle handler and delegates to the Go CLI tool. While this might be intentional, it's inconsistent with the PR's goal of "finalizing the linux host parts for linux only". Need to verify this is intentional.

## Critical Issues Summary

1. **command.sh PIPESTATUS bug**: The logic for capturing pipeline exit codes is incorrect - it will read the wrong status after the if statement consumes it
2. **logging redirect robustness**: No validation that `$K2S_LOG_FILE` directory exists before redirecting
3. **Lock cleanup**: Should use trap for guaranteed cleanup (minor - current approach works but less robust)
4. **Status.sh delegation**: Inconsistent with PR goal - verify this is intentional

---

## Files Reviewed

- ✅ internal/linux/lifecycle/dispatcher.go
- ✅ internal/setuporchestration/command_linux.go
- ✅ internal/setuporchestration/command_linux_test.go
- ✅ internal/setuporchestration/doc.go
- ✅ internal/setuporchestration/orchestrator_linux.go (deleted)
- ✅ internal/setuporchestration/orchestrator_linux_test.go (deleted)
- ✅ internal/setuporchestration/provisioning_linux.go (deleted)
- ✅ lib/manifests/clusterip-webhook/deployment.yaml
- ✅ lib/modules/linux/cluster/lifecycle.sh
- ✅ lib/modules/linux/common/command.sh
- ✅ lib/modules/linux/common/logging.sh
- ✅ lib/modules/linux/infra/proxy.sh
- ✅ lib/modules/linux/networking/dns.sh
- ✅ lib/modules/linux/node/packages.sh
- ✅ lib/scripts/linux/debian/linuxonly/Install.sh
- ✅ lib/scripts/linux/debian/linuxonly/Start.sh
- ✅ lib/scripts/linux/debian/linuxonly/Status.sh
- ✅ lib/scripts/linux/debian/linuxonly/Stop.sh
- ✅ lib/scripts/linux/debian/linuxonly/Uninstall.sh
- ✅ test/native-linux-layout.sh

PR #3134, opened 2026-09-17
#2829 Fix start issue on windows 10

krotz-dieter:
#2829 Fix start issue on windows 10


krotz-dieter (commented):
I'll review this PR systematically by examining the changes and verifying the context in the actual codebase.

Let me try fetching from the main branch to understand the context:

I see the repository is on GitHub.com, not GitHub Enterprise. Let me try again:

Good, I found the `Set-RoutesToKubemaster` function. Let me get the test file to review the complete context:

Now let me look at the PR directly to see if there are the new test cases in the diff:

Based on my thorough analysis of the PR changes, here's my review:

## PR #3134 Code Review

### Summary
This PR removes the `-AddressFamily IPv4` parameter from two `Get-NetRoute` cmdlet calls in the `Set-RoutesToKubemaster` function and adds comprehensive test coverage for the fix.

---

### Findings

#### ✅ **Issue Fix is Valid**
The fix correctly removes `-AddressFamily IPv4` from `Get-NetRoute` calls (lines 423 and 432 in the main module). The issue appears to be that on Windows 10, this parameter can throw "No MSFT_NetRoute objects found with property 'AddressFamily' equal to 'IPv4'" when certain network conditions exist (e.g., only IPv6 routes present).

Removing the parameter is safe because:
1. The subsequent `Where-Object` filter still correctly matches only IPv4 routes via `DestinationPrefix -eq $ipControlPlaneCIDR` where `$ipControlPlaneCIDR` is a specific IPv4 CIDR (172.19.1.0/24)
2. IPv6 routes won't match this filter, so behavior remains correct
3. Both IPv4 and IPv6 routes are now retrieved, but filtering ensures only relevant IPv4 routes are processed

#### ✅ **Test Coverage is Excellent**
Three new tests are added that validate:
1. **Empty persistent store scenario**: Verifies the code gracefully handles when `Get-NetRoute` is called without the `-AddressFamily` parameter and an empty result set is returned
2. **IPv6-only routes scenario**: Tests that the code correctly creates the on-link route when only IPv6 routes exist (validating the fix works on systems without IPv4 routes initially)
3. **Error propagation**: Ensures genuine query failures are properly surfaced (not swallowed)

The mock patterns confirm the fix design: tests verify that `-AddressFamily` is never passed to `Get-NetRoute`.

#### ⚠️ **Minor Test Observation**
In the new "accepts an empty persistent store during Windows-hosted Linux-only startup" test, the mock returns different data based on `$PolicyStore`:
```powershell
if ($PolicyStore -eq 'ActiveStore') {
    [pscustomobject]@{ DestinationPrefix = '172.19.1.0/24'; ... }
}
```
For `PersistentStore`, it returns nothing (empty). This correctly simulates the Windows 10 scenario where the persistent store might be empty.

#### ✅ **Logic Correctness**
- The function still correctly identifies and preserves the "on-link" route (NextHop = '0.0.0.0')
- Obsolete routes (with NextHop = $windowsHostIpAddress) are still properly removed from both stores
- No regression in existing behavior for systems with IPv4 routes

---

### Conclusion
**This PR is sound.** It correctly addresses the Windows 10 startup issue by removing an overly restrictive parameter that was causing cmdlet failures in certain network configurations. The fix maintains correct behavior through proper filtering, and the new tests provide strong validation that the Windows-hosted Linux-only and IPv6-only scenarios now work correctly.

---

## Files Reviewed
- ✅ lib/modules/windows/node/k2s.node.module/windowsnode/setup/windows-worker-node.module.psm1
- ✅ lib/modules/windows/node/k2s.node.module/windowsnode/setup/windows-worker-node.module.unit.tests.ps1

PR #3142, opened 2026-09-18
fix(node): persist swap disablement for Linux workers

GermanGLLopez:
persist swap disablement for Linux workers


krotz-dieter (commented):
Let me try getting the file from the main branch to understand the existing code structure:

I see the repository is at github.com, not the GHE instance. Let me try with the correct host:

Let me analyze the diff provided directly. I'll examine the code changes for correctness, security issues, and logic errors:

Let me verify the most critical issue by checking if there are any helper functions that handle the output parsing:

Now let me document my findings in a comprehensive review:

## Review Summary: PR #3142

### ⚠️ Critical Issues Found

**1. Broken Swap Unit Parsing (Line 139)**

The line `$swapUnitLines = @($swapUnitsResult.Output)` creates a single-element array containing the entire multi-line output string, rather than splitting lines into separate array items. This breaks the subsequent foreach loop and swap masking logic.

**Fix needed:** `$swapUnitLines = $swapUnitsResult.Output -split '\n'`

---

**2. Inconsistent Error Suppression (Line 330)**

The diagnostic command has inconsistent `2>/dev/null` redirection:
- Missing from `journalctl` and `systemctl show` commands
- Present on `systemctl is-active crio containerd`

This creates inconsistent error output behavior compared to the original version.

**Fix needed:** Add `2>/dev/null` to all commands consistently.

---

### ⚠️ Medium Risk: Breaking Change

**3. Swap Disablement Now Required for Service Startup (Line 254)**

The PR makes `Disable-LinuxWorkerNodeSwap` a hard requirement within `Start-LinuxWorkerNodeServices`. If this function throws an exception, service startup fails entirely. This changes existing behavior and could break deployments that don't expect this dependency.

**Recommendation:** Clarify whether this is intentional or should log warnings instead.

---

## Files Reviewed
- ✅ lib/scripts/windows/worker/linux/common/LinuxWorkerNode.Common.ps1

krotz-dieter (commented):
I'll review this PR carefully, analyzing the code changes for bugs, logic errors, security issues, and potential regressions. Let me start by fetching the source file to verify the context and function signatures.

Let me fetch from the main branch to understand the baseline context and then use the PR information:

I see the repository is on a GitHub Enterprise instance (shs.ghe.com). Let me analyze the provided diff directly since I have the full context. Here's my code review:

## PR #3142 Review: fix(node): persist swap disablement for Linux workers

### Analysis

#### 1. **Swap Disablement Approach Change (Disable-LinuxWorkerNodeSwap)**

**Positive:**
- Changes from a file-deletion approach to a systemd-based masking approach, which is more persistent and reliable
- Adds proper error verification at the end to confirm swap is actually disabled
- Improved sed regex from `/\sswap\s/d` to `/[[:space:]]swap[[:space:]]/d` is more portable

**Potential Issues:**
- **Line 139:** The parsing of `systemctl list-units` output assumes `[0]` index after split gives the unit name. With `--plain` format, this should work, but there's no validation that the split actually produced an array with a first element. If output is malformed, this could fail silently.
- **Line 140:** The condition `if ($swapUnit -like '*.swap')` is defensive but means non-standard swap units won't be masked. This is probably intentional but could miss edge cases.
- **Line 148-149:** The verification check `if (-not [string]::IsNullOrWhiteSpace(...))` throws an error if swap still exists. This is good, but the error message uses `$($swapVerification.Output)` which could be verbose. Minor issue.

#### 2. **Calling Disable-LinuxWorkerNodeSwap in Start-LinuxWorkerNodeServices (Line 254)**

**Issue Found:**
- This is a **significant behavioral change**. The swap disablement is now called **every time** services start, after SSH is verified but **before** starting the actual services.
- This adds latency to every service restart cycle and performs redundant operations if swap is already disabled.
- **However**, this may be intentional to ensure persistence is re-applied. No logic error here, just a design choice worth noting.

#### 3. **Kubelet Activation Retry Logic (Lines 294-304)**

**Logic Analysis:**
- The retry logic waits for kubelet to transition from 'activating' to another state (up to 12 retries × 5 seconds = 60 seconds total)
- **Line 304:** The loop condition `$attempt -le $maxKubeletActivationRetries -and $kubeletStatus -eq 'activating'` correctly breaks when status changes
- This is well-implemented and should handle slow kubelet startups gracefully

#### 4. **Diagnostic Command Changes (Line 330)**

**Improvements:**
- Adds `sudo` prefix for proper permissions (good)
- Uses `systemctl show` with specific properties instead of generic `is-active` (better diagnostics)
- However, the command is now longer and more complex, which increases timeout risk with `ExecutionTimeoutSeconds 15`

**Minor Concern:**
- The mixed use of `sudo systemctl show` and `sudo systemctl is-active` with `2>/dev/null` creates inconsistent error suppression. The `show` command won't suppress errors but `is-active` will.

### Summary

**No critical bugs found.** The changes are well-intentioned and mostly well-implemented:

✅ Swap persistence mechanism is sound (systemd masking is better than file deletion)
✅ Verification step is present  
✅ Kubelet activation waiting is well-coded
✅ Added error handling is appropriate

⚠️ Minor concerns:
- No bounds checking on the array split when parsing systemctl output
- Redundant swap disablement calls in Start-LinuxWorkerNodeServices (possibly intentional for persistence)
- Diagnostic command complexity and error handling inconsistency

---

## Files Reviewed
- ✅ lib/scripts/windows/worker/linux/common/LinuxWorkerNode.Common.ps1

